PR #315, opened 2023-07-18
Update docker compose dev: equate with stage env

Alex-Izquierdo:
Now that we have a docker-compose for macs we can add the latest updates from the stage env to the docker-compose-dev. 
Includes:
- Add podman service for activations to not depend on local podman anymore as we do in stage env. 
- Add the envvar for EDA_NUM_WORKERS and update the default value to match the stage env
- Remove postgres extra stuff not present in the stage or the operator env to use the default ones. 
- Add envvars for AWX integration. 
- Add UI pod
- Update pod names to the same values in stage (ui expects a hostname called "eda-api")

### DIFF
--- a/Taskfile.dist.yaml
+++ b/Taskfile.dist.yaml
@@ -160,7 +160,7 @@ tasks:
     cmds:
       - task: docker
         vars:
-          CLI_ARGS: run --rm api aap-eda-manage migrate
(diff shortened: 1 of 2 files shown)


hsong-rh (approved):
LGTM

Dostonbek1 (dismissed):

Dostonbek1 (approved):

hsong-rh (approved):

PR #316, opened 2023-07-19
[AAP-12935] Cleanup jobs from RQ when disable pending activations

hsong-rh:
This PR contains:
1. removing jobs from RQ when disables the pending activations;
2. the pending activations have correct `pending` state;
3. for OCP, fixing activation status to show `stopped` when disable it;
4. for OCP, fixing activation unexpected behaviors when restarting it;

Solves: 
[AAP-11762](https://issues.redhat.com/browse/AAP-11762)
[AAP-12935](https://issues.redhat.com/browse/AAP-12935)
[AAP-12986](https://issues.redhat.com/browse/AAP-12986)
[AAP-13109](https://issues.redhat.com/browse/AAP-13109)
[AAP-13113](https://issues.redhat.com/browse/AAP-13113)
[AAP-13416](https://issues.redhat.com/browse/AAP-13416)
[AAP-13782](https://issues.redhat.com/browse/AAP-13782)
[AAP-14479](https://issues.redhat.com/browse/AAP-14479)


src/aap_eda/api/views/activation.py
```diff
@@ -278,7 +271,30 @@ def disable(self, request, pk):
...
             activation.is_enabled = False
-            activation.save(update_fields=["is_enabled", "modified_at"])
+
+            job_id = activation.current_job_id
+
+            if job_id:
+                queue = get_queue()
+                job = queue.fetch_job(job_id)
+                if job and job.get_status() in [
+                    "queued",
```
Alex-Izquierdo, line 281: I feel all these states should be defined in a common dataclass and be consumed from there to avoid using arbitrary hardcoded strings. What do you think?

src/aap_eda/api/views/activation.py
```diff
@@ -231,35 +238,33 @@ def enable(self, request, pk):
...
             activation_id=pk,
             deployment_type=settings.DEPLOYMENT_TYPE,
             ws_base_url=settings.WEBSOCKET_BASE_URL,
             ssl_verify=settings.WEBSOCKET_SSL_VERIFY,
         )
 
+        activation.is_enabled = True
+        activation.failure_count = 0
+        activation.current_job_id = job.id
+        if job.get_status() == "queued":
```
Alex-Izquierdo, line 263: This should use JobStatus enum, don't?

src/aap_eda/api/views/activation.py
```diff
@@ -277,9 +282,34 @@ def disable(self, request, pk):
...
+                job = queue.fetch_job(job_id)
+                if job and job.get_status(refresh=True) in [
+                    JobStatus.QUEUED.value,
+                    JobStatus.STARTED.value,
+                    JobStatus.DEFERRED.value,
+                    JobStatus.SCHEDULED.value,
+                ]:
+                    job.delete(delete_dependents=True)
+                    logger.info(
+                        f"Job {job.id} is removed from the queue: {queue.name}"
```
Alex-Izquierdo, line 299: I think it would be useful if we print here the activation_id or activation_name

src/aap_eda/api/views/activation.py
```diff
@@ -97,7 +114,6 @@ def create(self, request):
         activation = self._get_activation_dependent_objects(
             response_serializer.data
         )
-        activation["status"] = ActivationStatus.STARTING.value
```
Alex-Izquierdo, line 100: I feel we should have here a new state "created", don't you think?

src/aap_eda/api/views/activation.py
```diff
@@ -296,6 +287,16 @@ def disable(self, request, pk):
                     deployment_type=settings.DEPLOYMENT_TYPE,
                 )
 
+            # update activation's details
+            activation.status = ActivationStatus.STOPPED
```
bzwei, line 291: The status should be STOPPING.
The whole block up to line 298 should be indent right to under the `if current_instance block` because the status should change only if the deactivate is applied.
We should have an else block to raise a response like 409.

src/aap_eda/api/views/activation.py
```diff
@@ -231,35 +226,33 @@ def enable(self, request, pk):
...
             activation_id=pk,
             deployment_type=settings.DEPLOYMENT_TYPE,
             ws_base_url=settings.WEBSOCKET_BASE_URL,
             ssl_verify=settings.WEBSOCKET_SSL_VERIFY,
         )
 
+        activation.is_enabled = True
+        activation.failure_count = 0
+        activation.current_job_id = job.id
+        if job.get_status() == JobStatus.QUEUED:
```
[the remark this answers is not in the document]
hsong-rh: The `if` here is only for the scenario that activations are more than workers. The `else` is the normal case, `status` will be updated accordingly

src/aap_eda/api/views/activation.py
hsong-rh: No, we should always update activation's status when it's disabled, no matter whether its activation instance needs to be cleaned up or not.

src/aap_eda/api/views/activation.py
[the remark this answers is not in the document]
hsong-rh: @bzwei The `if` is removed, so now we always mark the activation's status as `PENDING` when it's enabled, and `STOPPED` as it's disabled.

src/aap_eda/services/ruleset/activate_rulesets.py
```diff
@@ -135,28 +128,32 @@ def activate(
                     activation_instance=instance,
                     decision_environment=activation.decision_environment,
                 )
-                instance.status = ActivationStatus.COMPLETED
             else:
                 raise ActivationException(f"Unsupported {deployment_type}")
 
-            if str(instance.status) == ActivationStatus.COMPLETED.value:
+            if instance.status == ActivationStatus.COMPLETED:
```
bzwei, line 134: I feel previous statement `if str(instance.status) == ActivationStatus.COMPLETED.value` is safer.

hsong-rh: No need to convert to str, enum comparison is ok (https://docs.python.org/3/howto/enum.html#comparisons)

src/aap_eda/api/views/activation.py
hsong-rh: Yes, the initial status should be `STOPPING`, `STOPPED` is marked only when disable is done

src/aap_eda/services/ruleset/activate_rulesets.py
Alex-Izquierdo: We could extend [our enum classes](https://github.com/ansible/eda-server/blob/main/src/aap_eda/core/enums.py#L74) to inherit from str class, that would allow us to make safe comparisions with strings if we suspect that the value might be just an string. 
```
class ActivationStatus(str, DjangoEnum):
...
```

In that case `instance.status == ActivationStatus.COMPLETED:` will return true if instance.status is a string or a enum value. 

This would be equivalent to [StrEnum class](https://docs.python.org/3/library/enum.html#enum.StrEnum) included in python3.11

src/aap_eda/services/ruleset/activation_kubernetes.py
```diff
@@ -338,32 +349,34 @@ def run_activation_job(
...
                     namespace=namespace,
                     job_name=job_name,
                 )
+                logger.info(f"Service: {job_name} is deleted.")
 
-    def delete_secret(self, secret_name, namespace, job_name) -> None:
-        # wait until job is done
-        while True:
-            ret = client.BatchV1Api().list_namespaced_job(
+    def delete_secret(self, secret_name, namespace) -> None:
```
Alex-Izquierdo, line 361: Minor concern: with this change with the labels, any secret created before the upgrade will not be removed. 
Is there a reason to do it this way instead of capturing the exception?

src/aap_eda/services/ruleset/activation_kubernetes.py
```diff
@@ -338,32 +349,34 @@ def run_activation_job(
...
+        secrets = self.client_api.list_namespaced_secret(
+            namespace=namespace,
+            label_selector="aap=eda",
+        )
+
+        if secrets and secrets.items:
```
Alex-Izquierdo, line 367: If we call this function we expect to delete the secret. So if the secret is not found I feel we should have also a log record indicating it in case we should not raise an exception.

Alex-Izquierdo:
FYI: Recommended practices about labels in k8s: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

src/aap_eda/services/ruleset/activate_rulesets.py
hsong-rh: We have many defined DjangoEnum classes, should we change them all? or only change `ActivationStatus`?

src/aap_eda/services/ruleset/activation_podman.py
```diff
@@ -162,8 +165,11 @@ def run_worker_mode(
         finally:
             if container and self.client.containers.exists(container.id):
                 container_id = container.id
-                container.remove(force=True, v=True)
-                logger.info(f"Container {container_id} is cleaned up.")
+                try:
```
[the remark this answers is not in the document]
hsong-rh: The deactivation can happen at anytime to delete the container, even when passing our check on the `container`, which is in the memory and the `exists`. We have to make sure NotFound in this case is expected.

src/aap_eda/services/ruleset/activation_kubernetes.py
hsong-rh: Good catch, rewrote the delete_secret logic

src/aap_eda/services/ruleset/activate_rulesets.py
Alex-Izquierdo: Well, it was just an idea, not a requirement for this PR. IMO it would be weird to do it only on one class because the rest are in the same situation. Maybe is better to do it in another PR .

hsong-rh: @Alex-Izquierdo @bzwei So far keep it as it is, will do it in another PR if necessary

bzwei (approved):

mkanoor (approved):

PR #318, opened 2023-07-24
[AAP-12134] Add a new RQ to manage non-rulebook tasks

Dostonbek1:
## Problem Definition

Currently we have a fixed number of workers defined at installation time based on the hardware resources. With the current implementation each worker takes care of one activation. If the user creates enough activations, all the workers can be busy with activations preventing other async tasks like import projects.

## Solution

Implement a different queue for activations and ensure not all the workers are subscribed to that queue to ensure that there are always available workers for the rest of async tasks, for example project import.

## Testing 

1. Create multiple activations (preferably more than the number of workers available)
2. Start all activations and import a new project at the same time
3. All tasks should go through in the order they come in without any issues (e.g. activations getting stuck on starting, etc.)

JIRA ticket: [AAP-12134](https://issues.redhat.com/browse/AAP-12134)

NOTE: This PR should only be merged in sync with productization and after this PR https://github.com/ansible/eda-server/pull/320 gets merged.


tools/docker/docker-compose-stage.yaml
```diff
@@ -75,6 +75,27 @@ services:
     volumes:
       - ${EDA_HOST_PODMAN_SOCKET_URL:-/run/user/1000/podman/podman.sock}:/run/podman/podman.sock:z
 
+  eda-manager:
+    user: "${EDA_POD_USER_ID:-0}"
+    deploy:
+      replicas: ${EDA_NUM_WORKERS:-2}
```
hsong-rh, line 81: same here

Alex-Izquierdo (changes requested):
Why are you duplicating the dockerfile? I don't think that is necessary. It would be just another worker and only the args for the command line should change. 

Also, I feel "manager" is quite generic. Why not rename "default" for activations which it looks its goal and leave the "default" for the rest?

tools/docker/docker-compose-stage.yaml
```diff
@@ -75,6 +75,27 @@ services:
...
+  eda-manager:
+    user: "${EDA_POD_USER_ID:-0}"
+    deploy:
+      replicas: ${EDA_NUM_WORKERS:-1}
+    image: "${EDA_IMAGE:-quay.io/ansible/eda-server:main}"
+    environment: *common-env
+    command:
+      - aap-eda-manage
+      - rqworker
+      - --with-scheduler
```
Alex-Izquierdo, line 87: We must take care of this if we finally merge https://github.com/ansible/eda-server/pull/263
(now the PR link is broken, it will be fixed soon) It is the PR where is implemented the RQ scheduler, so the scheduler is executed in a different pod

Alex-Izquierdo:
Please update the description, the old PR for RQ scheduler get lost in the transition to a public repo. The new one is: https://github.com/ansible/eda-server/pull/320

tools/docker/docker-compose-dev.yaml
```diff
@@ -57,13 +57,32 @@ services:
...
+
+  activation-worker:
     deploy:
       replicas: ${EDA_NUM_WORKERS:-2}
     image: 'localhost/aap-eda-worker'
     build:
       context: ../../
       dockerfile: tools/docker/worker/Dockerfile
+      args:
+        - WORKER_CLASS: 'aap_eda.core.tasking.Activation'
```
Alex-Izquierdo, line 85: `WORKER_CLASS` variable should have a default value in the dockerfile, it is usually an antipattern to require in a mandatory build_arg to build an image. As this value is only consumed at execution time, not even is necessary to implement it as build_arg. It can be just an environment variable. I suggest one of the following approaches:

- Declare the envvar WORKER_CLASS with a default value in the dockerfiles, and then just override that environment variable in the docker-compose files. 

OR

- Not touch the dockerfile and just override the command line as you already do for docker-compose-stage and docker-compose-mac

The second one is the preferred one IMO, is less disruptive and would not require changes in the eda-operator

src/aap_eda/tasks/ruleset.py
```diff
@@ -54,9 +54,10 @@ def activate_rulesets(
 
 @job
```
Alex-Izquierdo, line 55: should not this be @job("activation") too?

Alex-Izquierdo (changes requested):

src/aap_eda/tasks/ruleset.py
hsong-rh: @Alex-Izquierdo No, the `deactivate` has to be put into `default` queue, otherwise will be stuck by activation tasks.

Taskfile.dist.yaml
```diff
@@ -38,12 +38,19 @@ tasks:
         vars:
           CLI_ARGS: shell
 
-  run:worker:
-    desc: "Run tasking worker locally"
+  run:worker:default:
```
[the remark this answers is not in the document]
hsong-rh: @Alex-Izquierdo Nice to have, but not required. I usually start the workers separately, easy to monitor the outputs.

src/aap_eda/core/tasking/__init__.py
```diff
@@ -129,3 +137,51 @@ def __init__(
             prepare_for_work,
             serializer,
         )
+
+
+class Activation(_Worker):
```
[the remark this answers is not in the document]
Dostonbek1: Should I rename the `Default` to `DefaultWorker` or just only `Activation` to `ActivationWorker`?

Taskfile.dist.yaml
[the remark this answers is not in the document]
Alex-Izquierdo: I'm suggesting to add new command to run both workers in addition to the current stuff. If you agree, we can leave it as it is and if anyone wants/needs that command we can add it in another PR.

src/aap_eda/core/tasking/__init__.py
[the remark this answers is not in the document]
Alex-Izquierdo: IMO, it would be better `DefaultWorker` and `ActivationWorker`

src/aap_eda/tasks/ruleset.py
Alex-Izquierdo: What do you think to add the @job("default") in these places? `explicit is better than implicit`
https://peps.python.org/pep-0020/

hsong-rh (approved):
LGTM

tools/docker/docker-compose-dev.yaml
```diff
@@ -71,13 +71,33 @@ services:
...
+        condition: service_healthy
+    volumes:
+      - '../../:/app/src:z'
+    restart: always
+
+  eda-activation-worker:
     deploy:
       replicas: ${EDA_NUM_WORKERS:-2}
-    image: 'localhost/aap-eda-worker'
+    image: 'localhost/eda-activation-worker'
```
Alex-Izquierdo, line 92: Here we don't need to build another tag, it can use the same image than the default worker, both can use the `localhost/aap-eda-worker`

Also I noticed that docker-compose-mac is not using this dockerfile but the regular image, we can do the same here, but in another PR. Take into account that this different worker image is to support the "local" deployment type, but now we have the podman type as default and we can use the same image for everything as we do in the docker-compose-mac.

Alex-Izquierdo (approved):
A minor optimization for the docker-compose otherwise LGTM

hsong-rh (approved):

PR #319, opened 2023-07-25
[AAP-14441] Add more details when updating projects

hsong-rh:
In this PR:
1. allow to update project with credential_id is 0 or None (clear credential from project);
2. raise 400 bad request if credential_id is invalid (the credential not found);
3. update fields only accept `name, description and credential_id`, 'url` cannot be updated.

Solves: [AAP-14441](https://issues.redhat.com/browse/AAP-14441)

### DIFF
--- a/src/aap_eda/api/serializers/__init__.py
+++ b/src/aap_eda/api/serializers/__init__.py
@@ -46,6 +46,7 @@ from .project import (
     ProjectReadSerializer,
     ProjectRefSerializer,
     ProjectSerializer,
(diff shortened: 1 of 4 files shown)


Dostonbek1 (commented):
I was able to remove credential from the project by setting the `credential_id` to 0 through the API but the UI doesn't have an option to remove the credential on Project Edit but only by completely deleting the credential. Will this change be made on the UI later or is this an intended behavior? @lgalis

Dostonbek1 (approved):

PR #320, opened 2023-07-26
[AAP-12383] Standalone RQ scheduler

cutwater:
* Install `rq-scheduler`.
* Add all-in-one management command `scheduler`, that performs cleanup of previously scheduled jobs, adds jobs specified in settings and executes RQ scheduler.
* Add new service to the docker development environment and K8S environment.
* Fix existing delated jobs.


Alex-Izquierdo:
Hi @cutwater, as we agreed, I take the ownership of this PR. thanks for everything. 

Closed in favor of: https://github.com/ansible/eda-server/pull/321

PR #321, opened 2023-07-26
[AAP-12383] Standalone RQ scheduler

Alex-Izquierdo:
* Install `rq-scheduler`.
* Add all-in-one management command `scheduler`, that performs cleanup of previously scheduled jobs, adds jobs specified in settings and executes RQ scheduler.
* Add new service to the docker development environment and K8S environment.
* Fix existing delated jobs.


src/aap_eda/core/management/commands/scheduler.py
```diff
@@ -0,0 +1,145 @@
...
+        job.delete()
+
+
+def add_startup_jobs(scheduler: Scheduler) -> None:
+    if not RQ_STARTUP_JOBS:
+        logger.info("No scheduled jobs. Skipping.")
+        return
+
+    for entry in RQ_STARTUP_JOBS:
+        logger.info('Adding startup job "%s"', entry["func"])
```
hsong-rh, line 102: Codes are duplicate here, only difference is the input job list, should we define a function for the duplicates?

src/aap_eda/core/management/commands/scheduler.py
```diff
@@ -0,0 +1,145 @@
...
+    if not RQ_CRON_JOBS:
+        logger.info("No scheduled jobs. Skipping.")
+        return
+
+    for entry in RQ_CRON_JOBS:
+        logger.info('Adding cron job "%s"', entry["func"])
+        scheduler.cron(**entry)
+
+
+class Command(rqscheduler.Command):
```
hsong-rh, line 138: We also need a `help` message for the command

bzwei (approved):

src/aap_eda/core/management/commands/scheduler.py
Alex-Izquierdo: Merged in a single function.

src/aap_eda/core/management/commands/scheduler.py
Alex-Izquierdo: Done!

Alex-Izquierdo:
Added changes as per suggestions, please review again.

hsong-rh (dismissed):

bzwei (dismissed):

Alex-Izquierdo:
It is weird, all the required checks pass but GH doesn't allow to merge the PR. I don't know what is happening it might be a bug in GitHub.  Since this only happens here, I will bypass the protection (which should be correct) to merge the PR. 
Screenshot from 2023-07-28 17-28-05

PR #323, opened 2023-07-27
[AAP-14512] Fix Audit Rules filter

Dostonbek1:


ttuffin (approved):

ttuffin:
I was able to reproduce it in another OCP environment and after applying the patch I've verified it has resolved the issue. I also confirmed that the patch has not negatively impacted the filtering on RHEL.

Alex-Izquierdo (approved):

PR #325, opened 2023-07-28
Minor fixes to docker-compose files

Alex-Izquierdo:
- Add missing volume mount for the scheduler
- Remove a wrong socket mount in stage. 
- Remove the dockerfile for the "local" deployment type, reconfigure the docker-compose file to use the same standard image


hsong-rh (approved):

Alex-Izquierdo:
Conflicts resolved, please review again.

tools/docker/docker-compose-stage.yaml
```diff
@@ -103,8 +103,6 @@ services:
       postgres:
         condition: service_healthy
     restart: always
-    volumes:
-      - ${EDA_HOST_PODMAN_SOCKET_URL:-/run/user/1000/podman/podman.sock}:/run/podman/podman.sock:z
```
Dostonbek1, line 107: Why are we removing volumes here but keeping it for `eda-default-worker`?

tools/docker/docker-compose-dev.yaml
```diff
@@ -70,12 +70,11 @@ services:
...
       postgres:
         condition: service_healthy
+    volumes:
+      - '../../:/app/src:z'
 
   eda-default-worker:
-    image: 'localhost/eda-default-worker'
-    build:
-      context: ../../
-      dockerfile: tools/docker/worker/Dockerfile
```
Dostonbek1, line 78: Since we are removing the Dockerfile, we might need to add this command here:
```
command:
    command:
      - aap-eda-manage
      - rqworker
      - --worker-class
      - aap_eda.core.tasking.DefaultWorker
```

Alex-Izquierdo: added

tools/docker/docker-compose-stage.yaml
Alex-Izquierdo: It should be removed in both. good catch. Updated.

Dostonbek1 (approved):

hsong-rh (approved):

PR #327, opened 2023-07-31
[AAP-13795] Add Activation missing last restarted field

Dostonbek1:
- Add missing last restarted date field on Activations Read Endpoint: `http://localhost:8000/api/eda/v1/activations/<activation_id>/`

### DIFF
--- a/src/aap_eda/api/views/activation.py
+++ b/src/aap_eda/api/views/activation.py
@@ -374,6 +374,11 @@ class ActivationViewSet(
             activation_id=activation["id"]
         )
         activation["instances"] = activation_instances
+        activation["restarted_at"] = (
+            activation_instances.latest("started_at").started_at
+            if activation_instances
+            else None
+        )
 
         return activation
 
--- a/tests/integration/api/test_activation.py
+++ b/tests/integration/api/test_activation.py
@@ -211,6 +211,7 @@ def test_create_activation(activate_rulesets: mock.Mock, client: APIClient):
     }
     assert activation.rulebook_name == TEST_RULEBOOK["name"]
     assert activation.rulebook_rulesets == TEST_RULESETS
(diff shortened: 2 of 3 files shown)


tests/integration/api/test_activation.py
```diff
@@ -334,6 +335,15 @@ def test_retrieve_activation(client: APIClient, with_project):
     assert data["extra_var"] == {
         "id": activation.extra_var.id,
     }
+    activation_instances = models.ActivationInstance.objects.filter(
+        activation_id=activation.id
+    )
+    if activation_instances:
```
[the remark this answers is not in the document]
Alex-Izquierdo: Ok, as we discussed, it seems a more general problem of these tests that would require a refactor.

Alex-Izquierdo (approved):
LGTM, it works as expected.

hsong-rh (approved):

src/aap_eda/api/views/activation.py
```diff
@@ -374,6 +374,11 @@ def _get_activation_dependent_objects(self, activation):
             activation_id=activation["id"]
         )
         activation["instances"] = activation_instances
+        activation["restarted_at"] = (
```
Alex-Izquierdo, line 377: Hi @Dostonbek1 I think this logic is wrong. you are setting the field from the first instance, but if there is only one instance, the activation has not been restarted so this field should not have value

Alex-Izquierdo: Also, the restart count is only increased with automatic restarts, so returning a value when restart count is 0 (the activation has been disabled/enabled manually but it has more than one instance) would be confusing.

Alex-Izquierdo: https://github.com/ansible/eda-server/pull/385

PR #329, opened 2023-08-01
Refactor WS params for activation workflow.

Alex-Izquierdo:
As suggested by @mkanoor, there are some params in the workflow activation that are passed through with necessity since these params can be taken from the global config. This PR simplifies removes the ws_url and ws_verify_ssl and simplifies the function signatures for the activation workflow. 

Internal ref: https://issues.redhat.com/browse/AAP-14621


src/aap_eda/services/ruleset/activate_rulesets.py
```diff
@@ -297,13 +293,7 @@ def _restart_activation(
...
 
         activation_db_logger.write(msg)
-        enqueue_restart_task(
-            seconds,
-            activation.id,
-            self.deployment_type,
-            self.ws_base_url,
-            self.ssl_verify,
-        )
+        enqueue_restart_task(seconds, activation.id)
```
hsong-rh, line 296: We still need `deployment_type` here, right? Its definition has 3 parameters

Alex-Izquierdo: Thanks for the warn, I forgot to push my last commit where I change the function to take the deployment type from the global config.

hsong-rh: Should we remove deployment_type from parameter list too?

hsong-rh: I mean we can alsoe remove deployment_type from here:
https://github.com/ansible/eda-server/pull/329/files#diff-3aa1d8dc47ee0af34b544249cc19758b5919dbddf7057e84dc032f49f6235f33L108

Alex-Izquierdo: Agree, I wanted to start this refactor at the scope of the enqueue_restart_task, to keep an small PR and doing this kind of refactors in small chunks. Do you agree to address it in later PR's?

hsong-rh (approved):

src/aap_eda/services/ruleset/activate_rulesets.py
```diff
@@ -86,15 +86,11 @@ def __init__(self, cwd: Optional[str] = None):
         self.service = AnsibleRulebookService(cwd)
 
     def activate(
-        self,
-        activation: models.Activation,
-        deployment_type: str,
-        ws_base_url: str,
-        ssl_verify: str,
+        self, activation: models.Activation, deployment_type: str
```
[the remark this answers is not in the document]
Alex-Izquierdo: @mkanoor Well, I thought the same, but I say the same to you as to @hsong-rh. This method is called in several places and it would require more changes in the code and the tests. I want to keep small the PR and do this kind of refactors in small chunks. Do you agree to address it in later iterations?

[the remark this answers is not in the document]
Alex-Izquierdo: On the other hand, I think changing from Podman to Kubernetes would definitely be a new feature. I'm not sure about the scope of such a change.

Alex-Izquierdo:
rebased, please review again. @hsong-rh @mkanoor

hsong-rh (approved):

Alex-Izquierdo:
needed to be rebased again :(

Dostonbek1 (approved):
tested, lgtm

hsong-rh (approved):

PR #330, opened 2023-08-01
[AAP-14635] Correct worker class name in deployment script

hsong-rh:
This PR contains:
1. added the missing change for worker class name in the deployment script, so OCP worker pod can start;
2. added the missing container `eda-activation-worker`;
3. rename `eda-worker` to `eda-default-worker`

Solves: [AAP-14635](https://issues.redhat.com/browse/AAP-14635)


Alex-Izquierdo (dismissed):

Alex-Izquierdo (approved):

mkanoor:
@hsong-rh Don't we need role bindings and service account for the activation-worker? From what I recall those role bindings and service accounts were needed to start activations in k8s env

Alex-Izquierdo:
Good catch! @mkanoor, we need it for new worker.

hsong-rh:
@mkanoor @Alex-Izquierdo Now `activation-worker` has role binding and service accounts, it will handle activation tasks. `default-worker` doesn't need roles and service account.

Alex-Izquierdo:
@hsong-rh I think both workers need it. Deactivate rulesets is executed in the default worker. And we expect that the default worker might do other tasks (scheduled tasks like monitor activations)

hsong-rh:
> @hsong-rh I think both workers need it. Deactivate rulesets is executed in the default worker. And we expect that the default worker might do other tasks (scheduled tasks like monitor activations)

OK, make sense, added them

Alex-Izquierdo (approved):
LGTM

Dostonbek1 (approved):
lgtm

PR #331, opened 2023-08-02
Check to make sure only one AWX token exists

Alex-Izquierdo:
Recreates the original PR https://github.com/ansible/eda-server/pull/280, which for some reason it can not be reopened after the migration to a public repo. 

Only grab the first token. So checking to make sure we have at least one token makes sense.
Jira: https://github.com/ansible/eda-server/pull/280

### DIFF
--- a/src/aap_eda/api/exceptions.py
+++ b/src/aap_eda/api/exceptions.py
@@ -29,6 +29,7 @@ __all__ = (
     "Conflict",
     "Unprocessable",
     "PermissionDenied",
+    "TooManyControllerTokens",
 )
 
 
@@ -70,7 +71,7 @@ class NoControllerToken(APIException):
 class TooManyControllerTokens(APIException):
     status_code = 422
     default_detail = (
-        "More than one controller token found, "
+        "Found an existing controller token, "
         "currently only 1 token is supported"
     )
 
--- a/tests/integration/api/test_awx_token.py
+++ b/tests/integration/api/test_awx_token.py
@@ -67,21 +67,6 @@ def test_create_controller_token(client: APIClient, user: models.User):
         row = cursor.fetchone()
         assert row[0].startswith("$encrypted$fernet-256$")
 
-    response = client.post(
-        f"{api_url_v1}/users/me/awx-tokens/",
-        data={
-            "name": "Test token 2",
-            "description": "Token description",
-            "token": "test-token-value",
-        },
-    )
-    assert response.status_code == status.HTTP_201_CREATED
-    data = response.json()
-    assert data["name"] == "Test token 2"
-    assert data["description"] == "Token description"
-    assert data["user_id"] == user.id
-    assert "token" not in data
-
 
 @pytest.mark.django_db
 def test_create_token_missing_field(client: APIClient, user: models.User):
@@ -109,7 +94,7 @@ def test_create_token_missing_field(client: APIClient, user: models.User):
 
 
 @pytest.mark.django_db(transaction=True)
-def test_create_token_duplicate_name(client: APIClient, user: models.User):
+def test_create_token_to_many_tokens(client: APIClient, user: models.User):
     models.AwxToken.objects.create(
         user=user, name="test-token", token="test-token-value"
     )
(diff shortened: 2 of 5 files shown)


mkanoor:
@Alex-Izquierdo I think we should also fix this related issue

https://github.com/ansible/eda-server/blob/ed741364238c4e618946ab99a3b46b87c35d9d7f/src/aap_eda/wsapi/consumers.py#L113

Since it blindly assumes that there is a token available when the ansible-rulebook comes back on the websocket channel looking for parameters and the missing token causes an exception to be thrown. For now I think the assumption is you always need a token as we make more changes we can parse the rulebook and see if it has run_job_template action and then prompt for a token but even in that case we need this code path to be blocked. There is a stack trace for this code in many JIRA tickets, the server dying ends up closing the websocket connection and the ansible rulebook ends saying error exchanging web socket data. https://issues.redhat.com/browse/AAP-11677 https://issues.redhat.com/browse/AAP-11681

Since we have the activation instance here we should mark it as broken from the server side itself if the token is missing. The  server provides the token to ansible-rulebook. 

https://github.com/ansible/eda-server/blob/ed741364238c4e618946ab99a3b46b87c35d9d7f/src/aap_eda/wsapi/consumers.py#L102
Since we have the activation id we should be able to set it to Failure from the server side itself instead of having the ansible-rulebook say that web socket connection closed.

mkanoor:
@Alex-Izquierdo There needs to be another check here for the token
https://github.com/ansible/eda-server/blob/ed741364238c4e618946ab99a3b46b87c35d9d7f/src/aap_eda/tasks/ruleset.py#L44
I think we need to figure out the token based on the user_id that created the activation since if multiple users create the activations we need to find on whose behalf the worker is running the activation.
https://github.com/ansible/eda-server/blob/ed741364238c4e618946ab99a3b46b87c35d9d7f/src/aap_eda/wsapi/consumers.py#L316
Maybe this function https://github.com/ansible/eda-server/blob/ed741364238c4e618946ab99a3b46b87c35d9d7f/src/aap_eda/wsapi/consumers.py#L308 can come out and be placed in a shared location and be called from multiple places. We can pass in the activation id instead of the message. It could also throw an exception saying "No tokens defined"

I think there are at least 3 places we need to block it

1. From the UI side when the activation is created the first time (exists)
Screenshot 2023-08-02 at 4 42 26 PM

3. When the activation is started by a worker before we call K8S or podman to run the activation (missing)
4. When the ansible-rulebook comes on the websocket channel asking for the payload (missing)

I think the UI also has code to ensure that only 1 token is created I think this PR is guarding on the backend so that a REST API call doesn't end up adding an extra token.

Alex-Izquierdo:
Hi @mkanoor I totally agree with all that you say, we must fix the usage of the token. But it looks like that is out of the scope of this PR which its goal is to prevent to have more than one token. I propose to create a new issue or update some existing one (like [this one](https://issues.redhat.com/browse/AAP-13081) ) and address all these issues in a different PR.

src/aap_eda/api/exceptions.py
```diff
@@ -70,7 +71,7 @@ class NoControllerToken(APIException):
 class TooManyControllerTokens(APIException):
     status_code = 422
```
Dostonbek1, line 72: Should we use this `status.HTTP_422_UNPROCESSABLE_ENTITY` instead of just hard-coded integer? Or we could inherit from `Unprocessable` class on line 48 and customize detail field since they fall under the same error code. Thoughts?

tests/integration/api/test_awx_token.py
```diff
@@ -109,7 +94,7 @@ def test_create_token_missing_field(client: APIClient, user: models.User):
 
 
 @pytest.mark.django_db(transaction=True)
-def test_create_token_duplicate_name(client: APIClient, user: models.User):
+def test_create_token_to_many_tokens(client: APIClient, user: models.User):
```
Dostonbek1, line 97: Possibly a typo: `_to_many_token` ?

Dostonbek1:
Except for the above minor comments, LGTM. Works as expected.
Screenshot 2023-08-09 at 5 00 25 PM

tests/integration/api/test_awx_token.py
Alex-Izquierdo: likely :)

src/aap_eda/api/exceptions.py
Alex-Izquierdo: Agree

PR #332, opened 2023-08-02
[AAP-12987] Fix Rule Fire Count lag

Dostonbek1:
- fixes the delay and mismatch between "Fire count" on Activations Page and the number of Audit Rules completed for the given activation

Solves: [AAP-12987](https://issues.redhat.com/browse/AAP-12987)

### DIFF
--- a/src/aap_eda/wsapi/consumers.py
+++ b/src/aap_eda/wsapi/consumers.py
@@ -224,7 +224,7 @@ class AnsibleRulebookConsumer(AsyncWebsocketConsumer):
             activation = models.Activation.objects.get(
                 pk=activation_instance.activation_id
             )
-            if activation.ruleset_stats:
+            if message.ruleset in activation.ruleset_stats:
                 activation.ruleset_stats[message.ruleset][
                     "rulesTriggered"
                 ] += 1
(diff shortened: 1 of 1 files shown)


Alex-Izquierdo (approved):

src/aap_eda/wsapi/consumers.py
```diff
@@ -217,6 +217,18 @@ def insert_audit_rule_data(self, message: ActionMessage) -> None:
...
+
+            activation_instance = models.ActivationInstance.objects.get(
+                pk=message.activation_id
+            )
+            activation = models.Activation.objects.get(
+                pk=activation_instance.activation_id
+            )
+            if message.ruleset in activation.ruleset_stats:
+                activation.ruleset_stats[message.ruleset][
+                    "rulesTriggered"
```
mkanoor, line 229: @Dostonbek1 This is controlled from the ansible-rulebook side we cannot increment this on the server side. A session stat is sent at every heartbeat interval, if no events have arrived into the ansible-rulebook we will still be sending the data from the ansible-rulebook and this code will incorrectly say rules have been triggered when no events have been processed. This seems like a different issue related to https://issues.redhat.com/browse/AAP-13874

Dostonbek1:
Closing this PR since the issue has been fixed on the ansible-rulebook side.

PR #338, opened 2023-08-04
[AAP-13064] Use the scheduler to monitor activations

bzwei:
Periodically check the healthiness of activation instances with the help by the recently enabled rq-scheduler.
Attempt to stop instances whose heartbeats are no longer received. This will help to clean any orphaned DEs. 
Restart related activates if the restart policy is on-failure or always.

Fixes AAP-13064: EDA server can not be restarted - staled states in activations


src/aap_eda/tasks/ruleset.py
```diff
@@ -55,24 +55,32 @@ def activate_rulesets(
 
 @job("default")
 def deactivate_rulesets(
+    user_deactivate: bool,
```
hsong-rh, line 58: The name `user_deactivate` is confused, how about changing it to `triggered_by_scheduler`?

src/aap_eda/tasks/ruleset.py
```diff
@@ -55,24 +55,32 @@ def activate_rulesets(
...
     activation_instance_id: int,
     deployment_type: str,
 ) -> None:
     instance = models.ActivationInstance.objects.get(pk=activation_instance_id)
     logger.info(f"Task started: Deactivate Activation ({instance.id})")
 
+    final_status = (
```
hsong-rh, line 65: Suggest to move this logic into `deactivate` function, it's confused and is not directly used here. Something like:

```
    ActivateRulesets().deactivate(
        instance,
        deployment_type,
        triggered_by_scheduler,
    )
```

src/aap_eda/services/ruleset/activate_rulesets.py
```diff
@@ -209,7 +210,7 @@ def deactivate(
             logger.info(
                 f"Stopped Activation, Name: {instance.name}, ID: {instance.id}"
             )
-            set_activation_status(instance, ActivationStatus.STOPPED)
```
hsong-rh, line 212: If the deactivation triggered by scheduler is marked as `failed`, does restart will happen if policy is on_failure? do we need to mark as `failed` in this case?

bzwei: Setting FAILED status by deactivation process does not trigger auto restart. Auto restart is done by either the activation process or monitor_activation on UNRESPONSIVE activations.

bzwei:
depend on #316

bzwei:
@hsong-rh @Alex-Izquierdo @Dostonbek1 please review

Dostonbek1:
@bzwei Failed activations don't seem to restart although restart policy is either `always` or `on-failure`. Tested with the following rulebooks:

1. [fail_job_template.yml](https://github.com/Dostonbek1/test-project-for-eda/blob/main/rulebooks/fail_job_template.yml)
2. [bad_job_template.yml](https://github.com/Dostonbek1/test-project-for-eda/blob/main/rulebooks/bad_job_template.yml)

ttuffin:
> @bzwei Failed activations don't seem to restart although restart policy is either `always` or `on-failure`. Tested with the following rulebooks:
> 
> 1. [fail_job_template.yml](https://github.com/Dostonbek1/test-project-for-eda/blob/main/rulebooks/fail_job_template.yml)
> 2. [bad_job_template.yml](https://github.com/Dostonbek1/test-project-for-eda/blob/main/rulebooks/bad_job_template.yml)
> 
> Screenshot 2023-08-24 at 10 52 35 AM

I came across [this bug](https://issues.redhat.com/browse/AAP-15370) yesterday on main branch, so it's not specific to this PR.

src/aap_eda/services/ruleset/activate_rulesets.py
```diff
@@ -313,7 +313,7 @@ def _restart_activation(
             logger.info(msg)
 
         activation_db_logger.write(msg)
-        enqueue_restart_task(seconds, activation.id)
+        # no further action is needed. The scheduler knows how to restart
```
hsong-rh, line 316: Seems like this function do nothing except logging messages. The real restart logic is handled by the scheduler, is this expected? My understanding is scheduler only do the healthy checking. If it's expected, should we remove these checkings (on_complete, on_failure, ...)?

src/aap_eda/tasks/ruleset.py
```diff
@@ -130,75 +139,93 @@ def deactivate_rulesets(
     ActivateRulesets().deactivate(
         instance,
         deployment_type,
+        final_status,
     )
 
 
-# This is job will be scheduled by a scheduler which currently is unavailable
+# Started by the scheduler, executed by the default worker
 def monitor_activations() -> None:
```
hsong-rh, line 147: Do you need to decorate this function to `default` queue?

mkanoor:
@bzwei how many monitor tasks would be running at a given time. If the scheduler interval is say 1 minute it there a possibility that multiple monitor tasks will get started. Since the monitor task is going to do the deactivation and then reque the next activation if a restart is needed it's possible that we might take a little bit of time for all these things to be accomplished. If there are too many unresponsive activations and they all need to be deactivated and restarted another monitor task can get started by the scheduler. Should the monitor task check if there is another monitor running before starting.

mkanoor:
@bzwei @hsong-rh There is another race condition that can happen here, if the deactivation is being handled by the monitor task for unresponsive activations and the deactivation task for user requested deactivations. If the scheduler pops a monitor task and at the same time the user disables the activation we would have 2 deactivations running for the same activation.

mkanoor:
> @bzwei Failed activations don't seem to restart although restart policy is either `always` or `on-failure`. Tested with the following rulebooks:
> 
> 1. [fail_job_template.yml](https://github.com/Dostonbek1/test-project-for-eda/blob/main/rulebooks/fail_job_template.yml)
> 2. [bad_job_template.yml](https://github.com/Dostonbek1/test-project-for-eda/blob/main/rulebooks/bad_job_template.yml)
> 
> Screenshot 2023-08-24 at 10 52 35 AM

@Dostonbek1 @ttuffin if an action fails that doesn't mean that the rulebook exit code is going to be non zero. Rulebooks are long running processes and there might be scenarios where an action would fail. We track and report each action separately and you will see the failures for the action in the UI but the ansible-rulebook will still end with an exit code of 0. ansible-rulebook doesn't stop when an action fails, it reports it and moves on, its upto the customer if they feel there are too many failures they can disable the activation from the UI and stop it. For actions we don't have an attribute like **on_failure** which would dictate what to do when an action fails. 
```
    on_failure: abort
```
The current default behavior is 
```
   on_failure: continue
```

mkanoor:
@bzwei When the monitor activation task deactivates the deactivation message in the logs look like
```
Start to disable instance 19 ...
Activation being stopped at user request
Container 1c879e08a58ab86f2bbd69156578cc2e7979a058bd6028a19df5d9e1e255ec21 is removed.
2023-08-25 19:07:41,089 - ansible_rulebook.rule_set_runner - INFO - Task Task-15 finished, active actions 0
```
Should we say the Activation being stopped because it's unresponsive as opposed to Activation being stopped at user request. Since the user didn't do this the monitor task did it

mkanoor:
@bzwei if a pod is paused using
```
podman pause <container_id>
```
We can test unresponsive activations.
But this brings about an issue in deactivation which only stops running containers but not paused containers leading to busy  activation worker.
https://github.com/ansible/eda-server/blob/249bd8e586a69bf4b601b4917db8ed329f867ad2/src/aap_eda/services/ruleset/deactivation_podman.py#L56

I think this piece of code needs to be adjusted for paused containers to maybe something like this
```
               container = self.client.containers.get(container_id)
                logger.info("Container status %s", container.status)
                if container.status in ("running", "paused"):
                    if container.status == "running":
                       container.stop(ignore=True)
                    container.remove(force=True, v=True)
```
This will ensure that we can remove the running as well as paused containers.

ttuffin:
> > @bzwei Failed activations don't seem to restart although restart policy is either `always` or `on-failure`. Tested with the following rulebooks:
> > 
> > 1. [fail_job_template.yml](https://github.com/Dostonbek1/test-project-for-eda/blob/main/rulebooks/fail_job_template.yml)
> > 2. [bad_job_template.yml](https://github.com/Dostonbek1/test-project-for-eda/blob/main/rulebooks/bad_job_template.yml)
> > 
> > Screenshot 2023-08-24 at 10 52 35 AM
> 
> @Dostonbek1 @ttuffin if an action fails that doesn't mean that the rulebook exit code is going to be non zero. Rulebooks are long running processes and there might be scenarios where an action would fail. We track and report each action separately and you will see the failures for the action in the UI but the ansible-rulebook will still end with an exit code of 0. ansible-rulebook doesn't stop when an action fails, it reports it and moves on, its upto the customer if they feel there are too many failures they can disable the activation from the UI and stop it. For actions we don't have an attribute like **on_failure** which would dictate what to do when an action fails.
> 
> ```
>     on_failure: abort
> ```
> 
> The current default behavior is
> 
> ```
>    on_failure: continue
> ```

@mkanoor if the activation status is marked as "Failed" then the user will expect the restart policy will be honored. During my testing it wasn't a failed action that caused the failure, but rather a source error.

mkanoor:
> > > @bzwei Failed activations don't seem to restart although restart policy is either `always` or `on-failure`. Tested with the following rulebooks:
> > > 
> > > 1. [fail_job_template.yml](https://github.com/Dostonbek1/test-project-for-eda/blob/main/rulebooks/fail_job_template.yml)
> > > 2. [bad_job_template.yml](https://github.com/Dostonbek1/test-project-for-eda/blob/main/rulebooks/bad_job_template.yml)
> > > 
> > > Screenshot 2023-08-24 at 10 52 35 AM
> > 
> > 
> > @Dostonbek1 @ttuffin if an action fails that doesn't mean that the rulebook exit code is going to be non zero. Rulebooks are long running processes and there might be scenarios where an action would fail. We track and report each action separately and you will see the failures for the action in the UI but the ansible-rulebook will still end with an exit code of 0. ansible-rulebook doesn't stop when an action fails, it reports it and moves on, its upto the customer if they feel there are too many failures they can disable the activation from the UI and stop it. For actions we don't have an attribute like **on_failure** which would dictate what to do when an action fails.
> > ```
> >     on_failure: abort
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The current default behavior is
> > ```
> >    on_failure: continue
> > ```
> 
> @mkanoor if the activation status is marked as "Failed" then the user will expect the restart policy will be honored. During my testing it wasn't a failed action that caused the failure, but rather a source error.

@ttuffin I do see the error now, when the activation fails it doesn't seem to do a restart
```
2023-08-28 14:46:50,548 - ansible_rulebook.app - INFO - Main complete

2023-08-28 14:46:50,550 - ansible_rulebook.cli - ERROR - Terminating One of the source plugins failed

Container exit code: 1
Activation mk failed: Activation failed
```

bzwei:
@Dostonbek1 @ttuffin What you have observed should be due to the fact the activation is not restartable. If the activation fails before it ever turns into the running state, it is marked at not restartable. I have enhanced the error message in this PR.
Screenshot 2023-08-28 at 11 58 34 AM

BTW, my test shows the rulebook `fail_job_template` activation completes even though the job template itself fails. It is reasonable because they reflect different facts.
Screenshot 2023-08-28 at 11 51 39 AM

src/aap_eda/tasks/ruleset.py
bzwei: In my test there is no such need.

src/aap_eda/services/ruleset/activate_rulesets.py
bzwei: Yes. This function now only does proper logging. The logging messages are different on_complete or on_failure and cannot be done by scheduler because it does not know the context.

bzwei: I renamed the functions to _log_xx to make them more meaningful.

bzwei:
> @bzwei if a pod is paused using
> 
> ```
> podman pause <container_id>
> ```
> 
> We can test unresponsive activations. But this brings about an issue in deactivation which only stops running containers but not paused containers leading to busy activation worker.
> 
> https://github.com/ansible/eda-server/blob/249bd8e586a69bf4b601b4917db8ed329f867ad2/src/aap_eda/services/ruleset/deactivation_podman.py#L56
> 
> I think this piece of code needs to be adjusted for paused containers to maybe something like this
> 
> ```
>                container = self.client.containers.get(container_id)
>                 logger.info("Container status %s", container.status)
>                 if container.status in ("running", "paused"):
>                     if container.status == "running":
>                        container.stop(ignore=True)
>                     container.remove(force=True, v=True)
> ```
> 
> This will ensure that we can remove the running as well as paused containers.



> @bzwei @hsong-rh There is another race condition that can happen here, if the deactivation is being handled by the monitor task for unresponsive activations and the deactivation task for user requested deactivations. If the scheduler pops a monitor task and at the same time the user disables the activation we would have 2 deactivations running for the same activation.

@mkanoor For the race condition concern, it is a general problem because we don't have db locking in our code. Even without the scheduler, there is still possibility that two UI sessions request the same deactivation at the same time. Fortunately The deactivate code checks the current status and it will not continue if it is already in stopping. The risk is thus minimal unless we want to add a lock to eliminate any possibility of collision. 
Currently there is one default worker to run the monitoring task. There is no chance of collision. If we increase the number of default workers and if multiple monitoring tasks happen to run in overlapping, as soon as one activation changes the status, it will not be handled by another task. 

For the enhancement of stopping a paused podman container it should be done by another PR.

bzwei:
> @bzwei When the monitor activation task deactivates the deactivation message in the logs look like
> 
> ```
> Start to disable instance 19 ...
> Activation being stopped at user request
> Container 1c879e08a58ab86f2bbd69156578cc2e7979a058bd6028a19df5d9e1e255ec21 is removed.
> 2023-08-25 19:07:41,089 - ansible_rulebook.rule_set_runner - INFO - Task Task-15 finished, active actions 0
> ```
> 
> Should we say the Activation being stopped because it's unresponsive as opposed to Activation being stopped at user request. Since the user didn't do this the monitor task did it

@mkanoor Thanks for finding the misleading message. I simply removed the phrase `at user request`, otherwise we need to pass a boolean parameter multiple times to the function where the message is logged. From the context of the logs we can easily identify whether the deactivate is by a user or by the system.

src/aap_eda/tasks/ruleset.py
```diff
@@ -71,7 +74,10 @@ def activate(activation_id: int, requester: str = "User") -> None:
 
 @job("default")
 def deactivate(
-    activation_id: int, requester: str = "User", delete: bool = False
+    activation_id: int,
+    requester: str = "User",
+    delete: bool = False,
+    final_status: ActivationStatus = ActivationStatus.STOPPED,
```
hsong-rh, line 80: we can use `requester` to set final_status inside the function and pass it to `_perform_deactivate`, no need to pass through from outside, right?

Alex-Izquierdo (commented):
I think @mkanoor has commented something similar. I think the task should have logic to ensure that is not running more than one instance.

src/aap_eda/settings/default.py
```diff
@@ -267,7 +267,7 @@ def _get_secret_key() -> str:
 RQ_QUEUES["activation"]["DB"] = settings.get("MQ_DB", 0)
 
 RQ_STARTUP_JOBS = []
-RQ_PERIODIC_JOBS = []
+RQ_PERIODIC_JOBS = [{"func": "aap_eda.tasks.ruleset.monitor", "interval": 60}]
```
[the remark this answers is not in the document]
bzwei: The minimal interval is 60 seconds when using the rq-scheduler with default setting. I feel it is sufficient. The heartbeat from ansible-rulebook is much slower than 1 minute.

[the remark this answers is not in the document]
Alex-Izquierdo: Yeah, and I also think that the heartbeat is too high hehehe. If the rest is ok with that, np.

src/aap_eda/tasks/ruleset.py
```diff
@@ -190,73 +204,92 @@ def _perform_deactivate(activation: models.Activation) -> None:
...
+        status=ActivationStatus.UNRESPONSIVE.value,
     ):
-        instance = (
-            models.ActivationInstance.objects.filter(activation=activation)
-            .order_by("-started_at")
-            .first()
+        logger.info(
+            f"Deactivate activation {activation.name} due to lost heartbeat"
+        )
+        deactivate(
```
hsong-rh, line 245: There is no `delay` after `deactivate`, did you do this on purpose? run it on scheduler?

bzwei: it is already running in the default worker. No reason to queue itself.

hsong-rh (dismissed):
LGTM

Alex-Izquierdo (changes requested):
When an activation is running, if the server is restarted, the activation state doesn't change and we only relies on the datetime - heartbeat. So, if the system restart takes less time, the activation is not handled. You have an activation that is in running state with no pod actually running. In the case of an upgrade from the AAP installer, this restart would be pretty fast, similar to stop/start the docker-compose.
This can be reproduced easily:

Start the docker-compose
Create a long running activation
wait to run
stop docker-compose
start it again some seconds later. 

The activation should be restored. 

I have not tested it in k8s, but in this case we have to take into account more cases that would fit in the concept of "restart", for example: worker pods are rescheduled or redeployed by k8s (this might happen for multiple reasons, eg: reallocate the pod in another node of the cluster, or an upgrade triggered by the operator or in a manual way)


I think we current strategy is not enough. To rely in the heartbeat its value should be pretty lower, otherwise we won't notice missing activation in many cases. 

What if we store in the db the pod name? that would allow us to establish the relation between the last activation instance and their related pod.

Alex-Izquierdo:
Also I see the following error:
```
eda-default-worker_1     | 2023-08-31 16:16:39,751 INFO     Deactivate activation test due to lost heartbeat
eda-default-worker_1     | 2023-08-31 16:16:39,754 ERROR    [Job fdd0c2de-93d2-4246-993d-7c33be17a041]: exception raised while executing (aap_eda.tasks.ruleset.monitor_activations)
eda-default-worker_1     | Traceback (most recent call last):
eda-default-worker_1     |   File "/app/venv/lib64/python3.9/site-packages/rq/worker.py", line 1359, in perform_job
eda-default-worker_1     |     rv = job.perform()
eda-default-worker_1     |   File "/app/venv/lib64/python3.9/site-packages/rq/job.py", line 1178, in perform
eda-default-worker_1     |     self._result = self._execute()
eda-default-worker_1     |   File "/app/venv/lib64/python3.9/site-packages/rq/job.py", line 1215, in _execute
eda-default-worker_1     |     result = self.func(*self.args, **self.kwargs)
eda-default-worker_1     |   File "/app/src/src/aap_eda/tasks/ruleset.py", line 217, in monitor_activations
eda-default-worker_1     |     _stop_unresponsive(now)
eda-default-worker_1     |   File "/app/src/src/aap_eda/tasks/ruleset.py", line 245, in _stop_unresponsive
eda-default-worker_1     |     deactivate(
eda-default-worker_1     | TypeError: deactivate() got an unexpected keyword argument 'is_delete'

```

src/aap_eda/tasks/ruleset.py
```diff
@@ -190,73 +204,92 @@ def _perform_deactivate(activation: models.Activation) -> None:
...
+            activation_id=activation.id,
+            requester="SCHEDULER",
+            is_delete=False,
```
Alex-Izquierdo, line 248: Wrong parameter name, is  `delete`

bzwei: thanks. this PR has been pending for too long and experienced a lot of rebases

Alex-Izquierdo (approved):
After a couple of basic tests looks it works fine. LGTM

hsong-rh (approved):

PR #339, opened 2023-08-04
[AAP-14656] Improve Project Update error output

Dostonbek1:
- Add validators to Project update serializer to return more clear and concise error output

Solves: [AAP-14636](https://issues.redhat.com/browse/AAP-14636)

### DIFF
--- a/src/aap_eda/api/views/project.py
+++ b/src/aap_eda/api/views/project.py
@@ -204,7 +204,7 @@ class ProjectViewSet(
     def partial_update(self, request, pk):
         project = get_object_or_404(models.Project, pk=pk)
         serializer = serializers.ProjectUpdateRequestSerializer(
-            data=request.data, partial=True
+            instance=project, data=request.data, partial=True
         )
         serializer.is_valid(raise_exception=True)
         credential_id = request.data.get("credential_id")
--- a/src/aap_eda/api/views/project.py
+++ b/src/aap_eda/api/views/project.py
@@ -204,7 +204,7 @@ class ProjectViewSet(
     def partial_update(self, request, pk):
         project = get_object_or_404(models.Project, pk=pk)
         serializer = serializers.ProjectUpdateRequestSerializer(
-            data=request.data
+            data=request.data, partial=True
         )
         serializer.is_valid(raise_exception=True)
         credential_id = request.data.get("credential_id")
--- a/tests/integration/api/test_project.py
+++ b/tests/integration/api/test_project.py
@@ -336,7 +336,7 @@ def test_update_project_not_found(client: APIClient):
 
 
 @pytest.mark.django_db
-def test_update_project_with_409(client: APIClient):
+def test_update_project_with_400(client: APIClient):
     credential = models.Credential.objects.create(
         name="credential", username="me", secret="sec"
     )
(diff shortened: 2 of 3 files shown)


src/aap_eda/api/views/project.py
```diff
@@ -229,11 +229,20 @@ def partial_update(self, request, pk):
...
+            project_name = serializer.validated_data["name"]
+            if not project_name:
+                raise api_exc.Conflict(
+                    detail="Project name cannot be an empty string."
+                )
+            project_name_exists = models.Project.objects.filter(
+                name=project_name
+            ).exists()
+            if project_name_exists:
+                raise api_exc.Conflict(
```
mkanoor, line 242: @Dostonbek1 Wouldn't this be taken care of by this line https://github.com/ansible/eda-server/blob/f7b2c73120dae4344288e74786d842d4049c6bc4/src/aap_eda/core/models/project.py#L51

Dostonbek1: @mkanoor This seems to work only on POST endpoint:
Screenshot 2023-08-07 at 1 22 36 PM

But on PATCH, I get the following:
Screenshot 2023-08-07 at 1 22 45 PM

tests/integration/api/test_project.py
```diff
@@ -357,26 +357,21 @@ def test_update_project_with_409(client: APIClient):
...
-
-@pytest.mark.django_db
-def test_update_project_with_400(client: APIClient):
-    project = models.Project.objects.create(
-        name="test-project-01",
-        url="https://git.example.com/acme/project-01",
-        git_hash="4673c67547cf6fe6a223a9dd49feb1d5f953449c",
+    # test empty string validator
+    response = client.patch(
+        f"{api_url_v1}/projects/{first.id}/", data={"name": ""}
```
mkanoor, line 362: @Dostonbek1 What would happened if the name is the original name in the update? In this case we are setting it to empty.

tests/integration/api/test_project.py
```diff
@@ -357,26 +357,21 @@ def test_update_project_with_409(client: APIClient):
...
+    # test non-existent dependent object reference
+    response = client.get(f"{api_url_v1}/projects/{first.id}/")
     data = {
         "name": "new project name",
         "credential_id": 2,
     }
-    response = client.patch(f"{api_url_v1}/projects/{project.id}/", data=data)
+    response = client.patch(f"{api_url_v1}/projects/{first.id}/", data=data)
     assert response.status_code == status.HTTP_400_BAD_REQUEST
     assert response.data["errors"] == "Credential [2] not found"
```
mkanoor, line 376: @Dostonbek1 Shouldn't this be Project instead of credential

tests/integration/api/test_project.py
Dostonbek1: Fixed it.

tests/integration/api/test_project.py
Dostonbek1: This is testing whether the dependent object ID, credential, exists or not.

tests/integration/api/test_project.py
```diff
@@ -357,26 +357,21 @@ def test_update_project_with_409(client: APIClient):
...
-    project = models.Project.objects.create(
-        name="test-project-01",
-        url="https://git.example.com/acme/project-01",
-        git_hash="4673c67547cf6fe6a223a9dd49feb1d5f953449c",
+    # test empty string validator
+    response = client.patch(
+        f"{api_url_v1}/projects/{first.id}/", data={"name": ""}
     )
-    response = client.get(f"{api_url_v1}/projects/{project.id}/")
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
```
mkanoor, line 364: @Dostonbek1 Can we check if the messages indicate that the value cannot be blank.

Dostonbek1: Done.

mkanoor (approved):

PR #340, opened 2023-08-08
[AAP-13065] Change audit rule's status to failed if one of its action's status failed

hsong-rh:
When a rule has multiple actions, if one of its action `failed`, the rule's status should be `failed` too.

Solve: [AAP-13065](https://issues.redhat.com/browse/AAP-13065)


src/aap_eda/wsapi/consumers.py
```diff
@@ -219,8 +219,12 @@ def insert_audit_rule_data(self, message: ActionMessage) -> None:
             logger.info(f"Audit rule [{audit_rule.name}] is created.")
         else:
             audit_rule.fired_at = message.rule_run_at
-            audit_rule.status = message.status
-            audit_rule.save()
+
+            # if rule has multiple actions and one of its action's status is
+            # 'failed', keep rule's status as 'failed'
+            if audit_rule.status != "failed":
```
mkanoor, line 225: @hsong-rh Should we do this only if the value is different, e.g. 
The first action comes in with status = "successful" and the second  one comes with status="successful"
Should we update the rule record?
```
if audit_rule.status != message.status:
    if audit_rule.status != "failed":
         audit_rule.status = message.status
         audit_rule.save()
```

hsong-rh: @mkanoor I just found we have to always update the rule record because the `fired_at` in the previous line:
`audit_rule.fired_at = message.rule_run_at`

mkanoor: @hsong-rh The rule_fired_at will not change, the actions will get executed at different points in time for a multi action rule but the rule will have the same rule_fired_at.

src/aap_eda/wsapi/consumers.py
```diff
@@ -219,7 +219,15 @@ def insert_audit_rule_data(self, message: ActionMessage) -> None:
...
-            audit_rule.status = message.status
+
+            # if rule has multiple actions and one of its action's status is
+            # 'failed', keep rule's status as 'failed'
+            if (
+                audit_rule.status != message.status
+                and audit_rule.status != "failed"
+            ):
+                audit_rule.status = message.status
+
```
mkanoor, line 230: @hsong-rh Shouldn't the save be inside the if block.

mkanoor (approved):

Dostonbek1 (approved):

PR #341, opened 2023-08-08
[AAP-11738] Fix 500 error on Activation creation

Dostonbek1:
- Fix 500 error thrown on Activation creation when non-existent rulebook ID is passed
- Improve unit test for create activation unprocessible entity
- Add a check for decision environment existence on Activation creation

Solves: [AAP-11738](https://issues.redhat.com/browse/AAP-11738)

### DIFF
--- a/tests/integration/api/test_activation.py
+++ b/tests/integration/api/test_activation.py
@@ -21,7 +21,12 @@ from rest_framework.test import APIClient
 
 from aap_eda.api import serializers
 from aap_eda.core import models
-from aap_eda.core.enums import ActivationStatus, RestartPolicy
+from aap_eda.core.enums import (
+    Action,
+    ActivationStatus,
+    ResourceType,
+    RestartPolicy,
+)
 from tests.integration.constants import api_url_v1
 
 TEST_ACTIVATION = {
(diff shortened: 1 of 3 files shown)


Alex-Izquierdo (commented):
The PR description is included by default in the commit msg so it allow us to determine easily what are the changes included. 
The PR description should be updated to include that the decision environment is also checked, don't?

tests/integration/api/test_activation.py
```diff
@@ -252,24 +252,76 @@ def test_create_activation_bad_entity(client: APIClient):
...
+    assert response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY
+    assert (
+        response.data["detail"]
+        == "Decision environment with ID=0 does not exist."
+    )
+    # test with non-existent rulebook
+    with mock.patch.object(
+        serializers.ActivationCreateSerializer,
+        "create",
+        mock.Mock(side_effect=IntegrityError),
```
bzwei, line 284: This test is only for testing whether a meaningful response is created on error. Shouldn't you add a real test for non-existing rulebook id without a mock?

Alex-Izquierdo (approved):
Good point using parameterization in tests!

Alex-Izquierdo (dismissed):

Alex-Izquierdo (approved):

PR #344, opened 2023-08-09
[AAP-13350] Update filtering for Activation Instance Logs

Dostonbek1:
* Remove unused filters from activation instance logs
  * Add 'log' filtering for activation instance logs

### DIFF
--- a/src/aap_eda/api/filters/__init__.py
+++ b/src/aap_eda/api/filters/__init__.py
@@ -12,7 +12,11 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from .activation import ActivationFilter, ActivationInstanceFilter
+from .activation import (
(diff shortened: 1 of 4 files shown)


hsong-rh (approved):

hsong-rh:
Should we also change the `lookup_expr` of Activation and ActivationInstance to `icontains`? Now they are set to `istartswith`, I feel they should in sync and `icontains` is more suitable.

Dostonbek1:
> Should we also change the `lookup_expr` of Activation and ActivationInstance to `icontains`? Now they are set to `istartswith`, I feel they should in sync and `icontains` is more suitable.

Not sure if there has been any discussion on which one to use but I was following what we already have for `Activations`.

bzwei (approved):

PR #348, opened 2023-08-10
[AAP-14635] Have 2 activation workers in minikube env

bzwei:
This is a follow-up for #330 

Docker-compose env has 2 activation worker be default. Minikube env should have the same.

Remove service account for activation worker since we can use the service account for the default worker.

Some typo fixes.

Fixes AAP-14635: Correct worker class name in deployment script


Alex-Izquierdo (approved):
LGTM

hsong-rh:
We had discussed in https://github.com/ansible/eda-server/pull/330 that we need roles settings for both activation and default workers

bzwei:
> We had discussed in #330 that we need roles settings for both activation and default workers

@hsong-rh This work does not remove the service account requirement for activation worker. It uses the same SA from default worker. This SA has all roles for both workers.

PR #351, opened 2023-08-10
[AAP-14982] Set default_timeout to -1 (no timeout) on queue class

hsong-rh:
Solves [AAP-14982](https://issues.redhat.com/browse/AAP-14982)

### DIFF
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -255,10 +255,12 @@ else:
         "default": {
             "HOST": settings.get("MQ_HOST", "localhost"),
             "PORT": settings.get("MQ_PORT", 6379),
+            "DEFAULT_TIMEOUT": -1,
         },
         "activation": {
             "HOST": settings.get("MQ_HOST", "localhost"),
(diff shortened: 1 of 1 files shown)


mkanoor:
@Alex-Izquierdo this is related to changes implemented in https://github.com/ansible/eda-server/pull/335

mkanoor (approved):

Alex-Izquierdo (approved):
Do we really want no-timeout on the default one? 
Otherwise lgtm

mkanoor:
@Alex-Izquierdo We can do that in a separate PR we need to decide what should that default time be. @benthomasson wanted to add a time limit to the default worker, maybe its 5 minutes to complete a project sync. But there is more to that since we might be either stuck or just gotten a network connection back and started writing the rulebooks into the DB when the task is aborted leaving partial/incomplete rulebooks in the DB. We might have to catch that error and remove all the partial rulebooks and flag it as a failed project sync. There is also the ability to do this on a task by task basis. e.g
@job(queue=default,timeout=30) which will be more granular per task

Alex-Izquierdo:
Totally agree with granular timeouts.

PR #353, opened 2023-08-11
Fix migration to keep old activations status

hsong-rh:
After adding `status` to Activation, we have to populate them with the original data, rather than the default `pending`


Alex-Izquierdo (approved):
LGTM

bzwei (approved):

PR #355, opened 2023-08-14
[AAP-13416] Move activation deletion to default worker

hsong-rh:
This PR refactored the activation deletion logic as follows:
1. marks the activation's status to `deleting` in api worker, so UI can display it with correct status;
2. disable the activation to cleanup the jobs in queue and the running containers if necessary in default worker;
3. delete the activation from DB in default worker;

Solves [AAP-13416](https://issues.redhat.com/browse/AAP-13416)

### DIFF
--- a/src/aap_eda/api/views/activation.py
+++ b/src/aap_eda/api/views/activation.py
@@ -175,12 +175,11 @@ class ActivationViewSet(
         return self.get_paginated_response(serializer.data)
 
     def perform_destroy(self, activation):
-        instances = models.ActivationInstance.objects.filter(
-            activation_id=activation.id
-        )
-        for instance in instances:
-            deactivate_rulesets.delay(instance.id, settings.DEPLOYMENT_TYPE)
-        super().perform_destroy(activation)
+        activation.status = ActivationStatus.DELETING
+        activation.save(update_fields=["status"])
+        logger.info(f"Now deleting {activation.name} ...")
+
+        deactivate.delay(activation_id=activation.id, is_delete=True)
 
     @extend_schema(
         description="List all instances for the Activation",
@@ -249,6 +248,7 @@ class ActivationViewSet(
         if activation.status in [
             ActivationStatus.STARTING,
             ActivationStatus.STOPPING,
+            ActivationStatus.DELETING,
             ActivationStatus.PENDING,
             ActivationStatus.RUNNING,
             ActivationStatus.UNRESPONSIVE,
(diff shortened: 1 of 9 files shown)


bzwei (dismissed):

hsong-rh:
To make this PR simple, I will create a separate one to focus on test cases.

src/aap_eda/api/views/activation.py
```diff
@@ -320,6 +323,10 @@ def disable(self, request, pk):
     @action(methods=["post"], detail=True, rbac_action=Action.RESTART)
     def restart(self, request, pk):
         activation = get_object_or_404(models.Activation, pk=pk)
+
+        if str(activation.status) == ActivationStatus.DELETING.value:
```
mkanoor, line 327: @hsong-rh Would it make sense to make a small function called _check_deleting and raise a 409 exception if its in deleting status.

```
def check_deleting(activation):
    if str(activation.status) == ActivationStatus.DELETING.value:
       raise ApiException(detail="Object is being deleted",code=409)
```
 https://github.com/encode/django-rest-framework/blob/5c07060ce0dcbdad1e38be8a509f5c23b6e9fc63/rest_framework/exceptions.py#L108

src/aap_eda/services/ruleset/activate_rulesets.py
```diff
@@ -75,26 +80,37 @@ def find_ports(rulebook_text: str):
...
+            ActivationStatus.STARTING.value,
+            ActivationStatus.RUNNING.value,
+            ActivationStatus.UNRESPONSIVE.value,
+        ]
+        activation_fields = ["status", "modified_at"]
+        if str(instance.status) not in running_states:
+            instance.activation.current_job_id = None
+            activation_fields.append("current_job_id")
+        if str(instance.status) == ActivationStatus.COMPLETED.value:
+            instance.activation.failure_count = 0
```
mkanoor, line 104: @hsong-rh Shouldn't this be preserved to record the number of times the activation has failed. If in the future we want to display the failed lists of activation we could have a link that shows all the failed activations.

src/aap_eda/services/ruleset/activate_rulesets.py
```diff
@@ -167,11 +187,12 @@ def activate(
                     activation_db_logger,
                 )
 
+            activation_db_logger.flush()
         except DeactivationException:
             msg = f"Activation {activation.name} is disabled"
-            instance.status = ActivationStatus.STOPPED
```
mkanoor, line 172: @hsong-rh Shouldn't we have a return here since the finally will run and we suppress the exception there, but if the activation worker is running ahead of the default worker they would both write to the DB.

src/aap_eda/services/ruleset/activate_rulesets.py
hsong-rh: We may discuss this in dev sync meeting, the current logic is to reset it to 0 whenever the activation is complete.

mkanoor (approved):
Tesed looks good

Dostonbek1 (approved):

PR #357, opened 2023-08-14
Update statues of activation and instance together

bzwei:
Follow up  #316 
Refactor: extract common code into a function. Reduce the number of DB saves where possible.

### DIFF
--- a/src/aap_eda/services/ruleset/activate_rulesets.py
+++ b/src/aap_eda/services/ruleset/activate_rulesets.py
@@ -74,6 +74,29 @@ def find_ports(rulebook_text: str):
     return found_ports
 
 
+def save_activation_and_instance(
+    instance: models.ActivationInstance, update_fields: list
+):
+    """Save instance and update the linked activation's status accordingly."""
+    instance.activation.status = instance.status
+    running_states = [
+        ActivationStatus.PENDING.value,
+        ActivationStatus.STARTING.value,
+        ActivationStatus.RUNNING.value,
+        ActivationStatus.UNRESPONSIVE.value,
+    ]
+    activation_fields = ["status", "modified_at"]
+    if str(instance.status) not in running_states:
+        instance.activation.current_job_id = None
+        activation_fields.append("current_job_id")
+    if str(instance.status) == ActivationStatus.COMPLETED.value:
+        instance.activation.failure_count = 0
+        activation_fields.append("failure_count")
+
+    instance.save(update_fields=update_fields)
+    instance.activation.save(update_fields=activation_fields)
+
+
 class ActivateRulesets:
     def __init__(self, cwd: Optional[str] = None):
         self.service = AnsibleRulebookService(cwd)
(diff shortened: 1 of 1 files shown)


Dostonbek1 (approved):

src/aap_eda/services/ruleset/activate_rulesets.py
```diff
@@ -74,6 +74,29 @@ def find_ports(rulebook_text: str):
     return found_ports
 
 
+def save_activation_and_instance(
```
hsong-rh, line 77: This function is kind of specific, should we define it as class method?

src/aap_eda/services/ruleset/activate_rulesets.py
```diff
@@ -74,6 +74,29 @@ def find_ports(rulebook_text: str):
...
+    ]
+    activation_fields = ["status", "modified_at"]
+    if str(instance.status) not in running_states:
+        instance.activation.current_job_id = None
+        activation_fields.append("current_job_id")
+    if str(instance.status) == ActivationStatus.COMPLETED.value:
+        instance.activation.failure_count = 0
+        activation_fields.append("failure_count")
+
+    instance.save(update_fields=update_fields)
```
hsong-rh, line 96: This part needs to rework when we consider the activation deletion (https://github.com/ansible/eda-server/pull/355). All DB access during the process of activation may fail due to unavailable activations and its instances.

mkanoor: @bzwei When a deactivation happens its a race condition between the default-worker which has terminated the pod and the activation worker which catches the fact that the activation has ended. They both go about updating the database instance and depending on which process gets the cycle and who updates it last their results will get stored. I think we need to divide the work so that 
1. Activation Worker only updates the Activation/Activation Instance when a normal termination happens
2. The Default Worker updates the Activation and Activation Instance when its deactivating or deleting an activation

Without this separation there will always be a race condition between the 2 processes vying to update the Activation and ActivationInstance.

There is an exception that is thrown in the ActivationWorker when a deactivation is going on, in those cases the Activation Worker should log a message and bail and let the default worker do all the cleanup and setting the status. This has been fixed in some sort for k8s but doesn't exist in podman activation

src/aap_eda/services/ruleset/activate_rulesets.py
bzwei: I plan to use it from tasks.ruleset too.

src/aap_eda/services/ruleset/activate_rulesets.py
bzwei: I understand the problem. They need to be fixed or resolve conflicts in the order of PR merging.

hsong-rh: > This has been fixed in some sort for k8s but doesn't exist in podman activation
@mkanoor k8s also has the problem. It only raise the flag in a corner case.

hsong-rh (approved):
Approved, further work will be addressed in https://github.com/ansible/eda-server/pull/355

mkanoor (approved):

PR #362, opened 2023-08-16
fix: handle multiple rule matching to the same fact/event

mkanoor:
When the same fact/event matches multiple rules we were not storing the events for the rules other than the first one. This PR ensures that the events are stored for every rule that matches.
https://issues.redhat.com/browse/AAP-15158


hsong-rh (approved):
Tested, worked properly.

Dostonbek1 (approved):

PR #367, opened 2023-08-21
Active entities

jshimkus-rh:
This is presented as a topic for discussion of transitioning EDA objects (in particular model objects) from passive (acted upon) to active (performing the required changes) roles.

The first commit is a simple modification whereby specific instances are retrieved via the model class rather than  an external operation executed against the model class.

The second commit incorporates the transition of an `Activation` to `STOPPING` via a method implemented on `Activation`.  This isolates the processing to the class itself reducing the coupling between the view and the model.  As the decoupling of the state changes to `Activation` is the greater motivation the change does not attempt to incorporate the  `deactivate.delay` within the added `Activation.stop` method to avoid the greater restructuring that would be necessary to prevent a circular import issue.

### DIFF
--- a/src/aap_eda/api/views/activation.py
+++ b/src/aap_eda/api/views/activation.py
@@ -172,8 +172,7 @@ class ActivationViewSet(
         return self.get_paginated_response(serializer.data)
 
     def perform_destroy(self, activation):
-        activation.status = ActivationStatus.DELETING
-        activation.save(update_fields=["status"])
+        activation.destroying()
         logger.info(f"Now deleting {activation.name} ...")
 
         deactivate.delay(activation_id=activation.id, is_delete=True)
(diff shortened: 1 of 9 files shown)


Alex-Izquierdo:
What is the purpose/benefit of calling object_or_404 from the model? 
I would say the model should not handle any logic related with views, it looks like a violation of the clear separation of concerns that Django emphasizes.

jshimkus-rh:
A reasonable question/concern.  In this case it shouldn't necessarily be taken as a literal proposal as opposed to a simplistic change, maintaining current functionality, whose important point is related to the consolidation of object management via the class; a separation of implementation from that of usage.

Arguably, in moving to an active entity approach for this specific use case it would be "purer" to have the instance look up return `None` (or raise some variant of `LookupError`, `IndexError` or `KeyError`) and the caller would respond to that.

PR #371, opened 2023-08-22
Refactor and improve consumer tests

Alex-Izquierdo:
I have several issues with consumer tests related with timeouts and false negatives because the process has no enough cycles to complete DB ops. This PR set a timeout for some receive operations and waits to complete the ops before the assertions. It also moves the instantiation of the websocket communicator to a fixture to follow good pytest patterns and ensure the connection is closed in the teardown.


Dostonbek1 (approved):
lgtm

PR #373, opened 2023-08-24
[AAP-13416] Fix exception handling in OCP when activation is disabled or deleted

hsong-rh:
This PR follows https://github.com/ansible/eda-server/pull/355 to solve issues found in the OCP environment when activation is created, deleted or disabled:
1. Handle the exceptions caused by activation deletion or disabling correctly;
2. When activation is in progress, fix its status from `starting` to `running`, in sync with podman env;
3. Add missing `current_job_id` in activation serializers so it can appear in API calls;


src/aap_eda/api/views/activation.py
```diff
@@ -115,11 +115,13 @@ def create(self, request):
         activation["rules_fired_count"] = 0
 
         if response.is_enabled:
-            activate_rulesets.delay(
+            job = activate_rulesets.delay(
                 is_restart=False,
                 activation_id=response.id,
                 deployment_type=settings.DEPLOYMENT_TYPE,
             )
+            response.current_job_id = job.id
```
mkanoor, line 123: @hsong-rh Is this current_job_id the same as the RQ Job ID, if we have this ID could we retrieve the Job Data from the RQ Job?

mkanoor:
@hsong-rh Do we need activation logs at the beginning like we do with podman https://github.com/ansible/eda-server/blob/bae76cc26cc380b49b9d5bcf935f3cbec8724ef7/src/aap_eda/services/ruleset/activation_podman.py#L127 otherwise there is no log line about what is happening with activation till the image is downloaded and we start running the ansible-rulebook. Is the "Running" state set based on when we start seeing the Session Stats come on the websocket (Daphne) channel.

hsong-rh:
> @hsong-rh Do we need activation logs at the beginning like we do with podman
> 
> https://github.com/ansible/eda-server/blob/bae76cc26cc380b49b9d5bcf935f3cbec8724ef7/src/aap_eda/services/ruleset/activation_podman.py#L127
> 
> otherwise there is no log line about what is happening with activation till the image is downloaded and we start running the ansible-rulebook. 

I totally agree K8S should be similar to podman and add more helpful information in activation instance logs, but it's better to open a separate jira ticket. 

> Is the "Running" state set based on when we start seeing the Session Stats come on the websocket (Daphne) channel.

No, this `Running` state is for the activation, the session stats are for the instance.

src/aap_eda/services/ruleset/activate_rulesets.py
```diff
@@ -375,6 +375,9 @@ def activate_in_k8s(
         job_name = f"activation-job-{activation_id}"
         pod_name = f"activation-pod-{activation_id}"
 
+        activation_instance.status = ActivationStatus.RUNNING
```
bzwei, line 378: It is not proper to set the RUNNING status this early nor at the end. The RUNNING status is already set at activation_kubernetes.py by watch_job_pod. It makes sense to change the activation status to RUNNING only when k8s detects pod in running state.

hsong-rh: @bzwei `activation_kubernetes.py` only set instance's status, not the activation.

bzwei: we should call `save_activation_and_instance` from there

hsong-rh: That will cause circular import errors. Will call it later when refactor k8s.

bzwei (approved):

mkanoor (approved):

PR #370, opened 2023-08-22
Refactor: remove unused deployment local and docker types

hsong-rh:
Jira: https://issues.redhat.com/browse/AAP-14892

PR #374, opened 2023-08-25
Refactor: remove deployment_type from function params

hsong-rh:


tests/integration/services/test_ruleset_activate.py
```diff
@@ -144,24 +145,10 @@ def get_activation_stats():
     }
 
 
-@pytest.mark.django_db
```
jshimkus-rh, line 147: Would this still be useful to have as a test using the default the restart policy?

tests/integration/services/test_ruleset_activate.py
```diff
@@ -144,24 +145,10 @@ def get_activation_stats():
...
-    ActivateRulesets().activate(
-        activation=init_data.activation, deployment_type="bad_type"
-    )
-
-    assert (
-        models.ActivationInstance.objects.first().status
-        == ActivationStatus.FAILED.value
-    )
-
-
```
jshimkus-rh, line 161: Would this still be useful to have as a test using the default the restart policy?

hsong-rh: `subprocess.run` is used by the deployment type of `local`, which had been removed by the previous refactor

src/aap_eda/settings/default.py
```diff
@@ -346,7 +346,7 @@ def _get_secret_key() -> str:
 # DEPLOYMENT SETTINGS
 # ---------------------------------------------------------
 
-DEPLOYMENT_TYPE = settings.get("DEPLOYMENT_TYPE", "local")
+DEPLOYMENT_TYPE = settings.get("DEPLOYMENT_TYPE", "podman")
```
mkanoor, line 349: @hsong-rh Are we going to remove the local deployment with this PR or is that separate, since we are setting the default to be "podman" with this PR?

jshimkus-rh (approved):

src/aap_eda/settings/default.py
hsong-rh: @mkanoor The `local` deployment had been removed by https://github.com/ansible/eda-server/pull/370

mkanoor (approved):

PR #375, opened 2023-08-28
[AAP-13112] Add git_hash field to Activation Instances

Dostonbek1:
## Problem

When the activation is created without a project, the git hash doesn't appear in the activation detail view. 

## Solution 

- Adds a new field `git_hash` to Activation Instances that stores the project's git hash that the rulebook is related to, which is populated at the time of activating the rulebook and instance creation, using the `rulebook` field, thus it no longer depends on the `project` field and still retains the git hash value even if the project is deleted later. 
- The field is set to _null_ if the rulebook is originally not part of any project. 
- Remove project_id field from POST Activations request payload

**NOTE:** Depends on https://github.com/ansible/ansible-ui/pull/830


mkanoor:
@Dostonbek1 Should't a project be a mandatory field when creating an activation?

Dostonbek1:
> @Dostonbek1 Should't a project be a mandatory field when creating an activation?

@mkanoor We don't really use the `project` value in Activations, @Alex-Izquierdo was suggesting we might as well remove it from POST Activation.

mkanoor:
@Dostonbek1 Yes makes sense, since from the rulebook we should be able to get to the project.

jshimkus-rh (dismissed):

src/aap_eda/api/serializers/activation.py
```diff
@@ -121,6 +122,8 @@ def create(self, validated_data, user):
             rulebook = models.Rulebook.objects.get(
                 pk=validated_data["rulebook_id"]
             )
+            if rulebook.project:
+                validated_data["git_hash"] = rulebook.project.git_hash
```
bzwei, line 126: Should we store a static copy of git_hash at creation time? It makes sense only if we also store a static copy of rulebook, but here rulebook is only a reference. Its git_hash can change over time. 

If we want to continue the same behavior as before by fetching live from project's git_hash, we should not create the git_hash field in model, but rather add the field to the serializer only.

Dostonbek1: @bzwei  As far as I know, Activations capture a snapshot of the rulebook at the time of the Activation creation (in `rulebook_rulesets` field) and uses the value from that field to send to ansible-rulebook to activate the rules/jobs. Ben had also confirmed before, that Activations are immutable, thus we removed PATCH endpoints for them. Therefore, I don't think it would make sense for the git hash for an existing rulebook activation to change over time.

bzwei (dismissed):

src/aap_eda/api/serializers/activation.py
Alex-Izquierdo: @Dostonbek1 I think @bzwei is right. Activations are no longer inmutable. We decided to update them if the project changes. 
https://github.com/ansible/eda-server/blob/main/src/aap_eda/services/project/imports.py#L171

bzwei: @Alex-Izquierdo We made a copy of the rulebook_rulesets at activation creation time. So even if the project resync later and its content changes, we will still use the old rulebook for existing activations. You'll need to create a new activation (not just instances) to use the updated rulebook.

mkanoor: @Dostonbek1 @Alex-Izquierdo @bzwei I think there is another issue here if the user updates the rulebooks repo with a project sync between activations we would be running different rulebooks for each activation instance. I think instead of storing the githash in the activation it should be stored in the activation instance so that we can record the githash for each activation run moreover it also helps when the customer says it was running fine till yesterday now its broken the githash in the activation instance will record the hash just before we start an activation.

In the image below all these activation instances ran with different playbook because of the project sync running before each one 
Screenshot 2023-08-28 at 4 06 26 PM

If we had the githash for each of these runs we could say that it was a different hash for the repo.

Even that doesn't help when the rule book remains the same but some other thing changes in the repo and we put up different hash for the repo. Instead I think we need to hash the rulebook and use that instead of the git repo hash. The git repo hash allows us to see the files based on the commit sha, but doesn't tell us the specifics of the rulebook anchored into the activation.

Alex-Izquierdo: I totally agree with @mkanoor that the hash must be in the instance because it might change between instances. The activation should show the hash from the latest instance. 

But creating a hash of the rulebook doesn't add any value to the user. What the user wants is to know which version of the rulebook was used for a given activation. If the hash has changed but the rulebook don't, that's fine. The user is equally able to query their git history to see what was the rulebook.

bzwei: @Alex-Izquierdo Sorry. I did not look carefully at the link you posted. You're right, the rulebook content is copied to activation after the resync. So we should copy the githash also at the same place if we want to store it in the activation.

It makes sense to save the githash in instances. If we do so shouldn't we store the ruleset content in instances also? We should not need to save either githash or ruleset content at activation level.

mkanoor: In this PR https://github.com/ansible/eda-server/pull/196 we update the rulebook directly in the activation so every new run of an activation uses the [latest data](https://github.com/ansible/eda-server/blob/249bd8e586a69bf4b601b4917db8ed329f867ad2/src/aap_eda/services/project/imports.py#L171). So we are always running with the latest rulebook, I think this should be optional if we update the rulebook without letting the users know they might get surprised by which rulebook is getting used.

mkanoor:
If we compare this to how the Tower/Controller wraps a playbook inside a Job Template a project is a required parameter.
Screenshot 2023-08-28 at 4 57 05 PM
So from a look and feel perspective it might be better to be consistent.

src/aap_eda/api/serializers/activation.py
Alex-Izquierdo: I think the most common expectation is to work always with the latest version (awx does the same).

mkanoor:
Another use case that comes to mind is if we detect that a rulebook has been updated and we have an running activation for that rulebook we could optionally stop and restart the activation with the new updated rulebook. We have all the pieces in place like restart, detecting if its newer and queue a restart job to the scheduler.

src/aap_eda/services/ruleset/activate_rulesets.py
```diff
@@ -113,10 +113,30 @@ def activate(self, activation: models.Activation) -> None:
         ssl_verify = settings.WEBSOCKET_SSL_VERIFY
         try:
             try:
+                project_hash = None
```
mkanoor, line 116: @Dostonbek1 With the current implementation when every rulebook has a project couldn't we use this all the time
```
activation.rulebook.project.git_hash
```
This would have the latest hash and it would get saved into the new ActivationInstance we create.

This works even when you have don't have the project is None since we are getting the data from the rulebook project relationship.

Dostonbek1: We have to have this check here for when we delete the project after creating an activation. Although the project is deleted, rulebook content is still saved into activation and activation can be restarted. In that case, rulebook will be null.

mkanoor: @Dostonbek1 I think it might make sense to store the git_hash at both the Activation and ActivationInstance level. The Activation git_hash can be updated with every project_sync by updating it [here](https://github.com/ansible/eda-server/blob/249bd8e586a69bf4b601b4917db8ed329f867ad2/src/aap_eda/services/project/imports.py#L171)
The ActivationInstance git_hash will be copied from the Activation git_hash when we run a new activation. This way even if the project gets deleted we would have the last git_hash in the Activation object. The 2 git hashes have to be displayed in 2 different screens.
The Activation git_hash should be visible here
Screenshot 2023-08-29 at 3 43 39 PM

And the Activation Instance git_hash should be available for every activation run here
Screenshot 2023-08-29 at 3 44 10 PM

Alex-Izquierdo: @mkanoor I think the activation hash should be the hash from the latest instance if exists. We should take into account the case where we have an instance running with hash B, but project was updated later and the activation is showing the  hash C, which would be incorrect. So sync_project task should update the hash only on the activations that don't have any instance running. What do you think?

mkanoor: @Alex-Izquierdo I am not sure I follow, the Activation.git_hash = <latest project hash>
ActivationInstance.git_hash = Activation.git_hash
These are not references they are the real values, the Activation gets the value from project_sync and the ActivationInstances gets from Activation just before a new Activation starts. They can deviate after this but each one has the git_hash value of their own. We are always copying the value from the Activation into the ActivationInstance. 

In this case both the ActivationInstance and Activation have an attribute called git_hash. The ActivationInstance.git_hash will never be updated after it has been created. The Activation.git_hash might keep changing with each project sync if there are differences.

Dostonbek1:
> I'm worried, are we okay with allowing the hash to be empty? I feel the hash should be there always or at least have a default string like "project deleted".
> 
> I would like to know more opinions about this before approve it.
> 
> On the other hand, I'm not sure if I missing something, but it looks like the removal of the project_id is not necessary for this implementation, if that is the case, I would prefer to do it in a separate PR for a cleaner history.

it looks to me that with the current implementation, there is no way git hash for Activation could be empty since the only way we can have rulebooks is through project imports. So I think it is safe to set git hash as mandatory unless implementation changes later in the future. 

As for removing `project_id` from POST Activation request payload, I will put it into a separate PR.

src/aap_eda/services/ruleset/activate_rulesets.py
Alex-Izquierdo: I mean that activation.git_hash should have the value of the latest activation_instance.git_hash if the instances exists. 
If we update the activation.git_hash from sync_project, it might happen that the activation.git_hash is different from the latest activation_instance.git_hash. That case must be taken into account.

mkanoor: The activation.git_hash will have the latest information from the project, so even if the project is deleted after that sync we have captured the latest hash. The ActivationInstance.git_hash tells you which hash was used for that activation run. If the Activation.git_hash != ActivationInstance.git_ash (latest run) then we know that we are running with an old rulebook. If we set Activation.git_hash = ActivationInstance.git_ash (latest run) then we don't have the latest project hash. The 2 screens above convey different things we would have to add a column to the History to display the git_hash for the Activation Instance. The git_hash in the Activation Instance is what the latest value is. If you want to capture what was the last run git_hash we would have to then add
* Activation.git_hash
* Activation.last_run_git_hash
* ActivationInstance.git_hash

I think the last_run_git_hash is redundant since you can get that from the latest instance and the UI can display that in the History as an additional column and be color coded if its different from the Activation.git_hash so you could spot that there is something odd with that ActivationInstance.

mkanoor: Another thing should this git_hashes be clickable/link and they would show you the tree in the git repo which has the corresponding hash.

Alex-Izquierdo: Your point about `Activation.git_hash != ActivationInstance.git_hash` is valid but I think the user expectation is to see for an activation that is running the hash of the latest instance.

Alex-Izquierdo: > Another thing should this git_hashes be clickable/link and they would show you the tree in the git repo which has the corresponding hash.

@mkanoor Is this a question or affirmation? The UI mock currently is defined only to provide the string. What you propose would be a new feature. We should need to know in advance what would be the url, it depends on the git server (gitea, github, gitlab...). AWX also provides only the string.

Dostonbek1: FYI, JIRA ticket for UI changes that Madhu mentioned: [AAP-15544](https://issues.redhat.com/browse/AAP-15544)

src/aap_eda/services/project/imports.py
```diff
@@ -163,13 +163,17 @@ def _sync_rulebook(
         rulebook_info: RulebookInfo,
     ):
         if rulebook.rulesets == rulebook_info.raw_content:
+            models.Activation.objects.filter(rulebook=rulebook).update(
```
bzwei, line 166: Why do we need this line given it has been set by line 176? If it is for migration purpose, we should do it in the migration file.

Dostonbek1: This is to make sure that we update the git hash for rulebook activations, even if the rulebook has not been changed but the project overall has changed.

bzwei: OK. You still need to enhance the migration file to set githash for existing activations.

Alex-Izquierdo, line 166: Minor, optimization: you can save DB queries if you store the git_hash in a variable for the later updates.

src/aap_eda/services/project/imports.py
```diff
@@ -163,13 +163,17 @@ def _sync_rulebook(
...
+                git_hash=rulebook.project.git_hash,
```
Alex-Izquierdo, line 167: Minor, optimization: you can save DB queries if you store the git_hash in a variable for the updates.

Alex-Izquierdo (approved):

src/aap_eda/core/migrations/0009_add_git_hash_to_activations.py
```diff
@@ -0,0 +1,34 @@
...
+
+def add_git_hash_to_activation(apps, schema_editor):
+    Activation = apps.get_model("core", "Activation")  # noqa: N806
+
+    for activation in Activation.objects.all():
+        git_hash = "project deleted"
+        if activation.rulebook and activation.rulebook.project:
+            git_hash = activation.rulebook.project.git_hash
+        activation.git_hash = git_hash
+        activation.save(update_fields=["git_hash"])
```
mkanoor, line 14: @Dostonbek1 do we need this for activation_instance.git_hash too?

Alex-Izquierdo: there is no way no know what was the hash for the instance

bzwei: This could be inaccurate. IMHO Leaving it blank might be better than an inaccurate value

Dostonbek1: @bzwei agreed.

mkanoor (approved):

PR #376, opened 2023-08-28
[AAP-15287] Add transaction atomic to solve race conditions when restarting activation

hsong-rh:
Sovles: [AAP-15287](https://issues.redhat.com/browse/AAP-15287)

### DIFF
--- a/src/aap_eda/tasks/ruleset.py
+++ b/src/aap_eda/tasks/ruleset.py
@@ -16,6 +16,7 @@ import logging
 from datetime import datetime, timedelta
 
 from django.conf import settings
+from django.db import transaction
 from django_rq import get_scheduler
 from rq.job import JobStatus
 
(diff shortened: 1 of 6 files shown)


src/aap_eda/tasks/ruleset.py
```diff
@@ -55,8 +52,47 @@ def activate_rulesets(is_restart: bool, activation_id: int) -> None:
...
+            activation.delete()
+        else:
+            activation.current_job_id = None
+            activation.status = ActivationStatus.STOPPED
+            activation.save()
+
+
+@job("default")
+def restart(activation_id: int, requester: str = "User") -> None:
+    logger.info(f"{requester} asks to restart activation {activation_id} ...")
```
mkanoor, line 80: @hsong-rh This log might be better as
```
  logger.info(f"Restarting activation id: {activation_id} requested by {requester}")
```

src/aap_eda/tasks/ruleset.py
```diff
@@ -29,7 +30,7 @@
 
 
 @job("activation")
-def activate_rulesets(is_restart: bool, activation_id: int) -> None:
+def activate(activation_id: int) -> None:
```
jshimkus-rh, line 33: Activation can be implicit as a result of other actions (e.g., create), but wouldn't it be reasonable to have this be consistent with the other methods in having a `requester` parameter?

jshimkus-rh: @mkanoor I figured as much.  It's just that it seems odd to have what appear to be similar actions with different interfaces.  I'd be comfortable enough if it was commented as to why it's different.

src/aap_eda/tasks/ruleset.py
```diff
@@ -38,10 +39,6 @@ def activate_rulesets(is_restart: bool, activation_id: int) -> None:
         logger.info(f"Activation id: {activation_id} is deleted")
         return
 
-    # TODO(hsong): move this to views
-    if is_restart:
```
mkanoor, line 42: @hsong-rh If the scheduler does a restart because the activation is unresponsive how would this increment work if its in the view?

src/aap_eda/tasks/ruleset.py
mkanoor: @jshimkus-rh You are correct, we could have activation either started by the User using the UI or the scheduler task when it sees that the Activation is non responsive. The requester is used in log messages to see who started the process was it a scheduler or the user. @hsong-rh we should add a requester field in this function so the log messages are clearer about who initiated the process (activate/disable/delete). I think delete is the only task that is done by the end user, but the restart and disable could be done by both the scheduler and the user.

src/aap_eda/tasks/ruleset.py
hsong-rh: As I known, the restart counter is only increased by manually clicking the restart button

Alex-Izquierdo: I don't think so, it is the contrary, the counter is increased for every automatic restart. It is a way to indicate to the user that something could go wrong.

Alex-Izquierdo: More context: https://github.com/ansible/eda-qa/blob/main/eda_qa/tests/api/test_activations.py#L286

hsong-rh: @Alex-Izquierdo https://github.com/ansible/eda-server/blob/main/src/aap_eda/api/views/activation.py#L340 so this increasing counter is not needed?

src/aap_eda/tasks/ruleset.py
hsong-rh: @jshimkus-rh @mkanoor the `requester` is added to `activate` too

src/aap_eda/tasks/ruleset.py
Alex-Izquierdo: @hsong-rh I think that is fine (has been there for five months). What I say that the counter should be increased as well with automatic restarts.

hsong-rh: @Alex-Izquierdo I remembered we only increase the counter in one side. Anyway I will move back the restart counter, thanks!

src/aap_eda/tasks/ruleset.py
```diff
@@ -38,11 +39,14 @@ def activate_rulesets(is_restart: bool, activation_id: int) -> None:
...
+    # Note: expect scheduler to set requester as "SCHEDULER"
+    if requester == "SCHEDULER":
```
bzwei, line 43: Move this block after line 51.

src/aap_eda/tasks/ruleset.py
```diff
@@ -29,34 +30,98 @@
...
+        ):
+            logger.info(f"Activation id: {activation_id} is deleted")
+            return
 
-    # TODO(hsong): move this to views
-    if is_restart:
-        activation.restart_count += 1
-        activation.save(update_fields=["restart_count", "modified_at"])
-    logger.info(f"Task started: Activate rulesets ({activation.name})")
+        logger.info(
```
bzwei, line 47: Before this line we should validate activation's `is_enabled` is True and status is `pending`, otherwise log and return.

bzwei, line 47: We should log this line at the beginning just like `restart` or `deactivate`.

src/aap_eda/tasks/ruleset.py
```diff
@@ -29,34 +30,98 @@
...
+        activation = (
+            models.Activation.objects.select_for_update()
+            .filter(id=activation_id)
+            .first()
+        )
+        if not activation:
+            logger.warning(f"Activation {activation_id} is deleted.")
+            return
+
+        _perform_deactivate(activation)
```
bzwei, line 91: Before this line we should validate activation status is in pending, running, starting, or unresponsive. Otherwise log and return.

src/aap_eda/tasks/ruleset.py
```diff
@@ -29,34 +30,98 @@
...
+
+        queue = get_queue(name="activation")
```
bzwei, line 120: Before this line we should validate the activation is restartable, otherwise don't bother to add it to the queue. It make sense to change the status to pending before queuing so the activate function only need to validate the activation is in pending status.

src/aap_eda/tasks/ruleset.py
```diff
@@ -29,34 +30,125 @@
...
+            logger.warning(
+                f"Cannot restart the activation {activation.name} when its "
+                f"status is {activation.status}"
+            )
+            return
+
+        activation.status = ActivationStatus.PENDING
+        activation.save(update_fields=["status", "modified_at"])
+
+        queue = get_queue(name="activation")
```
bzwei, line 147: Shouldn't we call `activate.delay(activation_id, requester)`? It aligns when how this method is called in other places, and preserves the requester.

src/aap_eda/tasks/ruleset.py
```diff
@@ -29,34 +30,125 @@
...
+        activation = (
+            models.Activation.objects.select_for_update()
+            .filter(id=activation_id)
+            .first()
+        )
+        if not activation:
+            logger.warning(f"Activation {activation_id} is deleted.")
+            return
+
+        _perform_deactivate(activation)
```
bzwei, line 132: pairing with line 93, we can check activation's current status. If it is already stopped/failed etc, we should skip `_perform_deactivate`.

bzwei (approved):

mkanoor (approved):

PR #377, opened 2023-08-28
fix: ensure awx token at activation workflow

Alex-Izquierdo:
The goal is to handle scenarios where the token is absent at two specific windows:
- During the period between activation creation and activation start.
- During the period between activation start and WS connection with the ansible-rulebook worker
The PR aims to enhance the messaging output that users receive, making sure they are well-informed about the absence of the token in the activation instance output. 

Jira: https://issues.redhat.com/browse/AAP-13081


src/aap_eda/core/models/user.py
```diff
@@ -32,6 +32,10 @@ class User(AbstractUser):
     roles = models.ManyToManyField("Role", related_name="users")
     modified_at = models.DateTimeField(auto_now=True, null=False)
 
+    def get_awx_tokens(self):
```
mkanoor, line 35: @Alex-Izquierdo Can this code be shared and reused with get_awx_token by passing in the user_id?

Alex-Izquierdo: @mkanoor I'm not following you. What do you want exactly? 
This method extends the user model as a shortcut to fetch the tokens (must be multiple because we don't have the constraint of only one token) and is called in a couple of places for validations. 

get_awx_token was implemented to return only the first one from the user_id and is tied to the consumer. 

Do you want to return the first token or an exception directly from the model?

mkanoor: @Alex-Izquierdo I am thinking that code should return an array of tokens, you can do the count check in one place and in the other place take the first token. This same code needs to be called before we start an [activation](https://github.com/ansible/eda-server/blob/0b746b0c2d3d96f28cc7fba1429aace089e30c8b/src/aap_eda/tasks/ruleset.py#L40C1-L40C1), because there is no point in calling activation and then have it come over the web socket to be rejected, we can stop it right away before an activation starts either via the scheduler doing a restart or via the UI. If the number of tokens is > 1 or the number of tokens is 0. We could just stop the activation from starting with an error saying more than 1 token has been found.

get_awx_tokens(user_id)

Alex-Izquierdo: Ok, np. I think I understand what you say about the re usability. 

About the places where we are checking the token, before the PR, we only checked it at the creation of the activation. Now, I have added the check before start the activation but [after initiate the instance](https://github.com/ansible/eda-server/pull/377/files#diff-475356a7e610a5db9004d54074241bde14df30fbfabebc4cdac68aaacba9e8ccR130). I can not move the check where you say because we need an activation_instance created to notice to the user what was the error. 
The check in the websocket is also necessary because there is a time window where the token can be deleted between when the activation_instance start and the websocket connection is established.

Alex-Izquierdo:
I am also thinking that it might be better if we simply prevent deleting the token if that user has created an activation.

mkanoor:
@Alex-Izquierdo Another thing we have noticed is the automated tests from QE end up deleting the token at the end of the run, if the activation failed and it has a restart policy of failed we will attempt to restart the activation but by that time the token gets deleted as part of the QE testing. When we get a failure like the "missing token" its better to catch it on the eda-server side instead of having ansible-rulebook container end with a non zero return code triggering the restart policy. With the token gone we are just wasting CPU cycles by retrying the failed activation to be retried ACTIVATION_MAX_RESTARTS_ON_FAILURE times.

Alex-Izquierdo:
After some discussion, I continue with the original approach.

Alex-Izquierdo:
@mkanoor get_awx_token moved to utils to be reused. Please review.

Dostonbek1 (approved):
tested both scenarios mentioned in the PR description, lgtm.

src/aap_eda/services/ruleset/activate_rulesets.py
```diff
@@ -126,6 +126,14 @@ def activate(self, activation: models.Activation) -> None:
 
             activation_db_logger = ActivationDbLogger(instance.id)
 
+            if not activation.user.get_awx_tokens().exists():
+                instance.status = ActivationStatus.FAILED
```
mkanoor, line 130: @Alex-Izquierdo is this going to trigger a restart because of Restart Policy is failed. We would have to prevent a restart.

Dostonbek1: @mkanoor From my testing, it didn't trigger a restart with `on-failure` restart policy.

Alex-Izquierdo: @mkanoor I thought on that as well, but I finally thought that we keep the restart logic simpler if we treat it in the same way, because in the end is another failed activation. Keep it simple principle. Adding a token is quite quick and it is something that in the most of the cases will happen because the user has forgotten it. 

To me the user story is:
- oh, there is no token, well, the activation will be restarted, I will add it to make it working the next attempt. 

It is basically the same case as if you use a rulebook with a job_template that does not exist, the activation continues its flow and at any time you can add that job_template to AWX so that everything works without you having to do anything else. 

Also keep in mind that if the user does not notice the problem at the time it occurs, the activation will still reach the limit of attempts and you will still have to manually restart it after adding the token. 

So, to end up in the same place, if we can give the user the opportunity to not have to restart the activation and just add the token, more convenient for the user and better UX. To me is not different than another failed activation scenario. 

what do you think @hsong-rh @bzwei

Alex-Izquierdo: @Dostonbek1 that is true, right now it is only restarted manually, but I think this is because we don't have yet the restart logic 100% fully working, @hsong-rh is working on that.

mkanoor: @Alex-Izquierdo We would be relying too much on the restart logic to handle user errors and the timing issues, when the monitor picks up the task. It's much simpler to say this is a well known error, the pre-reqs for running an activation have not been satisfied so we should just mark it as failed and not restart it. The same thing happens when a DE is missing we can't say mark it as failed and hope that the user can get in between the 5 restarts and sneak in a DE. They can just restart it once they have fixed the problem. This also creates a race conditions between when the user sets the value and when the monitor picks up the task leading to more confusion.

Alex-Izquierdo: @mkanoor The DE is not the same case, because the DE is bind to the activation. Anyway, that requires to know in advance the restart workflow, so I see two options: or we implement this as it is and we change it with the modifications in the restart workflow or I wait to have it implemented to update the PR. I will talk with @hsong-rh to be aligned on this.

Alex-Izquierdo:
PR temporally on-hold until decide how to deal with this kind of cases where the activation should not follow the flow of a regular failed activation.

PR #380, opened 2023-08-30
chore: update EDA_PODMAN_VERSION default to use only major version

jshimkus-rh:
Version 4.5 is no longer available.

### DIFF
--- a/tools/docker/docker-compose-dev.yaml
+++ b/tools/docker/docker-compose-dev.yaml
@@ -17,7 +17,7 @@ x-environment:
 services:
   podman:
     user: "1000"
-    image: quay.io/containers/podman:${EDA_PODMAN_VERSION:-v4.6}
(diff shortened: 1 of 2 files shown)


tools/docker/docker-compose-dev.yaml
```diff
@@ -17,7 +17,7 @@ x-environment:
 services:
   podman:
     user: "1000"
-    image: quay.io/containers/podman:${EDA_PODMAN_VERSION:-v4.5}
+    image: quay.io/containers/podman:${EDA_PODMAN_VERSION:-v4.6}
```
Alex-Izquierdo, line 20: I think it will be better just use v4, to avoid the same problem

jshimkus-rh: Works for me.

Alex-Izquierdo (approved):

Dostonbek1 (approved):

PR #389, opened 2023-09-06
Refactor activate_ruleset with strategy pattern

hsong-rh:
Use strategy pattern to refactor `activate_ruleset` service. There is one RulesetManager, and two RulesetHandlers, one for podman, the other for k8s. RulesetManager will take the task and assign right handler to process it.


src/aap_eda/services/ruleset/podman_ruleset_handler.py
```diff
@@ -0,0 +1,71 @@
...
+from .deactivation_podman import DeactivationPodman
+from .ruleset_handler import RulesetHandler
+
+logger = logging.getLogger(__name__)
+
+
+class PodmanRulesetHandler(RulesetHandler):
+    def activate(
+        self,
+        ws_url: str,
```
mkanoor, line 32: @hsong-rh Can't we get the ws_url and ssl_verify from the settings we don't have to pass those around. At the point where we use it to generate the command line for activation we can use it directly from settings. The command line for ansible-rulebook is going to be the same for both podman and k8s.

src/aap_eda/services/ruleset/ruleset_manager.py
```diff
@@ -0,0 +1,257 @@
...
+                "according to the activation's restart policy."
+            )
+            logger.warning(msg)
+        else:
+            seconds = int(settings.ACTIVATION_RESTART_SECONDS_ON_COMPLETE)
+            msg = (
+                f"Activation {activation.name} completed successfully. Will "
+                f"restart in {seconds} seconds according to its restart policy"
+            )
+            logger.info(msg)
```
mkanoor, line 257: @hsong-rh Should this also have a command line generator method which can be called from the k8s and podman code, the cmdline is the same and we might want to add extra parameters down the road as we add new options into the Activation

Alex-Izquierdo: We already have something like that in the e2e test suite for rulebook: https://github.com/ansible/ansible-rulebook/blob/main/tests/e2e/utils.py#L18

src/aap_eda/services/ruleset/rulebook_cmd.py
```diff
@@ -0,0 +1,74 @@
...
+    Represents ansible_rulebook command and their arguments, and
+    provides methods to render it for cmd runners.
+    """
+
+    program_name: str = "ansible-rulebook"
+    activation_instance_id: Union[str, int, None] = None
+    verbose: bool = False
+    debug: bool = False
+    websocket: Optional[str] = None
+    websocket_ssl_verify: Optional[str] = None
```
mkanoor, line 32: @hsong-rh Can we read in the settings here directly as the default value

hsong-rh:
Refactor work moved to the other branch, closed this.

PR #390, opened 2023-09-06
Set current_job_id for activations started by scheduler

bzwei:
Minor fix. Self explained.

### DIFF
--- a/src/aap_eda/tasks/ruleset.py
+++ b/src/aap_eda/tasks/ruleset.py
@@ -155,7 +155,7 @@ def restart(activation_id: int, requester: str = "User") -> None:
         activation.status = ActivationStatus.PENDING
         activation.save(update_fields=["status", "modified_at"])
 
-        activate.delay(activation_id, requester)
+        _schedule_activate(activation, requester)
 
 
 def _perform_deactivate(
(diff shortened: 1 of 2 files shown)


hsong-rh:
@bzwei https://github.com/ansible/eda-server/blob/main/src/aap_eda/tasks/ruleset.py#L158 needs the similar fix, can you do it together?

hsong-rh (approved):

Alex-Izquierdo:
Shouldn't we do the same when we call `deactivate.delay`?

hsong-rh:
> Shouldn't we do the same when we call `deactivate.delay`?

@Alex-Izquierdo https://github.com/ansible/eda-server/blob/main/src/aap_eda/tasks/ruleset.py#L114 set it to None already

hsong-rh:
@bzwei should we move the saving current_job_id logic inside `activate`, so no need to call it from outside multiple times? Something like:
```
def activate(activation_id: int, requester: str = "User") -> None:
....
    job = get_current_job()
    activation.current_job_id = job.id
    activation.save(update_fields=["current_job_id"])
....
```

bzwei:
> @bzwei should we move the saving current_job_id logic inside `activate`, so no need to call it from outside multiple times? Something like:
> 
> ```
> def activate(activation_id: int, requester: str = "User") -> None:
> ....
>     job = get_current_job()
>     activation.current_job_id = job.id
>     activation.save(update_fields=["current_job_id"])
> ....
> ```

Unfortunately we can't do this. We need to save the job id as soon as it is queued, but not when it is executed.

Alex-Izquierdo (approved):

PR #393, opened 2023-09-07
[AAP-15867] Except DatabaseError in parsing heartbeat message

hsong-rh:
The DatabaseError raised in parsing heartbeat message may be the reason of GH e2e tests failure.

Solves: [AAP-15867](https://issues.redhat.com/browse/AAP-15867)

### DIFF
--- a/src/aap_eda/wsapi/consumers.py
+++ b/src/aap_eda/wsapi/consumers.py
@@ -7,7 +7,7 @@ from enum import Enum
 from channels.db import database_sync_to_async
 from channels.generic.websocket import AsyncWebsocketConsumer
 from django.conf import settings
-from django.db import IntegrityError
+from django.db import DatabaseError
 
 from aap_eda.core import models
 from aap_eda.core.enums import ActivationStatus
@@ -98,8 +98,8 @@ class AnsibleRulebookConsumer(AsyncWebsocketConsumer):
                 await self.handle_heartbeat(HeartbeatMessage.parse_obj(data))
             else:
                 logger.warning(f"Unsupported message received: {data}")
-        except IntegrityError as e:
-            logger.error(f"{str(e)}")
+        except DatabaseError as err:
+            logger.error(f"Failed to parse {data} due to DB error: {err}")
 
     async def handle_workers(self, message: WorkerMessage):
         logger.info(f"Start to handle workers: {message}")


src/aap_eda/wsapi/consumers.py
```diff
@@ -98,8 +98,8 @@ async def receive(self, text_data=None, bytes_data=None):
                 await self.handle_heartbeat(HeartbeatMessage.parse_obj(data))
             else:
                 logger.warning(f"Unsupported message received: {data}")
-        except IntegrityError as e:
-            logger.error(f"{str(e)}")
+        except (IntegrityError, DatabaseError) as err:
```
jshimkus-rh, line 101: Catching the generic DatabaseError could end up masking other exceptions might be problematic; e.g., continuing to run when stopping/dying would be better.
Given django is using DatabaseError in the case this is to address check that `type(err) == DatabaseError`, while far from ideal, would not let others slip by.
Unless we know that any db exception is tolerable, in which case `IntegrityError` isn't needed.

bzwei (dismissed):

src/aap_eda/wsapi/consumers.py
jshimkus-rh: Though I'm not a fan of doing so, `str(err)` could also be checked for containing `"did not affect any rows"`.  
It would add a bit more certainty and at least we'd find out if django ever changed the particular text.

hsong-rh: @jshimkus-rh Thank you for your review.
> Catching the generic DatabaseError could end up masking other exceptions might be problematic; e.g., continuing to run when stopping/dying would be better. Given django is using DatabaseError in the case this is to address check that `type(err) == DatabaseError`, while far from ideal, would not let others slip by. Unless we know that any db exception is tolerable, in which case `IntegrityError` isn't needed.

These two exceptions are raised due to the deactivation, which will update/delete the final records and stop the process. The API worker here just logs the message and continue to run.

As you suggested, the log message is changes with more meaningful info

jshimkus-rh: That doesn't actually address the concern.
All of the message handling is in the same `try/except` construct so any `DatabaseError` (including subclasses) generated from any of them could inadvertently end up being ignored (other than the log message).

It seems like you may be saying that for `MessageType.SESSION_STATS` these are expected exceptions.  Assuming that to be the case handling `MessageType.SESSION_STATS` in its own `try/except` construct and tolerating those exceptions there while the other `MessageType`s don't suppress those exceptions would address the concern.  

Or maybe you mean something else?

hsong-rh: Actually all database accesses related with activation instances (which can be removed by default worker) need ignore these exceptions. It's not only for `MessageType.SESSION_STATS`.

jshimkus-rh: OK.  If that's the case though, `IntegrityError` isn't necessary as it's a subclass of `DatabaseError` and that'f be all that's needed on the `except'`.

jshimkus-rh (approved):
Approved.

Alex-Izquierdo (approved):

bzwei (approved):

PR #394, opened 2023-09-08
Ensure saved activation and instance attributes are accurate

bzwei:
Follow up recent PRs with a few minor enhancements and bug fixes, center around saving activation and instances:
1. Replace `filter().update()` with `save(update_fields=)`. This way modified attributes remain in the in memory objects.
2. Call `refresh_from_db()` after a long running activation instance
3. Populate `status_update_at` when an activation is saved
4. Replace `except (IntegrityError, DatabaseError)` with `except DatabaseError`

### DIFF
--- a/src/aap_eda/services/ruleset/activation_kubernetes.py
+++ b/src/aap_eda/services/ruleset/activation_kubernetes.py
@@ -18,7 +18,7 @@ import logging
 import time
 
 from django.conf import settings
-from django.db import DatabaseError, IntegrityError
+from django.db import DatabaseError
 from django.utils import timezone
 from kubernetes import client, config, watch
 from kubernetes.client import exceptions
@@ -415,19 +415,21 @@ class ActivationKubernetes:
         self, instance: models.ActivationInstance, status: ActivationStatus
     ) -> None:
         try:
+            now = timezone.now()
             instance.status = status
-            instance.updated_at = timezone.now()
+            instance.updated_at = now
             instance.save(update_fields=["status", "updated_at"])
 
-            update_fields = ["status", "modified_at"]
+            update_fields = ["status", "modified_at", "status_updated_at"]
             activation = instance.activation
             activation.refresh_from_db()
             activation.status = status
+            activation.status_updated_at = now
             if status == ActivationStatus.RUNNING and not activation.is_valid:
                 activation.is_valid = True
                 update_fields.append("is_valid")
             activation.save(update_fields=update_fields)
-        except (IntegrityError, DatabaseError):
+        except DatabaseError:
             message = f"Activation instance {instance.id} is not present."
             raise ActivationRecordNotFound(message)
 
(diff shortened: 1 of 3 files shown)


hsong-rh (commented):
https://github.com/ansible/eda-server/pull/394/files#diff-b02e2b6212c1721597487e00297cca8a0890cbbc2a770d67046a369a6074505eR183 also need to remove IntegrityError

hsong-rh (approved):

src/aap_eda/services/ruleset/activation_kubernetes.py
```diff
@@ -415,19 +415,21 @@ def set_activation_status(
         self, instance: models.ActivationInstance, status: ActivationStatus
     ) -> None:
         try:
+            now = timezone.now()
             instance.status = status
-            instance.updated_at = timezone.now()
+            instance.updated_at = now
```
Alex-Izquierdo, line 420: I wonder why we don't set this field with `auto_now=True` is there a reason for it?

src/aap_eda/services/ruleset/activation_kubernetes.py
```diff
@@ -415,19 +415,21 @@ def set_activation_status(
...
             instance.save(update_fields=["status", "updated_at"])
 
-            update_fields = ["status", "modified_at"]
+            update_fields = ["status", "modified_at", "status_updated_at"]
```
Alex-Izquierdo, line 423: Wouldn't it better to override the "save" method for the activation model to update "status_updated_at" every time "status" is updated?

src/aap_eda/services/ruleset/activation_podman.py
```diff
@@ -145,20 +141,25 @@ def run_worker_mode(
...
+            activation = activation_instance.activation
+            activation.status = ActivationStatus.RUNNING
+            activation.is_valid = True
+            activation.status_updated_at = now
+            activation.save(
+                update_fields=[
+                    "status",
+                    "status_updated_at",
+                    "is_valid",
+                    "modified_at",
```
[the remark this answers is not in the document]
bzwei: when save with `update_fields` we need to pick explicitly `modified_at` even though we don't need to provide a value.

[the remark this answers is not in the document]
Alex-Izquierdo: Indeed, I wrongly supposed that it would work OTB. Anyway, taking care of updating these fields seems to me a bit error prone. Isn't it still better to centralize this logic in the model?

src/aap_eda/services/ruleset/activation_kubernetes.py
bzwei: The value can be set by an external value https://github.com/ansible/eda-server/blob/main/src/aap_eda/wsapi/consumers.py#L151

src/aap_eda/services/ruleset/activation_kubernetes.py
bzwei: The value can be set by an external value https://github.com/ansible/eda-server/blob/main/src/aap_eda/wsapi/consumers.py#L156

Alex-Izquierdo: gotcha, fair point. That logic is a different discussion :)

Alex-Izquierdo (approved):
LGTM

PR #395, opened 2023-09-11
feat(model): add ModifiedAtUpdaterMixin to ensure modified_at is always updated

Alex-Izquierdo:
Add common logic to the models to ensure modified_at field is always updated without need to take care of it every time `update_fields`is param is used


Alex-Izquierdo:
If you like this approach, my plan would be to extend it in later iterations to cover other similar cases like for example "status_updated_at":
https://github.com/ansible/eda-server/blob/main/src/aap_eda/wsapi/consumers.py#L156 where we should set the datetime of the server.

jshimkus-rh:
> If you like this approach, my plan would be to extend it in later iterations to cover other similar cases like for example "status_updated_at":
> https://github.com/ansible/eda-server/blob/main/src/aap_eda/wsapi/consumers.py#L156 where we should set the datetime of the server.

jshimkus-rh:
Sorry 'bout that.  Clicked the wrong button. :(

jshimkus-rh:
Curious as to the reason for not incorporating the mixin in User (which has an auto_now modified_at)?

Alex-Izquierdo:
> Curious as to the reason for not incorporating the mixin in User (which has an auto_now modified_at)?
@jshimkus-rh 
Because I missed it. I add it :)

src/aap_eda/core/models/mixins.py
```diff
@@ -0,0 +1,21 @@
...
+class ModifiedAtUpdaterMixin:
+    """Mixin to ensure that the `modified_at` field is updated.
+
+    Ensure that the `modified_at` field is updated when the model
+    is saved with specific fields (`update_fields`) specified for update.
+    Only works for `auto_now` fields.
+    """
+
+    def save(self, *args, **kwargs):
+        for field in self._meta.fields:
```
jshimkus-rh, line 13: Rather than iterating the code could use `self._meta.get_field("modified_at")`; would need to catch the `FieldDoesNotExist` exception.

src/aap_eda/core/models/mixins.py
```diff
@@ -0,0 +1,21 @@
...
+            if (
+                field.name == "modified_at"
+                and hasattr(field, "auto_now")
+                and field.auto_now
+                and kwargs.get("update_fields", [])
+            ):
+                self.modified_at = timezone.now()
```
jshimkus-rh, line 20: Given the "fail safe" nature of the implementation it's the kind of change that  could be incorporated as a default for everything; though I can imagine that some code might not want that behavior as a default.  In which case there'd have to be a conscious decision to include it and that could be used to support the argument that the checking isn't necessary.

src/aap_eda/core/models/mixins.py
Alex-Izquierdo: Good idea! thanks

src/aap_eda/core/models/mixins.py
Alex-Izquierdo: I preferred to be conservative. I think for modified_at field we can do it always, but there are some "updated_at" that has not the auto_now set (which is something that I want to change later because it is wrong) 

But I would agree with you, as long as we are adding the mixin on purpose, the check may be not necessary.

src/aap_eda/core/models/mixins.py
Alex-Izquierdo: updated.

jshimkus-rh (approved):

src/aap_eda/core/models/mixins.py
```diff
@@ -0,0 +1,25 @@
...
+            field = self._meta.get_field("modified_at")
+        except FieldDoesNotExist:
+            field = None
+        if (
+            field is not None
+            and hasattr(field, "auto_now")
+            and field.auto_now
+            and kwargs.get("update_fields", [])
+        ):
+            self.modified_at = timezone.now()
```
bzwei, line 24: Would kwargs["update_fields"].append("modified_at") work? It should automatically populate the timestamp.
The concern is the in memory object may not have the latest modified_at value after the save.

I am surprised to see you only set the new value for `modified_at` but not include this field in `update_fields` array and the new value is still saved. How can it work?

Alex-Izquierdo: It doesn't. I forgot to add that line. The test is wrong, I need to refresh the object. Thanks, very good catch!

src/aap_eda/core/models/mixins.py
```diff
@@ -0,0 +1,25 @@
+from django.core.exceptions import FieldDoesNotExist
```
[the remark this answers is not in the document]
Alex-Izquierdo: updated.

Alex-Izquierdo:
test_retrieve_user_details now fails, not sure why. I need to investigate

Alex-Izquierdo:
This PR was based on the assumption that we want this field updated always, but I have found cases when it is not true. So I believe this is a wrong approach and it is better to explicitly update the fields.

PR #397, opened 2023-09-12
Unique activation task

bzwei:
Ensure only one task can be queued per activation-id.
Same for activation_monitor task.


src/aap_eda/tasks/ruleset.py
```diff
@@ -208,6 +225,17 @@ def _perform_deactivate(
 
 # Started by the scheduler, executed by the default worker
 def monitor_activations() -> None:
+    job_id = "monitor_activations"
+    queue = get_queue("default")
+    if _job_in_queue(queue, job_id):
+        logger.info(
+            "Skip monitor_activations because another similar task is running"
```
Alex-Izquierdo, line 232: I think this msg is not accurate. Is not a "similar" but the exact same one. Not necessarily running, you are checking enqueued states. The id might be useful for debugging. 

```suggestion
            f"Skip monitor_activations task id {job_id} "
            "because another instance of this task is already enqueued."
```

src/aap_eda/api/views/activation.py
```diff
@@ -109,7 +109,7 @@ def create(self, request):
         activation["rules_fired_count"] = 0
 
         if response.is_enabled:
-            job = activate.delay(activation_id=response.id)
+            job = activate(activation_id=response.id)
```
Alex-Izquierdo, line 112: activate function is not returning anything.

Alex-Izquierdo (commented):
According to the type hint activate function returns None, but their are callers expect the job.

tests/integration/tasks/test_ruleset.py
```diff
@@ -23,9 +23,9 @@
 from aap_eda.core.enums import ActivationStatus, RestartPolicy
 from aap_eda.services.ruleset.activate_rulesets import ActivateRulesets
 from aap_eda.tasks.ruleset import (
-    activate,
+    _activate,
```
mkanoor, line 26: @bzwei Why do we have to import private functions, can we not make them private if you want to import them into other files.

bzwei: function `activate` is to place `_activate` into queue. `_activate` is imported here purely for testing purpose. I guess we could lift it to public if it is a rule or best practice for import.

src/aap_eda/tasks/ruleset.py
```diff
@@ -158,6 +176,18 @@ def restart(activation_id: int, requester: str = "User") -> None:
         _schedule_activate(activation, requester)
 
 
+def _job_in_queue(queue: Queue, job_id: str) -> Optional[Job]:
+    job = queue.fetch_job(job_id)
+    if job and job.get_status(refresh=True) in [
+        JobStatus.QUEUED,
+        JobStatus.STARTED,
+        JobStatus.DEFERRED,
+        JobStatus.SCHEDULED,
```
mkanoor, line 185: @bzwei Since we are cancelling the RQ job before we call deactivation, if a job is being Canceled before the deactivation has started wouldn't this code let it thru and then kick off an activation while the deactivation is still running.

bzwei: This work does not address the problem you raised, nor does it make it worse. Since activation is normally slower than deactivation, activating a new instance is likely safe while an old instance is deactivating.

Alex-Izquierdo: Not related with this PR but I think is a fair question. Should we cancel the task before deactivation? It seems wrong to me and it is a place for race conditions, I feel the task should be cancel at the end.

src/aap_eda/tasks/ruleset.py
```diff
@@ -158,6 +176,18 @@ def restart(activation_id: int, requester: str = "User") -> None:
         _schedule_activate(activation, requester)
 
 
+def _job_in_queue(queue: Queue, job_id: str) -> Optional[Job]:
```
Alex-Izquierdo, line 179: This function and your approach to make a task unique is something that is going to be reusable. At least I would follow the pattern for https://github.com/ansible/eda-server/pull/396. 

Can be this function public and located in a common place?

tests/integration/tasks/test_ruleset.py
Alex-Izquierdo: Why we wouldn't test private functions?

bzwei:
@Alex-Izquierdo I made public functions `unique_enqueue` and `job_from_queue` in `aap_eda.core.tasking`

Alex-Izquierdo (approved):

mkanoor (approved):

PR #398, opened 2023-09-12
[AAP-16008- API] Users can not delete themselves

msmagnanijr:
Please see: https://issues.redhat.com/browse/AAP-16008


msmagnanijr:
@Alex-Izquierdo I am still facing difficulties displaying the message in the user interface. Could you please review my issue? I may need to update the test as well.

AAP-16008

Alex-Izquierdo:
> I am still facing difficulties displaying the message in the user interface. Could you please review it? I will update the test.

The UI must be updated as well @msmagnanijr @lgalis has a ticket for it. But I think it would be good if we sync the changes of both sides.

tests/integration/api/test_user.py
```diff
@@ -272,9 +272,13 @@ def test_delete_user(
 ):
     user_id = user.id
     response = client.delete(f"{api_url_v1}/users/{user_id}/")
-    assert response.status_code == status.HTTP_204_NO_CONTENT
 
-    assert models.User.objects.filter(id=user_id).count() == 0
+    if user_id != user.id:
```
Alex-Izquierdo, line 276: I think this kind of conditionals in tests are not a good practice and might result in a lack of coverage and a false sense of confidence.
A test should be, in general, single case scoped. When the logic of the test doesn't change across the cases, when for example you only want to test a set of different inputs, [pytest parametrization](https://docs.pytest.org/en/7.3.x/how-to/parametrize.html) is a good choice. 

But in this case It looks like we should have a different test.

Alex-Izquierdo (changes requested):

tests/integration/api/test_user.py
msmagnanijr: Got it! It really makes sense. I will update the test.

src/aap_eda/api/views/user.py
```diff
@@ -236,3 +240,12 @@ def get_serializer_class(self):
 
     def get_response_serializer_class(self):
         return serializers.UserDetailSerializer
+
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
+
+        if instance == request.user:
```
hsong-rh, line 247: @msmagnanijr Should we check if the user is admin or not? Only admin has the right to delete, right?

msmagnanijr: @hsong-rh From what I understand, not even the admin can delete itself. What do you think? @Alex-Izquierdo

hsong-rh: The title **Admin users can delete themselves in EDA** tells me admin can do it 😄

msmagnanijr: @hsong-rh you have a point :sweat_smile:

Alex-Izquierdo: @msmagnanijr @hsong-rh the original bug was raised concerning an admin user, but in later discussions we decided to prevent it for any user. The UI will apply a similar restriction, disabling the button.

Alex-Izquierdo (approved):

PR #399, opened 2023-09-12
[AAP-16014] Fail activation on exception

bzwei:
Fixes AAP-16014: Activation not marked failed if exception is raised

This bug was introduced by recent bug fixing work on deactivation (#355), thus is needed before our next release.

### DIFF
--- a/src/aap_eda/services/ruleset/activate_rulesets.py
+++ b/src/aap_eda/services/ruleset/activate_rulesets.py
@@ -173,24 +173,24 @@ class ActivateRulesets:
                     activation_db_logger,
                 )
 
-            activation_db_logger.flush()
-
-            now = timezone.now()
-            instance.ended_at = now
-            instance.updated_at = now
-            save_activation_and_instance(
-                instance=instance,
-                update_fields=["status", "ended_at", "updated_at"],
-            )
-        except (DeactivationException, ActivationRecordNotFound) as error:
+            self._final_update(instance, activation_db_logger)
+        except (
+            DeactivationException,
+            ActivationRecordNotFound,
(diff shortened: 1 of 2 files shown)


src/aap_eda/services/ruleset/activate_rulesets.py
```diff
@@ -191,6 +191,14 @@ def activate(self, activation: models.Activation) -> None:
                 instance,
                 activation_db_logger,
             )
+            activation_db_logger.flush()
```
hsong-rh, line 194: The activation instance may be removed by deactivation when this line is called. It will raise exception and not update the following status.

jshimkus-rh: Doing the `activation.save(`) in `_log_activate_failure` would address this?  Even if not, having one branch in `_log_activate_failure` doing a save and the other not doesn't seem correct.

bzwei: Deactivation does not remove any instance or activation. If the activation/instance were deleted then we don't need to update the status.

hsong-rh: The activation/instance may be deleted during activation is running. Activation itself doesn't know anything about deletion, so will continue to update the status.

bzwei: I have further enhanced the error handling. If we catch error from a deleted DB record, just log the error and stop.

hsong-rh (approved):
LGTM

jshimkus-rh (approved):

PR #402, opened 2023-09-13
[AAP-14680] Add check for DE deletion if in use by Activations 

Dostonbek1:
* add a check for activations using DE on delete
  * add force flag to force delete DE
  * add filter by decision_environment_id for list Activations

Solves: [AAP-14680](https://issues.redhat.com/browse/AAP-14680)

### DIFF
--- a/src/aap_eda/api/views/decision_environment.py
+++ b/src/aap_eda/api/views/decision_environment.py
@@ -21,7 +21,7 @@ from drf_spectacular.utils import (
 from rest_framework import mixins, status, viewsets
 from rest_framework.response import Response
 
-from aap_eda.api import filters, serializers
+from aap_eda.api import exceptions as api_exc, filters, serializers
 from aap_eda.core import models
 from aap_eda.core.enums import ResourceType
 
@@ -59,14 +59,6 @@ from .mixins import (
             status.HTTP_200_OK: OpenApiResponse(
                 serializers.DecisionEnvironmentSerializer,
                 description="Update successful. Return an updated decision environment.",  # noqa: E501
-            )
-        },
-    ),
-    destroy=extend_schema(
-        description="Delete a decision environment by id",
-        responses={
-            status.HTTP_204_NO_CONTENT: OpenApiResponse(
-                None, description="Delete successful."
             )
         },
     ),
@@ -117,3 +109,26 @@ class DecisionEnvironmentViewSet(
                 decision_environment.data
             ).data
         )
+
+    @extend_schema(
+        description="Delete a decision environment by id",
+        responses=status.HTTP_204_NO_CONTENT,
+    )
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
+        force_delete = request.GET.get("force")
+
+        activations = models.Activation.objects.filter(
+            decision_environment_id=instance.id
+        ).values_list("name", flat=True)
+
+        if activations and not force_delete:
+            raise api_exc.Conflict(
+                "Decision Environment cannot be deleted; "
+                "it is being used by the following Activations: "
+                f"{list(activations)}. If you want to force delete, "
+                "please add /?force=True query param."
+            )
+
+        self.perform_destroy(instance)
+        return Response(status=status.HTTP_204_NO_CONTENT)
--- a/src/aap_eda/api/views/decision_environment.py
+++ b/src/aap_eda/api/views/decision_environment.py
@@ -116,7 +116,7 @@ class DecisionEnvironmentViewSet(
     )
     def destroy(self, request, *args, **kwargs):
         instance = self.get_object()
-        force_delete = request.GET.get("force")
+        force_delete = request.query_params.get("force", False)
 
         activations = models.Activation.objects.filter(
             decision_environment_id=instance.id
--- a/src/aap_eda/api/views/decision_environment.py
+++ b/src/aap_eda/api/views/decision_environment.py
@@ -14,6 +14,7 @@
 
 from django_filters import rest_framework as defaultfilters
 from drf_spectacular.utils import (
+    OpenApiParameter,
     OpenApiResponse,
     extend_schema,
     extend_schema_view,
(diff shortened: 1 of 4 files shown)


hsong-rh:
@Dostonbek1 Can you unit tests to assert new added decision_environment_name exists in the response of GET/LIST of activations?

src/aap_eda/api/serializers/activation.py
```diff
@@ -129,6 +131,12 @@ def create(self, validated_data, user):
         validated_data["rulebook_rulesets"] = rulebook.rulesets
         validated_data["git_hash"] = rulebook.project.git_hash
         validated_data["project_id"] = rulebook.project.id
+        if validated_data["decision_environment_id"]:
+            validated_data[
+                "decision_environment_name"
+            ] = models.DecisionEnvironment.objects.get(
```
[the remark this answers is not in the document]
Dostonbek1: We already have that check [here](https://github.com/ansible/eda-server/blob/aea444d606828512e3994455447b010906a09830/src/aap_eda/api/views/activation.py#L37), which gets executed when we try to use `decision_environment_id` in validated_data. I thought it would be redundant to throw an exception here as well. Thoughts?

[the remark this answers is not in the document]
hsong-rh: Is that function `handle_activation_create_conflict` called because this: https://github.com/ansible/eda-server/pull/402/files#diff-b75687856dd6a2e0a5e5a1d5d322ed63ca815ecd7fbfbb6aadb25009b4d0e89bL126? so I think we need a similar check on decision environment.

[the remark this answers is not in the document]
Dostonbek1: It gets called because of the IntegrityError on this [line](https://github.com/ansible/eda-server/pull/402/files#diff-b75687856dd6a2e0a5e5a1d5d322ed63ca815ecd7fbfbb6aadb25009b4d0e89bR140) when any of the 3 related object IDs are incorrect.

Alex-Izquierdo:
if the DE environment is removed the activation would be partially disabled. I wonder if we should return "deleted" instead of overload the activation table.

bzwei:
> if the DE environment is removed the activation would be partially disabled. I wonder if we should return "deleted" instead of overload the activation table.

Agree. There is no much value to preserve the name of DE since the whole record is gone. If we really want to keep evidence for debugging, we should not actually delete db records, but instead mark them as deleted.

src/aap_eda/api/views/decision_environment.py
```diff
@@ -117,3 +109,18 @@ def retrieve(self, request, pk):
...
+    @extend_schema(
+        description="Delete a decision environment by id",
+        responses=status.HTTP_204_NO_CONTENT,
+    )
+    def perform_destroy(self, instance):
+        activations = models.Activation.objects.filter(
+            decision_environment_id=instance.id
+        ).values_list("id", flat=True)
+        if activations:
+            raise api_exc.Conflict(
```
Alex-Izquierdo, line 122: To me this is the right approach, it should be done also for extra_Vars, credentials and token. But it needs to be discussed, last conversation with @benthomasson he didn't want to do this. Even if we go ahead with this, UI should be aligned to capture these cases.

src/aap_eda/api/views/decision_environment.py
```diff
@@ -117,3 +109,18 @@ def retrieve(self, request, pk):
                 decision_environment.data
             ).data
         )
+
+    @extend_schema(
+        description="Delete a decision environment by id",
+        responses=status.HTTP_204_NO_CONTENT,
+    )
+    def perform_destroy(self, instance):
+        activations = models.Activation.objects.filter(
```
Alex-Izquierdo, line 118: If this message is user faced it should be the activation names not the ids.

benthomasson:
Let's not prevent deletion, but instead provide a warning that an activation is using this DE and then allow the user to delete it.

src/aap_eda/api/views/decision_environment.py
```diff
@@ -117,3 +109,26 @@ def retrieve(self, request, pk):
...
             ).data
         )
+
+    @extend_schema(
+        description="Delete a decision environment by id",
+        responses=status.HTTP_204_NO_CONTENT,
+    )
+    def destroy(self, request, *args, **kwargs):
+        instance = self.get_object()
+        force_delete = request.query_params.get("force", False)
```
Alex-Izquierdo, line 119: If force == "false", you will delete the object because `bool("false") == False` . You must to use string comparison and checking the positive case for safety: 
```suggestion
        force_delete = request.query_params.get("force", "false")
        ...
        if activations and force_delete.lower() != "true":
```

Alex-Izquierdo (changes requested):

src/aap_eda/api/views/decision_environment.py
```diff
@@ -117,3 +109,26 @@ def retrieve(self, request, pk):
...
+
+        activations = models.Activation.objects.filter(
+            decision_environment_id=instance.id
+        ).values_list("name", flat=True)
+
+        if activations and not force_delete:
+            raise api_exc.Conflict(
+                "Decision Environment cannot be deleted; "
+                "it is being used by the following Activations: "
+                f"{list(activations)}. If you want to force delete, "
```
Alex-Izquierdo, line 129: We should not put here the activations. might be many. It is simpler just to say that are being used by activations. The api rest is considered a lower level interface.

src/aap_eda/api/views/decision_environment.py
```diff
@@ -117,3 +109,26 @@ def retrieve(self, request, pk):
                 decision_environment.data
             ).data
         )
+
+    @extend_schema(
```
Alex-Izquierdo, line 113: new param must be added to the openapi spec. 
```suggestion
    @extend_schema(
        parameters=[
            OpenApiParameter(
                name='force',
                description='Force deletion even if there are dependent objects',
                required=False,
                type='boolean',
            )
        ],
```

Alex-Izquierdo (changes requested):

src/aap_eda/api/views/decision_environment.py
Alex-Izquierdo: on a second thought we might want to support as well "1" and "yes" so maybe it's better to write a helper to convert it. (I couldn't find any builtin tool for that)

src/aap_eda/api/views/decision_environment.py
Dostonbek1: removed.

src/aap_eda/api/views/decision_environment.py
Dostonbek1: added.

src/aap_eda/api/views/decision_environment.py
Dostonbek1: done.

Alex-Izquierdo (approved):

src/aap_eda/api/views/decision_environment.py
```diff
@@ -117,3 +110,33 @@ def retrieve(self, request, pk):
...
+
+        activations_exist = models.Activation.objects.filter(
+            decision_environment_id=instance.id
+        ).exists()
+
+        if activations_exist and force_delete not in ["true", "1", "yes"]:
+            raise api_exc.Conflict(
+                "Decision Environment is being used by Activations "
+                "and cannot be deleted. If you want to force delete, "
+                "please add /?force=True query param."
```
mkanoor, line 138: @Dostonbek1 Should this be lowercase or uppercase True?

Dostonbek1: It doesn't really matter. On line 128, we deal with different variations of the param string. So it can be `True`, `true`, `TRUE`, etc. and we do `.lower()` for comparison.

mkanoor (approved):

PR #405, opened 2023-09-13
chore: upgrade pg image and improve startup order

Alex-Izquierdo:
c8s images doesn't have support for ARM, moving it to c9s that has it. 
Also includes a fix related with the startup order. If the DB is clean, the workers might run before complete the migrations and fail.


bzwei:
Tested. The new image works on ARM.

bzwei (approved):

hsong-rh (approved):
docker-compose tested on Mac, worked properly

PR #406, opened 2023-09-13
(chore) Use str as additional base for core.enums.

jshimkus-rh:
Allows direct comparison to enumeration instances.

### DIFF
--- a/src/aap_eda/core/enums.py
+++ b/src/aap_eda/core/enums.py
@@ -12,10 +12,10 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from enum import StrEnum
+from enum import Enum
 
 
-class DjangoStrEnum(StrEnum):
+class DjangoStrEnum(str, Enum):
     @classmethod
     def choices(cls):
         return tuple((e.value, e.value) for e in cls)
@@ -23,6 +23,9 @@ class DjangoStrEnum(StrEnum):
     @classmethod
     def values(cls):
         return tuple(e.value for e in cls)
+
+    def __str__(self):
+        return str(self.value)
 
 
 # =======================================================================
--- a/src/aap_eda/core/enums.py
+++ b/src/aap_eda/core/enums.py
@@ -24,6 +24,18 @@ class DjangoEnum(Enum):
     def values(cls):
         return tuple(e.value for e in cls)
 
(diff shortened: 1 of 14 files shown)


src/aap_eda/core/enums.py
```diff
@@ -24,6 +24,18 @@ def choices(cls):
     def values(cls):
         return tuple(e.value for e in cls)
 
+    def __eq__(self, other):
```
Alex-Izquierdo, line 27: - What if the enum is not used with strings? (unlikely but possible TBH) Somehow I feel would be better and simpler if our string enums inherits explicitly also from str. Equivalent to the new [StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum)
eg:
```
class RestartPolicy(str, DjangoEnum):
```
- If we go with this approach should the class be renamed to `DjangoStrEnum`?

jshimkus-rh: An excellent suggestion.

jshimkus-rh: Ugh.  StrEnum is new in 3.11.

Alex-Izquierdo: yeah, that's why I suggested inherit from str where we want to deal with strEnums:
```
class RestartPolicy(str, DjangoEnum):
```

src/aap_eda/core/enums.py
```diff
@@ -12,10 +12,10 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from enum import Enum
+from enum import StrEnum
```
Alex-Izquierdo, line 15: we need to support python3.9

jshimkus-rh: BTW, the original check for a string type was predicated on the idea that an appropriate instance can be created based on a string which would throw an exception if not one of the defined enum values.  That, together with "just because something might be coercible to a string doesn't mean it's a good candidate for a comparand."

Interestingly (or not :-)) even 3.11's StrEnum doesn't handle byte objects for comparison (those being, in my experience, the things most likely to show up when you were expecting a string).

Alex-Izquierdo: But I think it is good to enforce the type, isn't?

bzwei (approved):

PR #409, opened 2023-09-18
fix: handle errors at pulling images for podman - AAP-16189

Alex-Izquierdo:
Due to a bug in podman we are not handling exception when the image pull fails, raising a misleading error message after try to run the activation. 
Jira: https://issues.redhat.com/browse/AAP-16189

### DIFF
--- a/src/aap_eda/services/ruleset/activation_podman.py
+++ b/src/aap_eda/services/ruleset/activation_podman.py
@@ -29,6 +29,7 @@ from podman.errors.exceptions import APIError, NotFound
 
 from aap_eda.core import models
 from aap_eda.core.enums import ActivationStatus
+from aap_eda.services.exceptions import PodmanImagePullError
 
 from .activation_db_logger import ActivationDbLogger
 from .exceptions import ActivationException, ActivationRecordNotFound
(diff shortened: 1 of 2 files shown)


src/aap_eda/services/ruleset/activation_podman.py
```diff
@@ -277,9 +279,17 @@ def _pull_image(self) -> Image:
...
-            )
+            image = self.client.images.pull(image_url, **kwargs)
+
+            # https://github.com/containers/podman-py/issues/301
+            if not image.id:
+                msg = (
+                    f"Image {image_url} pull failed. The image url "
+                    "or the credentials may be incorrect."
+                )
+                logger.exception(msg)
```
hsong-rh, line 290: This method should only be called from an exception handler, right?

hsong-rh (approved):

Dostonbek1 (approved):
Works as expected:
Screenshot 2023-09-19 at 9 50 03 AM

Minor comment: I think there was a discussion of catching "un-pullable" images on DE creation. For example, when creating a DE, we could try to test-pull the image and see if it is valid, which would save us from unnecessary errors on Activation creation. But I think error catching here is still necessary because although the image URL is valid, the pull still might fail for other reasons, such as network connectivity.

PR #410, opened 2023-09-18
[AAP-15464] Schedule a recurring task to detect dead activation workers

bzwei:
Detect activations that connect to dead/inactive activation workers due to service restart or worker crashing.
Fail the activation if the rulebook container no longer exists, otherwise attempt to reconnect the container to an idle activation worker.

The original thought was to use a startup job to detect such activations, but it cannot detect the case when only one activation worker crashes but other server/works are working fine. So we append this work to the existing activation_monitor task.

Rename the activation's `current_job_id` column to `worker_name` to store the worker's name. There is no longer the need to store `current_job_id` because the job id is now uniquely determined by the `activation_id`.


src/aap_eda/core/tasking/__init__.py
```diff
@@ -150,6 +150,9 @@ class ActivationWorker(_Worker):
     Uses JSONSerializer as a default one.
     """
 
+    # Assumption: Each process has only one ActivationWorker
```
Alex-Izquierdo, line 153: Is this assumption sustainable beyond the short term? I feel it might be an approach that we will have to remade sooner or later. I see potential problems with it: it might complicate tests if we want to instantiate more than a worker. It might not be compatible with [pooling](https://python-rq.org/docs/workers/#worker-pool), it will not support multi threading or asynchronous implementations. 

The Job already have a property named "worker_name".

src/aap_eda/wsapi/consumers.py
```diff
@@ -148,19 +147,14 @@ def handle_heartbeat(self, message: HeartbeatMessage) -> None:
...
-            instance.save(update_fields=["status", "updated_at"])
+            instance.save(update_fields=["updated_at"])
 
             activation = instance.activation
-            activation.status = ActivationStatus.RUNNING
-            activation.status_updated_at = message.reported_at
             activation.ruleset_stats[
                 message.stats["ruleSetName"]
             ] = message.stats
-            activation.save(
```
hsong-rh, line 161: if `status_updated_at` is removed from here, will `restart` be affected? it's used to calculate restart time interval, right?

src/aap_eda/services/ruleset/activate_rulesets.py
```diff
@@ -402,60 +413,63 @@ def activate_in_k8s(
         activation_id = activation.pk
         job_name = f"activation-job-{activation_id}-{activation_instance.id}"
         pod_name = f"activation-pod-{activation_id}-{activation_instance.id}"
+        secret_name = None
+        job_spec = None
+
+        if not reconnect:
```
hsong-rh, line 419: So `reconnect` logic here expects the k8s container, secret, service, etc. are all available when reconnection happens, right?

src/aap_eda/services/ruleset/activation_kubernetes.py
```diff
@@ -288,27 +289,35 @@ def run_activation_job(
...
+            while not done:
+                if_job_exists = self.batch_api.list_namespaced_job(
+                    namespace=namespace,
+                    label_selector=f"job-name={job_name}",
+                    timeout_seconds=0,
+                )
 
-            if not if_job_exists.items:
-                break
+                if not if_job_exists.items:
```
hsong-rh, line 310: If this always return `True`, we will stuck here. can we force it to break after sleep a while?

src/aap_eda/wsapi/consumers.py
bzwei: Restart logic depends on `status_updated_at` and `status == failed`. So if the status is still running the timestamp is not checked.

src/aap_eda/services/ruleset/activate_rulesets.py
bzwei: Yes. Here it skips creating container/secret/service etc. Later when it actually reconnects the container, it will check whether the container exists.

src/aap_eda/services/ruleset/activation_kubernetes.py
bzwei: This is the existing logic (when reconnect == False). We should fix it in another PR if it is a real bug.

src/aap_eda/api/views/activation.py
```diff
@@ -108,10 +108,7 @@ def create(self, request):
         activation["rules_count"] = 0
         activation["rules_fired_count"] = 0
 
-        if response.is_enabled:
```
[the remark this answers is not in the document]
bzwei: Good catch

hsong-rh (dismissed):
LGTM

src/aap_eda/tasks/ruleset.py
```diff
@@ -286,13 +303,41 @@ def _start_failed(now: timezone.datetime):
...
+        if not _worker_alive(activation.worker_name):
+            logger.info(
+                f"The activation worker for actvation_id {activation.id} is "
+                "dead. Will find another worker to reconnect to the instance"
+            )
+            job_id = _activation_job_id(activation.id)
+            job = job_from_queue(get_queue(name="activation"), job_id)
+            if job:
+                job.cancel()
+            activate(activation.id, "SCHEDULER", True)
```
Alex-Izquierdo, line 327: Can we use named parameters? It is hard to read specially when are involved booleans that you can not know what mean without look at the function signature

src/aap_eda/services/ruleset/activation_podman.py
```diff
@@ -93,75 +97,22 @@ def run_worker_mode(
...
-            activation.status_updated_at = now
-            activation.save(
-                update_fields=[
-                    "status",
-                    "status_updated_at",
-                    "is_valid",
-                    "modified_at",
-                ]
-            )
+            container = None
```
Alex-Izquierdo, line 103: This initialization must happen outside the try-except block because the variable is used in the `finally` section and if some exception is caught the variable is unbound and we will have an unexpected NameError exception.

src/aap_eda/services/ruleset/activation_podman.py
```diff
@@ -195,6 +149,98 @@ def run_worker_mode(
...
+            raise ReconnectContainerNotFound(
+                f"Container {activation_instance.activation_pod_id} not found"
+            )
+        container = containers[0]
+        if container.attrs["State"] != "running":
+            raise ReconnectContainerNotFound(
+                f"Container {activation_instance.activation_pod_id} not in "
+                "running state"
+            )
+        return container
```
Alex-Izquierdo, line 242: IMO the responsibility of this method is confusing. It is raising an exception if the container is not in running state. I feel at least should be named "_fetch_running_container" or it should be in charge of just fetching the container, the logic to raise ReconnectContainerNotFound should be executed by the caller when it is called under the block "if reconnect:" which makes more obvious that it is a logic related with the reconnection.

Alex-Izquierdo (commented):
Apologies if I come across as demanding or particular about naming, but I believe that since this logic is complex and could be challenging to read, especially for future contributors or even for us when we revisit this code, clarity is crucial. The easier the code is to read and understand, the better.

Please know all my comments are made with the best intentions. :)

src/aap_eda/tasks/ruleset.py
```diff
@@ -32,17 +32,29 @@
...
+    activation_id: int, requester: str = "User", reconnect: bool = False
+) -> None:
+    job_id = _activation_job_id(activation_id)
+    unique_enqueue(
+        "activation", job_id, _activate, activation_id, requester, reconnect
     )
 
 
-def _activate(activation_id: int, requester: str = "User") -> None:
+def _activation_job_id(activation_id: int) -> str:
```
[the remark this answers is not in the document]
bzwei: The Activation model represents data persisted in database. Here `activation_job_id` is the rq job id. I feel it is outside the model. The method is private because the job id is used only by methods in this module. We can move it to a common place if other module also needs to access it.

src/aap_eda/tasks/ruleset.py
```diff
@@ -32,17 +32,29 @@
...
+    activation_id: int, requester: str = "User", reconnect: bool = False
+) -> None:
+    action = "Reconnecting" if reconnect else "Activating"
     logger.info(
-        f"Activating activation id: {activation_id} requested "
-        f"by {requester}"
+        f"{action} activation id: {activation_id} requested " f"by {requester}"
+    )
+
+    job = job_from_queue(
```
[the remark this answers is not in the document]
bzwei: It is called when the job is executed by the rq worker, so the job must exist.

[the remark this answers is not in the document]
Alex-Izquierdo: ok!

src/aap_eda/tasks/ruleset.py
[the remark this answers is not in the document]
Alex-Izquierdo: I see your point. This is a common topic of discussion in software design. 
In my understanding a model is a representation of a domain entity that has both stored data and behaviors associated with that entity, serving as a bridge between the database and application logic. It can contain persistent attributes (stored in the database), computed attributes (would be this case), and methods that aren't stored but are derived or executed at runtime. In MVC architectures is quite usual, other architectures like domain-driven put a stronger emphasis on separate the business logic from the data model but we don't have such a pattern or entity for activations that operates on our business logic. Maybe we should have it. 


As it is now I consider 'job_id' intrinsic to the activation, given its pattern and the dependency of the application logic on it. I’m open to involving more team members for their insights to reach a consensus.  @mkanoor @hsong-rh @Dostonbek1 @benthomasson 

(I won’t hold back the PR over this; it’s a dialogue worth having, not a blocker.)

mkanoor:
@bzwei How do we guarantee that the worker name is unique, a user could start 1...n workers, when they come back up after a reboot or a restart how do we preserve the same names.
Since we are using process based jobs, wouldn't RQ handle these jobs and we could just query RQ to figure out if the job that it was running is alive? When the process ends for whatever reason I think RQ would trap it first before us.

https://python-rq.org/docs/jobs/#retrieving-jobs
I think we should depend as much as we can on RQ to for job management.

If we have an orphaned activation we should use RQ to delete the JOB or just attach to that podman instance to read the logs.

src/aap_eda/wsapi/consumers.py
```diff
@@ -148,19 +147,14 @@ def handle_heartbeat(self, message: HeartbeatMessage) -> None:
         ).first()
 
         if instance:
-            instance.status = ActivationStatus.RUNNING
             instance.updated_at = message.reported_at
-            instance.save(update_fields=["status", "updated_at"])
+            instance.save(update_fields=["updated_at"])
```
mkanoor, line 151: @bzwei we should do this in a separate PR

bzwei:
> @bzwei How do we guarantee that the worker name is unique, a user could start 1...n workers, when they come back up after a reboot or a restart how do we preserve the same names. Since we are using process based jobs, wouldn't RQ handle these jobs and we could just query RQ to figure out if the job that it was running is alive? When the process ends for whatever reason I think RQ would trap it first before us.
> 
> https://python-rq.org/docs/jobs/#retrieving-jobs I think we should depend as much as we can on RQ to for job management.
> 
> If we have an orphaned activation we should use RQ to delete the JOB or just attach to that podman instance to read the logs.

@mkanoor The worker name is an uuid, so it is unique. Unfortunately the job status reported back from RQ is not accurate. I have tested multiple times when the worker is dead the job however is still reported as running even after timeout period.

src/aap_eda/wsapi/consumers.py
bzwei: This change has been extracted in #435.

Alex-Izquierdo:
> > @bzwei How do we guarantee that the worker name is unique, a user could start 1...n workers, when they come back up after a reboot or a restart how do we preserve the same names. Since we are using process based jobs, wouldn't RQ handle these jobs and we could just query RQ to figure out if the job that it was running is alive? When the process ends for whatever reason I think RQ would trap it first before us.
> > https://python-rq.org/docs/jobs/#retrieving-jobs I think we should depend as much as we can on RQ to for job management.
> > If we have an orphaned activation we should use RQ to delete the JOB or just attach to that podman instance to read the logs.
> 
> @mkanoor The worker name is an uuid, so it is unique. Unfortunately the job status reported back from RQ is not accurate. I have tested multiple times when the worker is dead the job however is still reported as running even after timeout period.

Have you experimented with DEFAULT_JOB_MONITORING_INTERVAL and DEFAULT_WORKER_TTL ? We are using the default values. In general I feel all the defaults of RQ quite higher

bzwei:
replaced by the rework https://github.com/ansible/eda-server/pull/551

PR #411, opened 2023-09-19
fix: check encryption key at startup

Alex-Izquierdo:
Prevent the application from starting if the encryption key is incorrect. 
Jira: https://issues.redhat.com/browse/AAP-16219


Alex-Izquierdo:
discarded, we won't prevent the startup of the application

PR #413, opened 2023-09-20
Changed to use v6 Redis image for k8s deployment

hsong-rh:
The `quay.io/sclorg/redis-6-c9s:latest` redis image broke when deploying in k8s env. Changed it to use v6: `docker.io/library/redis:6`

### DIFF
--- a/tools/deploy/redis/deployment.yaml
+++ b/tools/deploy/redis/deployment.yaml
@@ -17,7 +17,7 @@ spec:
     spec:
       containers:
         - env: []
-          image: quay.io/sclorg/redis-6-c9s:latest
+          image: docker.io/library/redis:6
           name: redis
           ports:
             - containerPort: 6379


Alex-Izquierdo:
Is this not necessary also for docker-compose? I think @jshimkus-rh has issues as well

bzwei:
> > Is this not necessary also for docker-compose? I think @jshimkus-rh has issues as well
> 
> Really? my docker-compose works properly

docker-compose works for me on macbook M2

jshimkus-rh:
> > > Is this not necessary also for docker-compose? I think @jshimkus-rh has issues as well
> > 
> > 
> > Really? my docker-compose works properly
> 
> docker-compose works for me on macbook M2

It "works" for me as well although the image downloaded on my M1 Max is AMD64 rather than ARM64.
Unlike docker.io/library/redis:6, quay.io/sclorg/redis-6-c9s:latest does not appear to have any non-x86/AMD alternative architecture images.

Alex-Izquierdo (approved):

bzwei (approved):

PR #415, opened 2023-09-21
chore: update ui port for development environments

Alex-Izquierdo:
Follow up https://github.com/ansible/eda-server/pull/412
We added the new envvar, but it requires to change as well the default port. Now the UI image listens by default at the https default port (443) and is mapped to 8443 (by default system doesn't allow to map ports under 1024)

### DIFF
--- a/scripts/eda_kube.sh
+++ b/scripts/eda_kube.sh
@@ -190,7 +190,7 @@ port-forward() {
 port-forward-ui() {
   local _local_port=${1}
   local _svc_name=eda-ui
-  local _svc_port=8080
+  local _svc_port=8443
 
   log-debug "port-forward ${_svc_name} ${_local_port} ${_svc_port}"
   port-forward "${_svc_name}" "${_local_port}" "${_svc_port}"
--- a/scripts/eda_kube.sh
+++ b/scripts/eda_kube.sh
@@ -228,7 +228,7 @@ case ${CMD} in
   "clean") clean-deployment "${VERSION}";;
   "deploy") deploy "${VERSION}" ;;
   "port-forward-api") port-forward-api 8000 ;;
-  "port-forward-ui") port-forward-ui 8080 ;;
+  "port-forward-ui") port-forward-ui 8443 ;;
   "port-forward-pg") port-forward-pg 5432 ;;
   "eda-logs") get-eda-logs ;;
   "help") usage ;;
--- a/scripts/eda_kube.sh
+++ b/scripts/eda_kube.sh
@@ -190,7 +190,7 @@ port-forward() {
 port-forward-ui() {
   local _local_port=${1}
   local _svc_name=eda-ui
-  local _svc_port=8443
+  local _svc_port=8080
 
   log-debug "port-forward ${_svc_name} ${_local_port} ${_svc_port}"
   port-forward "${_svc_name}" "${_local_port}" "${_svc_port}"
@@ -228,7 +228,7 @@ case ${CMD} in
   "clean") clean-deployment "${VERSION}";;
   "deploy") deploy "${VERSION}" ;;
   "port-forward-api") port-forward-api 8000 ;;
-  "port-forward-ui") port-forward-ui 8443 ;;
+  "port-forward-ui") port-forward-ui 8080 ;;
   "port-forward-pg") port-forward-pg 5432 ;;
   "eda-logs") get-eda-logs ;;
   "help") usage ;;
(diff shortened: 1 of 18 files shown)


tools/docker/docker-compose-mac.yml
```diff
@@ -7,7 +7,7 @@ x-environment:
   - EDA_SECRET_KEY=secret
   - EDA_ALLOWED_HOSTS=['*']
   - EDA_DEPLOYMENT_TYPE=podman
-  - EDA_WEBSOCKET_BASE_URL=${EDA_WEBSOCKET_BASE_URL:-ws://host.containers.internal:8080}
+  - EDA_WEBSOCKET_BASE_URL=${EDA_WEBSOCKET_BASE_URL:-ws://host.containers.internal:8443}
```
bzwei, line 10: need `-wss`

bzwei (approved):

bzwei:
tested docker-compose on mac

tools/docker/docker-compose-mac.yml
```diff
@@ -21,7 +21,7 @@ services:
     image: "${EDA_UI_IMAGE:-quay.io/ansible/eda-ui:main}"
     environment: *common-env
     ports:
-      - '8080:8080'
+      - '8443:8080'
```
bzwei, line 24: should be `- '8443:433'`

tools/docker/docker-compose-stage.yaml
```diff
@@ -32,7 +32,7 @@ services:
     image: "${EDA_UI_IMAGE:-quay.io/ansible/eda-ui:main}"
     environment: *common-env
     ports:
-      - '8080:8080'
+      - '8443:8080'
```
bzwei, line 35: should be - '8443:433'

tools/docker/docker-compose-mac.yml
Alex-Izquierdo: Good catch, fixed.

mkanoor (dismissed):

mkanoor:
Tested with
```
podman machine init --cpus 2 --memory 2048 --disk-size 100 --now --image-path https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/37.20230122.3.0/aarch64/fedora-coreos-37.20230122.3.0-qemu.aarch64.qcow2.xz

docker-compose -f ./tools/docker/docker-compose-mac.yml up

```
With the latest fedora podman is broken

scripts/eda_kube.sh
```diff
@@ -228,7 +228,7 @@ case ${CMD} in
   "clean") clean-deployment "${VERSION}";;
   "deploy") deploy "${VERSION}" ;;
   "port-forward-api") port-forward-api 8000 ;;
-  "port-forward-ui") port-forward-ui 8080 ;;
+  "port-forward-ui") port-forward-ui 8443 ;;
```
bzwei, line 231: in line 193 we need to change the port from 8080 to 8443. 
even with this change I still got error:
```
(rulebook) billwei@bilwei-mac eda-server % kubectl port-forward svc/eda-ui 8443:8443
Forwarding from 127.0.0.1:8443 -> 443
Forwarding from [::1]:8443 -> 443
Handling connection for 8443
E0921 15:32:03.540365   11077 portforward.go:409] an error occurred forwarding 8443 -> 443: error forwarding port 443 to pod 442050f0ce9cacbb4b5b28d7d39d8e26f453b4c66b18e72d80978a9d260aa216, uid : exit status 1: 2023/09/21 19:32:03 socat[51063] E connect(5, AF=2 127.0.0.1:443, 16): Connection refused
```

bzwei: This error may have to do with https://github.com/kubernetes/kubectl/issues/1169

Alex-Izquierdo: line 193 updated, I can not reproduce the mentioned error.

Alex-Izquierdo: The problem is because minikube is not consuming the eda-ui image built from the project but we use our custom dockerfile/nginx conf so minikube doesn't need to be updated. 
I think we should consume the image provided by ansible-ui team, but that can be changed in a different PR.

ttuffin:
Verified on podman 4.6.2-1.

> With the latest fedora podman is broken

@mkanoor you can downgrade to `crun` package to get it working again. More info here: https://github.com/containers/podman/issues/19930#issuecomment-1715395823

ttuffin (approved):

bzwei (approved):

PR #416, opened 2023-09-21
[AAP-16639] Add ERROR status and message to activation and instance records

hsong-rh:
This PR is the first step to enhance activation: add error status and message to the DB. Next step in the separate pr will populate them when necessary.

https://issues.redhat.com/browse/AAP-16339


src/aap_eda/core/models/activation.py
```diff
@@ -72,6 +72,7 @@ class Meta:
     created_at = models.DateTimeField(auto_now_add=True, null=False)
     modified_at = models.DateTimeField(auto_now=True, null=False)
     status_updated_at = models.DateTimeField(null=True)
+    message = models.TextField(default="")
```
Alex-Izquierdo, line 75: This message is intended to be used only for a reason for the error state, using just "message" as a name might be confusing. It would be equivalent to the "import error" of [projects](https://github.com/ansible/eda-server/blob/main/src/aap_eda/core/models/project.py#L70). I suggest to use "error_message" instead. 

Following the same implementation in project, should this field be nullable with null as default?

mkanoor: @Alex-Izquierdo I think the message corresponds to the state it is in. We have multiple states we can't have a message column for every state. 
If we have "Starting State" => We could write which worker started the Job
Once it gets into the Error State the message would record the error
For Pending we could put a message => "Waiting for a worker to be available"

hsong-rh: @Alex-Izquierdo 
> Following the same implementation in project, should this field be nullable with null as default?

Yes, I will change the field accordingly.

For the message field name, I am prefer still using `message`, which is more generic, like @mkanoor mentioned.

Alex-Izquierdo: In that case I suggest to rename it to "status_message", my concern is the lack of meaning of the generic word "message"

Alex-Izquierdo: @hsong-rh According to the suggestion of the madhu, to ensure that the field "message" is aligned with the status I suggest to add a validator in the model and force the "message" field to be changed whenever the status field is changed.  In this way we can prevent errors such as returning a message that belongs to a previous transition.

mkanoor (approved):

src/aap_eda/core/models/activation.py
Alex-Izquierdo: It is fine to me to do it in a different PR. Regarding the default message, how can be the default message for so different states? or each state will have their own message? Anyway it can be discussed in that PR.

Alex-Izquierdo (approved):

src/aap_eda/core/models/activation.py
hsong-rh: Yes, each state will have its own default message, kind of like a map table. We can discuss more details in the next PR.

PR #419, opened 2023-09-22
Need redis v7

bzwei:
#413 rolled back to the wrong version. Previous correct version was 7.


hsong-rh:
@bzwei Why have to use redis v7? The reason to choose v6 was to keep the same major version as others

Alex-Izquierdo:
Indeed, why we need v7? that might be a problem. Our downstream builds and operator uses v6.

bzwei:
@hsong-rh @Alex-Izquierdo The previous version for minikube is v7. With v6 I got the following error:
```
1:C 22 Sep 2023 14:52:54.405 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
1:C 22 Sep 2023 14:52:54.405 # Redis version=6.2.13, bits=64, commit=00000000, modified=0, pid=1, just started
1:C 22 Sep 2023 14:52:54.405 # Warning: no config file specified, using the default config. In order to specify a config file use redis-server /path/to/redis.conf
1:M 22 Sep 2023 14:52:54.405 * monotonic clock: POSIX clock_gettime
1:M 22 Sep 2023 14:52:54.405 * Running mode=standalone, port=6379.
1:M 22 Sep 2023 14:52:54.405 # Server initialized
1:M 22 Sep 2023 14:52:54.408 # Can't handle RDB format version 11
1:M 22 Sep 2023 14:52:54.409 # Fatal error loading the DB: Invalid argument. Exiting.
```

Alex-Izquierdo:
I can not reproduce the issue. Based on the message `Fatal error loading the DB: Invalid argument. Exiting.` I wonder if might be because the PVC created with another image has not been deleted before the change to the current one. 

We should try to use the same version that we use in the operator and AAP unless it is impossible.

PR #420, opened 2023-09-25
[AAP-16339] Add default status_message to activations and instances when updating their status

hsong-rh:
This is the followup of https://github.com/ansible/eda-server/pull/416: update the `status_message` of activations/instances with default values when their status is changed.

### DIFF
--- a/src/aap_eda/core/enums.py
+++ b/src/aap_eda/core/enums.py
@@ -87,3 +87,17 @@ class CredentialType(DjangoStrEnum):
     REGISTRY = "Container Registry"
     GITHUB = "GitHub Personal Access Token"
     GITLAB = "GitLab Personal Access Token"
+
+
+ACTIVATION_STATUS_MESSAGE_MAP = {
+    ActivationStatus.PENDING: "Wait for a worker to be available to start activation",  # noqa: E501
+    ActivationStatus.STARTING: "Worker is starting activation",
+    ActivationStatus.RUNNING: "Container running activation",
+    ActivationStatus.STOPPING: "Activation is being disabled",
+    ActivationStatus.DELETING: "Activation is being deleted",
+    ActivationStatus.COMPLETED: "Activation has completed",
+    ActivationStatus.FAILED: "Activation has failed",
+    ActivationStatus.STOPPED: "Activation has stopped",
+    ActivationStatus.UNRESPONSIVE: "Activation is not responsive",
+    ActivationStatus.ERROR: "Activation is in an error state",
+}
--- a/src/aap_eda/core/models/activation.py
+++ b/src/aap_eda/core/models/activation.py
@@ -14,7 +14,16 @@
 
 from django.db import models
 
-from aap_eda.core.enums import ActivationStatus, RestartPolicy
+from aap_eda.core.enums import (
+    ACTIVATION_STATUS_MESSAGE_MAP,
+    ActivationStatus,
+    RestartPolicy,
+)
+from aap_eda.core.exceptions import (
+    StatusRequiredError,
+    UnknownStatusError,
+    UpdateFieldsRequiredError,
+)
 
 from .user import User
 
@@ -74,6 +83,57 @@ class Activation(models.Model):
     status_updated_at = models.DateTimeField(null=True)
     status_message = models.TextField(null=True, default=None)
 
+    def save(self, *args, **kwargs):
+        # when creating
+        if self._state.adding:
+            if self.status_message is None:
+                self._set_status_message()
+        else:
+            if not bool(kwargs) or "update_fields" not in kwargs:
+                raise UpdateFieldsRequiredError(
+                    "update_fields is required to use when saving "
+                    "due to race conditions"
+                )
+            else:
+                if "status" in kwargs["update_fields"]:
+                    self._is_valid_status()
+
+            if (
+                "status_message" in kwargs["update_fields"]
+                and "status" not in kwargs["update_fields"]
+            ):
+                raise StatusRequiredError(
+                    "status_message cannot be set by itself, "
+                    "it requires status and status_message together"
+                )
+            # when updating without status_message
+            elif (
+                "status" in kwargs["update_fields"]
+                and "status_message" not in kwargs["update_fields"]
+            ):
+                self._set_status_message()
+                kwargs["update_fields"].append("status_message")
+
+        super().save(*args, **kwargs)
+
+    def _set_status_message(self):
+        self.status_message = self._get_default_status_message()
+
+        if self.status == ActivationStatus.PENDING and not self.is_enabled:
+            self.status_message = "Activation is marked as disabled"
+
+    def _get_default_status_message(self):
+        try:
+            return ACTIVATION_STATUS_MESSAGE_MAP[self.status]
+        except KeyError:
+            raise UnknownStatusError(f"Status [{self.status}] is invalid")
+
+    def _is_valid_status(self):
+        try:
+            ActivationStatus(self.status)
+        except ValueError as error:
+            raise UnknownStatusError(error)
+
 
 class ActivationInstance(models.Model):
     class Meta:
@@ -92,6 +152,51 @@ class ActivationInstance(models.Model):
     ended_at = models.DateTimeField(null=True)
     activation_pod_id = models.TextField(null=True)
     status_message = models.TextField(null=True, default=None)
+
+    def save(self, *args, **kwargs):
+        # when creating
+        if self._state.adding:
+            if self.status_message is None:
+                self.status_message = self._get_default_status_message()
+        else:
+            if not bool(kwargs) or "update_fields" not in kwargs:
+                raise UpdateFieldsRequiredError(
+                    "update_fields is required to use when saving "
+                    "due to race conditions"
+                )
+            else:
+                if "status" in kwargs["update_fields"]:
+                    self._is_valid_status()
+
+            if (
+                "status_message" in kwargs["update_fields"]
+                and "status" not in kwargs["update_fields"]
+            ):
+                raise StatusRequiredError(
+                    "status_message cannot be set by itself, "
+                    "it requires status and status_message together"
+                )
+            # when updating without status_message
+            elif (
+                "status" in kwargs["update_fields"]
+                and "status_message" not in kwargs["update_fields"]
+            ):
+                self.status_message = self._get_default_status_message()
+                kwargs["update_fields"].append("status_message")
+
+        super().save(*args, **kwargs)
+
+    def _get_default_status_message(self):
+        try:
+            return ACTIVATION_STATUS_MESSAGE_MAP[self.status]
+        except KeyError:
+            raise UnknownStatusError(f"Status [{self.status}] is invalid")
+
+    def _is_valid_status(self):
+        try:
+            ActivationStatus(self.status)
+        except ValueError as error:
+            raise UnknownStatusError(error)
 
 
 class ActivationInstanceLog(models.Model):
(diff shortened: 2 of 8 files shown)


hsong-rh:
@mkanoor @Alex-Izquierdo Please review the default messages and discuss if the approach is doable.

src/aap_eda/core/enums.py
```diff
@@ -87,3 +87,17 @@ class CredentialType(DjangoStrEnum):
     REGISTRY = "Container Registry"
     GITHUB = "GitHub Personal Access Token"
     GITLAB = "GitLab Personal Access Token"
+
+
+ACTIVATION_STATUS_MESSAGE_MAP = {
+    ActivationStatus.PENDING: "Wait for worker to start the activation",
```
Alex-Izquierdo, line 93: This would not be accurate in all cases. Pending is our default initial status. when we create a disabled activation is this its status. We should have a "disabled" status.

src/aap_eda/core/enums.py
```diff
@@ -87,3 +87,17 @@ class CredentialType(DjangoStrEnum):
...
+    ActivationStatus.STARTING: "Worker starts the activation",
```
mkanoor, line 94: @hsong-rh I think the message would be better as "Worker starting activation"

src/aap_eda/core/enums.py
hsong-rh: @Alex-Izquierdo how about based on `is_enabled` flag, we set different default message? 
```
if is_enabled:
    status_message = "Wait for a worker to be available to start activation"
else:
   status_message = "Activation is marked as disabled"
```
In this case, we don't need to add `disabled` status

src/aap_eda/core/models/activation.py
```diff
@@ -74,6 +78,31 @@ class Meta:
...
 
+    def save(
+        self,
+        force_insert=False,
+        force_update=False,
+        using=None,
+        update_fields=None,
+    ):
+        # when creating
+        if update_fields is None and self.status == ActivationStatus.PENDING:
```
Alex-Izquierdo, line 89: Why are you adding pending status in this conditional? I think we want to check any transition and compare the current value with the old one. You can do this fetching the object from the DB before, but that adds an additional query to the DB, there is a way to do it without a query as is described here: https://stackoverflow.com/questions/23361057/django-comparing-old-and-new-field-value-before-saving/64116052#64116052

src/aap_eda/core/models/activation.py
```diff
@@ -74,6 +78,31 @@ class Meta:
     status_updated_at = models.DateTimeField(null=True)
     status_message = models.TextField(null=True, default=None)
 
+    def save(
+        self,
+        force_insert=False,
+        force_update=False,
+        using=None,
+        update_fields=None,
```
Alex-Izquierdo, line 86: I think we should not limit the api of the method overriding the params. Instead we should use `*args, **kwargs` and passing through them in the call to the parent.

src/aap_eda/core/enums.py
Alex-Izquierdo: I feel a pending state for an activation that is not enabled is confusing and a bad UX. But that is another discussion. For the scope of the PR, that conditional is what we need to have useful messages, IMO.

src/aap_eda/core/models/activation.py
hsong-rh: We just update `status_message` whenever `status` is changed, no need to compare with the old value.

src/aap_eda/core/models/activation.py
```diff
@@ -74,6 +78,26 @@ class Meta:
     status_updated_at = models.DateTimeField(null=True)
     status_message = models.TextField(null=True, default=None)
 
+    def save(self, *args, **kwargs):
```
Alex-Izquierdo, line 81: It is missing the logic to ensure the status_message if you call "save" method without update_fields (only ensures the pending status) that's why I say you have to compare to the previous value, to check if the instance is updating that field or not.

src/aap_eda/core/models/activation.py
```diff
@@ -93,6 +117,19 @@ class Meta:
     activation_pod_id = models.TextField(null=True)
     status_message = models.TextField(null=True, default=None)
 
+    def save(self, *args, **kwargs):
+        if kwargs is None or kwargs.get("update_fields") is None:
```
Alex-Izquierdo, line 121: with this logic as it is, you would override any status_message by its default every time the save method is called without update_fields. You need to compare with the previous value to know if status_message should have a new default message or not.

mkanoor: @hsong-rh Can we just use the elif portion of this and not use the if portion? Because we want to set the status and the status_message as a pair all the time. if the caller has not provided the status_message we provide the default like the elif portion is doing it.

mkanoor: @hsong-rh I think we might also want to prevent someone from just changing the status_message without the status variable defined. If it's currently in Error state, we can't just have an update_field with status_message set to "Everything is OK". If they have to change the status_message they should change the status and the status_message.

mkanoor: There was also the discussion last week that since we many have too many processes (api-worker, activation-worker, default-worker) working with the Activation and ActivationInstance we should not allow saves without update_fields, since one process can end up clobbering a previous workers updates.

src/aap_eda/core/models/activation.py
hsong-rh: The missing logic is added

src/aap_eda/core/models/activation.py
hsong-rh: The select_for_update/atomic transaction is used in the outside block, here the activation record is locked already, and update the fields. 

@mkanoor @Alex-Izquierdo 
1. now normal `save()` without any parameters will raise a validation error;
2. `save(update_fields=["status_message"])` will raise a validation error;
3. `save(update_fields=["status"])` will also update `status_message` with default if it's none

mkanoor:
@hsong-rh Can we add tests for Activation model with different save options?

src/aap_eda/core/models/activation.py
```diff
@@ -74,6 +80,46 @@ class Meta:
...
     status_message = models.TextField(null=True, default=None)
 
+    def save(self, *args, **kwargs):
+        # when creating
+        if self._state.adding:
+            if self.status_message is None:
+                self._set_status_message()
+        else:
+            if kwargs is None:
+                raise ValidationError(
```
Alex-Izquierdo, line 90: This is not a validation error. Can you use a custom exception UpdateFieldsRequiredError?

src/aap_eda/core/models/activation.py
```diff
@@ -74,6 +80,46 @@ class Meta:
...
+                        "update_fields is required to use when saving "
+                        "due to race conditions"
+                    )
+                )
+            # when updating without status
+            elif (
+                "status_message" in kwargs["update_fields"]
+                and "status" not in kwargs["update_fields"]
+            ):
+                raise ValidationError(
```
Alex-Izquierdo, line 101: This is not a validation error. Can you use a custom exception StatusMessageRequiredError?

src/aap_eda/core/models/activation.py
```diff
@@ -74,6 +80,46 @@ class Meta:
     status_updated_at = models.DateTimeField(null=True)
     status_message = models.TextField(null=True, default=None)
 
+    def save(self, *args, **kwargs):
+        # when creating
+        if self._state.adding:
+            if self.status_message is None:
+                self._set_status_message()
+        else:
+            if kwargs is None:
```
Alex-Izquierdo, line 89: save method supports several parameters, we want to enforce the update_fields param, dont?

src/aap_eda/core/models/activation.py
```diff
@@ -74,6 +80,46 @@ class Meta:
     status_updated_at = models.DateTimeField(null=True)
     status_message = models.TextField(null=True, default=None)
 
+    def save(self, *args, **kwargs):
```
Alex-Izquierdo, line 83: This is an important restriction on the model and it should have their own tests to ensure the expected behavior.

src/aap_eda/core/models/activation.py
```diff
@@ -93,6 +139,44 @@ class Meta:
...
+            elif (
+                "status" in kwargs["update_fields"]
+                and "status_message" not in kwargs["update_fields"]
+            ):
+                self.status_message = ACTIVATION_STATUS_MESSAGE_MAP[
+                    self.status
+                ]
+                kwargs["update_fields"].append("status_message")
+
+        super().save(*args, **kwargs)
```
mkanoor, line 178: @Alex-Izquierdo @hsong-rh 
Would it make sense to wrap this in an select_for_update or should that be done by the caller?

Alex-Izquierdo: I would leave it to the caller, let's see how this restriction works for now.

src/aap_eda/core/models/activation.py
```diff
@@ -74,6 +82,42 @@ class Meta:
...
+                raise UpdateFieldsRequiredError(
+                    "update_fields is required to use when saving "
+                    "due to race conditions"
+                )
+            # when updating without status
+            elif (
+                "status_message" in kwargs["update_fields"]
+                and "status" not in kwargs["update_fields"]
+            ):
+                raise StatusMessageRequiredError(
```
mkanoor, line 101: @hsong-rh This should be StatusRequiredError instead of StatusMessageRequiredError. Because we are saying that status is mandatory, status_message is optional and status_message can't be set by itself. Maybe the error message should say "status_message cannot be set by itself, it requires status and status_message together".

tests/integration/core/test_activation.py
```diff
@@ -0,0 +1,90 @@
...
+        ActivationStatus.STOPPING,
+        ActivationStatus.DELETING,
+        ActivationStatus.COMPLETED,
+        ActivationStatus.FAILED,
+        ActivationStatus.STOPPED,
+        ActivationStatus.UNRESPONSIVE,
+        ActivationStatus.ERROR,
+        ActivationStatus.PENDING,
+    ]:
+        activation.status = status
```
mkanoor, line 74: @hsong-rh Do we have tests for an Invalid Status Name, that gets passed?

hsong-rh: @mkanoor Added in the tests

src/aap_eda/core/models/activation.py
```diff
@@ -74,6 +83,48 @@ class Meta:
...
+        self.status_message = self._get_default_status_message()
+
+        if self.status == ActivationStatus.PENDING and not self.is_enabled:
+            self.status_message = "Activation is marked as disabled"
+
+    def _get_default_status_message(self):
+        try:
+            return ACTIVATION_STATUS_MESSAGE_MAP[self.status]
+        except KeyError:
+            raise UnknownStatusError(f"Status [{self.status}] is invalid")
```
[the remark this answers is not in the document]
hsong-rh: @mkanoor `_set_status_message` is called in both creating and updating time, also https://github.com/ansible/eda-server/pull/420/files#diff-e03abd6702de56997bfecb4910d346ed584c41874856257890ec949b8484ad39R57 is the place for updating test

mkanoor (approved):

src/aap_eda/core/models/activation.py
```diff
@@ -74,6 +83,57 @@ class Meta:
     status_updated_at = models.DateTimeField(null=True)
     status_message = models.TextField(null=True, default=None)
 
+    def save(self, *args, **kwargs):
+        # when creating
+        if self._state.adding:
+            if self.status_message is None:
+                self._set_status_message()
+        else:
+            if not bool(kwargs):
```
Alex-Izquierdo, line 92: We could call the method with any other param but without the update_fields and it would pass this check. If want to enforce the use of update_field, we must check for its existence.

Alex-Izquierdo (approved):

mkanoor (approved):

PR #424, opened 2023-09-26
[AAP-16339] Update to ERROR state when activations fail in validation

hsong-rh:
This PR is the follow-up of both https://github.com/ansible/eda-server/pull/416 and https://github.com/ansible/eda-server/pull/420: 
1. Extract validation logic for creating activations in one place. The validation contains the following:
        - awx token: only one token is allowed
        - decision environment needs to exist; 
        - if the rulebook is related, which needs to be verified (only check during creation time)
        - if extra var is related, which needs to be verified
3. Add a validation check at the beginning of `restart, enable, and monitor`, and update the activation to ERROR status if the validation fails.


src/aap_eda/services/validation.py
```diff
@@ -0,0 +1,97 @@
...
+    elif tokens > 1:
+        raise ValidationError(
+            _(
+                "More than one controller token found, "
+                "currently only 1 token is supported"
+            )
+        )
+
+
+def _validate_websocket() -> None:
```
Alex-Izquierdo, line 50: I think this check is wrong.  We are rejecting a valid request for a value that is set in the server and the user has no control. This would be a server error, not a client error. This check should follow a different flow and returns a 500 response

src/aap_eda/services/validation.py
```diff
@@ -0,0 +1,97 @@
...
+
+def _validate_rulebook(rulebook_id: int) -> None:
+    try:
+        models.Rulebook.objects.get(id=rulebook_id)
+    except models.Rulebook.DoesNotExist:
+        raise ValidationError(_("No rulebook specified"))
+
+
+def _validate_extra_var(extra_var_id: int) -> None:
+    if not extra_var_id:
```
Alex-Izquierdo, line 93: wrong conditional.

src/aap_eda/services/validation.py
```diff
@@ -0,0 +1,97 @@
...
+        try:
+            models.ExtraVar.objects.get(id=extra_var_id)
+        except models.ExtraVar.DoesNotExist:
+            raise ValidationError(_("No extra_var specified"))
```
Alex-Izquierdo, line 97: The message is incorrect. Extra var id is optional, this function should check if the extra_var exists when it is provided.

Alex-Izquierdo (commented):
IMO we are doing a misuse of the "create" method of the serializer.  Create is for the purpose to transform the data into a model not for validation. 
For validation we should implement the method "validate" which is called by the "is_valid" method: 
ref: https://www.django-rest-framework.org/api-guide/serializers/#validation

src/aap_eda/services/validation.py
```diff
@@ -0,0 +1,97 @@
...
+                "currently only 1 token is supported"
+            )
+        )
+
+
+def _validate_websocket() -> None:
+    ws_url = f"{settings.WEBSOCKET_BASE_URL}{ACTIVATION_PATH}"
+    parsed_url = urllib.parse.urlparse(ws_url)
+
+    if parsed_url.scheme not in ["ws", "wss"]:
```
mkanoor, line 54: @hsong-rh This doesn't seem correct, we should be doing this validation when the app starts up with something like this https://www.dynaconf.com/validation/#validating-in-python-programmatically. If the settings has a bogus value it needs to validated at startup. We shouldn't have to do this with every activation instance, since the settings  doesn't change with each activation.

mkanoor: We can get rid of this check and do the validation at startup in a different PR.

Alex-Izquierdo: I proposed to check the encryption key at startup and it was rejected. I think this is the same scenario. So we should not allow any activation if the ws url is wrong. The check is fine but as I said in a previous comment, this is not a validation error. It should follow a different workflow and returns 500, a wrong ws url is a server error, not a client error.

hsong-rh:
> IMO we are doing a misuse of the "create" method of the serializer. Create is for the purpose to transform the data into a model not for validation. 

Seems like the best practice of DRF is "Thin views, Thick serializers" (https://stackoverflow.com/questions/41094013/when-to-use-serializers-create-and-modelviewsets-perform-create)

Alex-Izquierdo:
> > IMO we are doing a misuse of the "create" method of the serializer. Create is for the purpose to transform the data into a model not for validation.
> 
> Seems like the best practice of DRF is "Thin views, Thick serializers" (https://stackoverflow.com/questions/41094013/when-to-use-serializers-create-and-modelviewsets-perform-create)

I'm not sure what do you mean. That's is not incompatible with I'm saying. I'm saying to perform the validations in the serializer not in the view. But in the right method for it, not in create.

src/aap_eda/services/validation.py
hsong-rh: @mkanoor @Alex-Izquierdo Changed to keep the websocket checking as it is now, will change it to validate at startup in separate PR

mkanoor:
@hsong-rh If there are no extra vars I am seeing this error
Screenshot 2023-10-02 at 11 05 20 AM

src/aap_eda/services/validation.py
```diff
@@ -0,0 +1,114 @@
...
+from django.db import transaction
+from django.utils.translation import gettext_lazy as _
+
+from aap_eda.core import models
+from aap_eda.core.enums import ActivationStatus
+
+logger = logging.getLogger(__name__)
+
+
+def validate_activation(activation_id: int) -> bool:
```
mkanoor, line 26: @hsong-rh Does this have to be a service, its not talking to any external server. Can't it just be a mixin that can be called at creation time and post creation.

mkanoor:
@hsong-rh @Alex-Izquierdo Would it make sense to store the token as a hidden relation field in the Activation object. I am thinking in the future we could have different tokens and the user would be able to pick one of them or None (if the rulebook has custom actions and no run_job_template). During this new field migration if we find that there are activations we could fill in the 
   * token_id = The id of the controller token

This way we would have a consistent way to validate the related fields.

   * decision_environment_id
   * rulebook_id
   * token_id
   * user_id

This can be done at creation time and also post creation.

The current create serializer doesn't seem to be validating everything in the serializer and the logic is split over in the view and the serializer. This is the relationship validation in the [view](https://github.com/ansible/eda-server/blob/5f0c10999ffb133a37d17ba167f8f1684b23fd87/src/aap_eda/api/views/activation.py#L326) I think this should have been in the serializer and it should have been able to validate the existence of these objects else raise a ValidationError.

hsong-rh:
@mkanoor @Alex-Izquierdo I read the DRF doc for serializers. Seems like when we decided to append `-id` to relation fields (https://issues.redhat.com/browse/AAP-10811), we then lost the type of `serializers.PrimaryKeyRelatedField`, which is auto generated by `model.serializers` for us for the relations. Our serializer classes now use `serializers.IntegerField` for foreign keys. Default validation has no effects on them. We may need to add individual field validator for each foreign key if we want to keep the current way.

@mkanoor Validators are related with serializers and are only used in creating and updating DB records.  Some of our cases are different. We want to make sure if the activation is in good state before it will be restarted, enabled, or reconnected. I don't think we can define a single logic and use it in both creation and post creation time. Maybe part of logic in validators, and part of in mixins, how do think?

Alex-Izquierdo:
> @hsong-rh @Alex-Izquierdo Would it make sense to store the token as a hidden relation field in the Activation object. I am thinking in the future we could have different tokens and the user would be able to pick one of them or None (if the rulebook has custom actions and no run_job_template). During this new field migration if we find that there are activations we could fill in the
> 
> * token_id = The id of the controller token
> 
> This way we would have a consistent way to validate the related fields.
> 
> * decision_environment_id
> * rulebook_id
> * token_id
> * user_id
> 
> This can be done at creation time and also post creation.
> 
> The current create serializer doesn't seem to be validating everything in the serializer and the logic is split over in the view and the serializer. This is the relationship validation in the [view](https://github.com/ansible/eda-server/blob/5f0c10999ffb133a37d17ba167f8f1684b23fd87/src/aap_eda/api/views/activation.py#L326) I think this should have been in the serializer and it should have been able to validate the existence of these objects else raise a ValidationError.

@mkanoor @hsong-rh I don't have a strong opinion on it. Eventually we will have multiple tokens, that's for sure. What would be the benefit of that against keeping the related id and performs validations as we already do with DE's and credentials? Why a token should be different?  I think if we implement the capability to pick a token for an activation, that field should not be immutable for rotation purposes. What would be the flow for rotating the keys?

src/aap_eda/services/validation.py
Alex-Izquierdo: I agree that conceptually speaking is not a service. But why a mixin? It may make harder the re usability if we are going to use the same logic in other places, like at restarting and activation.

src/aap_eda/api/serializers/activation.py
```diff
@@ -120,12 +115,18 @@ class Meta:
...
-            rulebook = models.Rulebook.objects.get(
-                pk=validated_data["rulebook_id"]
-            )
-        except models.Rulebook.DoesNotExist:
-            raise IntegrityError
+        rulebook_id = validated_data["rulebook_id"]
+        decision_environment_id = validated_data["decision_environment_id"]
+        extra_var_id = validated_data.get("extra_var_id")
+
+        validate_activation_creation(
```
Alex-Izquierdo, line 122: This a response of [this comment](https://github.com/ansible/eda-server/pull/424#issuecomment-1745585351):
I'm not saying that we must perform field validation. This seems clearly an object validation case. I may missing something of your reasoning, my point is that such a validation should not be performed in the create method. We already make a call to the .is_valid() method as is the standard the way and part of this call is the object validation method "validate()". My point is that this logic should be there:
https://www.django-rest-framework.org/api-guide/serializers/#object-level-validation

hsong-rh:
@Alex-Izquierdo @Alex-Izquierdo Based on our discussion, now all validations are checked by validators in serializers.

mkanoor:
@hsong-rh Since token is a non field we get this error
```
 "status_message": "{'non_field_errors': 'No controller token specified'}"
  ```
This is because we don't have a token relationship to Activation, we can change the non_field_errors to msg or error? It would have been ideal if we had the token relationship.

**Comment by hsong-rh** [comment]
> @hsong-rh Since token is a non field we get this error
> 
> ```
>  "status_message": "{'non_field_errors': 'No controller token specified'}"
> ```
> 
> This is because we don't have a token relationship to Activation, we can change the non_field_errors to msg or error? It would have been ideal if we had the token relationship.

OK, now the status message is changed as:
```
 "status_message": "{'errors': 'No controller token specified'}"
```

File: src/aap_eda/api/serializers/activation.py [hunk 7]
```diff
@@ -220,48 +231,130 @@ class Meta:
...
+            "modified_at",
+        ]
+        read_only_fields = [
+            "id",
+            "created_at",
+            "modified_at",
+        ]
+
+
+def get_rules_count(ruleset_stats: dict) -> tuple[int, int]:
```
Alex-Izquierdo, line 332: Not sure why the tests are passing, because you are renaming this function and removing the old `_get_rules_count` which is used here: https://github.com/ansible/eda-server/blob/bcceb144fe74dc2fa04a427745adab03a35819e4/tests/integration/api/test_activation.py#L619 but I don't see in the PR that line modified.

Alex-Izquierdo: Oh, I see, we are duplicating the code:
https://github.com/ansible/eda-server/blob/bcceb144fe74dc2fa04a427745adab03a35819e4/tests/integration/api/test_activation.py#L655

All of this seems quite wrong, we assert a value set by a copy of the function.
I don't say to change these tests in this PR but at least add a TODO comment.

Alex-Izquierdo (commented):
Static function `is_activation_valid` should have their own tests, specially when are are mocking it in all the integration tests involved. 

Same for `get_rules_count`. Any static function should have their own tests.

src/aap_eda/api/serializers/activation.py
```diff
@@ -220,48 +231,130 @@ class Meta:
...
+def parse_validation_errors(errors: dict) -> str:
+    messages = {}
+
+    for key in errors:
+        if key == "non_field_errors":
+            messages["errors"] = str(errors[key][0])
+        else:
+            messages[key] = str(errors[key][0])
+
+    return str(messages)
```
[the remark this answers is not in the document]
Alex-Izquierdo: Highlights:
- dict comprehensions are more efficient than regular loops. 
- No need to compare the `non_field_errors` key in each iteration as keys are unique within a dict and key access is a O(1) complexity operation.

src/aap_eda/api/views/activation.py
```diff
@@ -243,6 +215,17 @@ def enable(self, request, pk):
...
             return Response(status=status.HTTP_409_CONFLICT)
 
+        valid, error = is_activation_valid(activation)
+        if not valid:
+            activation.status = ActivationStatus.ERROR
+            activation.status_message = error
+            activation.save(update_fields=["status", "status_message"])
+            logger.error(f"Failed to enable {activation.name}: {error}")
+
+            return Response(
```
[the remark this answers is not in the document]
hsong-rh: Based on DRF validation docs (https://www.django-rest-framework.org/api-guide/serializers/#validation), validation errors will return HTTP 400 Bad Request responses by default.

src/aap_eda/api/serializers/activation.py
hsong-rh: Removed the duplicate and added a test for `get_rules_count` function

src/aap_eda/api/views/activation.py
```diff
@@ -87,48 +68,41 @@ class ActivationViewSet(
...
-        serializer.is_valid(raise_exception=True)
-
-        try:
-            response = serializer.create(
-                serializer.validated_data, request.user
+        context = {"request": request}
+        serializer = serializers.ActivationCreateSerializer(
+            data=request.data, context=context
+        )
+        valid = serializer.is_valid()
```
mkanoor, line 82: @hsong-rh Can we just have it raise the exception wouldn't that be converted to a bad request?

hsong-rh: The reason we removed `raise_exception=True` from `is_valid` is because we want to rename `non_field_errors`

src/aap_eda/api/serializers/activation.py
```diff
@@ -98,13 +88,53 @@ class Meta:
...
+        validators=[validation.check_if_rulebook_exists]
+    )
     extra_var_id = serializers.IntegerField(required=False, allow_null=True)
-    decision_environment_id = serializers.IntegerField()
+    decision_environment_id = serializers.IntegerField(
+        validators=[validation.check_if_de_exists]
+    )
+    user = serializers.HiddenField(default=serializers.CurrentUserDefault())
+
+    def validate(self, data):
```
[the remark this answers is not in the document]
Alex-Izquierdo: What about this comment? you have resolved it without any answer.

Alex-Izquierdo (approved):

src/aap_eda/api/serializers/activation.py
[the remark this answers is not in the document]
hsong-rh: @Alex-Izquierdo Validation is added for credentials when the decision environment has it.

PR #427, opened 2023-09-28
feat: aap-16610 handle deletion of dependent credentials

Alex-Izquierdo:
- Filter activations by credential id
- New force flag at delete credentials, prevent deletion of dependent credentials without flag. 
Jira: https://issues.redhat.com/browse/AAP-16610

### DIFF
--- a/src/aap_eda/api/views/credential.py
+++ b/src/aap_eda/api/views/credential.py
@@ -98,6 +98,7 @@ class CredentialViewSet(
     CreateModelMixin,
     PartialUpdateOnlyModelMixin,
     mixins.RetrieveModelMixin,
+    mixins.DestroyModelMixin,
     mixins.ListModelMixin,
     viewsets.GenericViewSet,
 ):
--- a/src/aap_eda/api/views/credential.py
+++ b/src/aap_eda/api/views/credential.py
@@ -98,7 +98,6 @@ class CredentialViewSet(
     CreateModelMixin,
     PartialUpdateOnlyModelMixin,
     mixins.RetrieveModelMixin,
-    mixins.DestroyModelMixin,
     mixins.ListModelMixin,
     viewsets.GenericViewSet,
 ):
@@ -147,4 +146,5 @@ class CredentialViewSet(
                 "and cannot be deleted. If you want to force delete, "
                 "please add /?force=true query param."
             )
-        return super().destroy(request, *args, **kwargs)
+        self.perform_destroy(credential)
+        return Response(status=status.HTTP_204_NO_CONTENT)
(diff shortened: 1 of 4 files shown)


src/aap_eda/api/views/credential.py
```diff
@@ -117,3 +126,25 @@ def get_serializer_class(self):
 
     def get_response_serializer_class(self):
         return serializers.CredentialSerializer
+
+    def destroy(self, request, *args, **kwargs):
```
hsong-rh, line 130: @Alex-Izquierdo We need to change urls.py to add this endpoint, right?

Alex-Izquierdo: This is not a new endpoint, just we are overriding the delete logic.

src/aap_eda/api/views/credential.py
```diff
@@ -117,3 +126,25 @@ def get_serializer_class(self):
...
+            decision_environment__credential=credential
+        ).exists()
+
+        if is_used and not force:
+            raise exceptions.Forbidden(
+                "Credential is being used by Activations "
+                "and cannot be deleted. If you want to force delete, "
+                "please add /?force=true query param."
+            )
+        return super().destroy(request, *args, **kwargs)
```
Dostonbek1, line 150: Replacing it with the following would save us calling `self.get_object()` once again
```
self.perform_destroy(instance)
return Response(status=status.HTTP_204_NO_CONTENT)
```
I think `self.get_object()` makes another query to fetch the object (which we already have, `credential`), which is redundant. Thoughts?

tests/integration/api/test_activation.py
```diff
@@ -365,6 +374,25 @@ def test_list_activations_filter_decision_environment_id(client: APIClient):
...
+    fks = create_activation_related_data()
+    create_activation(fks)
+    credential_id = fks["credential_id"]
+
+    url = f"{api_url_v1}/activations/?credential_id={credential_id}"
+    response = client.get(url)
+    assert response.status_code == status.HTTP_200_OK
+    assert len(response.data["results"]) == 1
+
+    url = f"{api_url_v1}/activations/?credential_id=2"
```
Dostonbek1, line 390: This might fail if we have more than 1 credentials for tests in the future. Should we use an arbitrary number like 0 to work for all cases?

src/aap_eda/api/views/credential.py
Alex-Izquierdo: The point to me was not breaking down the process flow and continue with the flow of the superclass. Override a specific operation doesn't need necessarily to replace the whole common logic. Efficiency vs extensibility and maintenance.

Alex-Izquierdo: changed

tests/integration/api/test_activation.py
Alex-Izquierdo: I don't think so, the test uses a clean DB. I have increased the number just in case. Anyway I want to rework the fixtures thing in later PR's

Dostonbek1 (approved):

hsong-rh (approved):

PR #429, opened 2023-09-29
chore(config): adjust liveness default values

Alex-Izquierdo:
Adjust values for liveness check as we discussed and agreed. It also changes how the the timeout is set. Timeout would have unexpected outcomes if is lower than the liveness check. The timeout is the time that we wait for the last expected liveness check, so it should be sum to that value.


bzwei:
We need to more thoroughly test this new setting. Together with #410, we want to reconnect first as much as possible if the service was interrupt for a long time for upgrade for example.Also we may want to first implement a pure heartbeat without stats at the ansible-rulebook side as agreed.

PR #430, opened 2023-10-02
(chore) Add api fallback exception handler.

jshimkus-rh:
Prevents stack trace on unexpected exceptions in deployment. If there is no already extant Response instance and debug is not enabled generates an internal server error Response containing contextual information as to the origin of the exception (what api and parameters were specified).

### DIFF
--- a/tests/integration/api/test_fallback_handler.py
+++ b/tests/integration/api/test_fallback_handler.py
@@ -19,11 +19,8 @@ def raise_exception():
 @mock.patch("aap_eda.api.views.tasks.list_jobs", new=raise_exception)
 def test_debug_unexpected_exception(client: APIClient, settings):
     settings.DEBUG = True
-    try:
+    with pytest.raises(FallbackException):
         client.get(f"{api_url_v1}/tasks/")
-        raise Exception
-    except FallbackException:
-        pass
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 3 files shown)


bzwei (approved):

tests/integration/api/test_fallback_handler.py
```diff
@@ -0,0 +1,34 @@
...
+    settings.DEBUG = True
+    with pytest.raises(FallbackException):
+        client.get(f"{api_url_v1}/tasks/")
+
+
+@pytest.mark.django_db
+@mock.patch("aap_eda.api.views.tasks.list_jobs", new=raise_exception)
+def test_non_debug_unexpected_exception(client: APIClient, settings):
+    settings.DEBUG = False
+    response = client.get(f"{api_url_v1}/tasks/")
```
Alex-Izquierdo, line 30: Can you use a different endpoint for the test? This endpoint is not used at all and is going to be removed.

Alex-Izquierdo (approved):

tests/integration/api/test_fallback_handler.py
jshimkus-rh: Switched to 'projects'.

Alex-Izquierdo (approved):

bzwei (approved):

PR #446, opened 2023-10-20
Arbitrate queued activation requests

bzwei:
If multiple requests have been queued for an activation we don't have to execute them one by one. The arbitrate logic consolidates them base on the sequence only, regardless of the current status of the activation. Many of the consolidations may not actually happen because the view already has validation based on the current status. Furthermore the manager has validations too before the request is actually executed.

### DIFF
--- a/src/aap_eda/tasks/activation_request_queue.py
+++ b/src/aap_eda/tasks/activation_request_queue.py
@@ -23,9 +23,10 @@ def push(activation_id: int, request: ActivationRequest) -> None:
 
 
 def peek_all(activation_id: int) -> list[ActivationRequestQueue]:
-    return ActivationRequestQueue.objects.filter(
+    requests = ActivationRequestQueue.objects.filter(
         activation_id=activation_id
     ).all()
+    return _arbitrate(requests)
 
 
 def pop_until(activation_id: int, queue_id: int) -> None:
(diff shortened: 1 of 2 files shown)


tests/integration/tasks/test_activation_request_queue.py
```diff
@@ -42,19 +42,79 @@ def init_activations():
...
+        {
+            "queued": [
+                ActivationRequest.RESTART,
+                ActivationRequest.STOP,
+                ActivationRequest.START,
+            ],
+            "dequeued": [
+                ActivationRequest.STOP,
+                ActivationRequest.START,
+            ],
```
Alex-Izquierdo, line 98: I think in this case should be contrary. Restart is stop+start, so the dequeued output should be restart, why schedule a stop and later a start when we can do both in the same verb manager.restart()

src/aap_eda/tasks/activation_request_queue.py
```diff
@@ -39,3 +40,50 @@ def list_activations() -> list[int]:
...
+def _arbitrate(
+    requests: list[ActivationRequestQueue],
+) -> list[ActivationRequestQueue]:
+    if len(requests) < 2:
+        return requests
+
+    ref_request = None
+    qualified_requests = []
+    starts = [ActivationRequest.START, ActivationRequest.RESTART]
+    for request in requests:
```
Alex-Izquierdo, line 54: Suggestion:

src/aap_eda/tasks/activation_request_queue.py
```diff
@@ -23,9 +23,10 @@ def push(activation_id: int, request: ActivationRequest) -> None:
 
 
 def peek_all(activation_id: int) -> list[ActivationRequestQueue]:
-    return ActivationRequestQueue.objects.filter(
+    requests = ActivationRequestQueue.objects.filter(
         activation_id=activation_id
     ).all()
+    return _arbitrate(requests)
```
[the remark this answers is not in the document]
bzwei: Per our discussion, at this moment we don't need a very fancy arbitrator. We can refactor it if we really have such need.

tests/integration/tasks/test_activation_request_queue.py
bzwei: Per our discussion there will no RESTART, STOP, START combination in real.

PR #458, opened 2023-10-24
refactor: change exceptions in K8S

hsong-rh:


src/aap_eda/services/activation/engine/kubernetes.py
```diff
@@ -419,90 +472,47 @@ def _wait_for_pod_to_start(self) -> ActivationStatus:
                             exit_code = statuses[0].state.terminated.exit_code
                             reason = statuses[0].state.terminated.reason
                             if exit_code in SUCCESSFUL_EXIT_CODES:
-                                raise DeactivationException(
-                                    f"Container exited with code {exit_code}"
-                                )
+                                # Failed is due to deactivation removed the pod
+                                status = ActivationStatus.STOPPED
```
Alex-Izquierdo, line 476: Podman is returning COMPLETED, we should be aligned on this to catch from the manager always the same status for a succesful exit code.

Alex-Izquierdo: We need to document the mapping of activation status to container status.

hsong-rh: Yeah, here we need more tests. When the pod stopped with `Failed` status and its exit code was in `SUCCESSFUL_EXIT_CODES`, the previous logic means the pod was removed by `disable` or `delete`. Its status is marked as `STOPPED`

PR #468, opened 2023-10-26
chore: ensure rq workers use our custom classes

Alex-Izquierdo:
rqworker cli overrides the params with defaults at the worker initialization preventing to use our extended ones.

### DIFF
--- a/src/aap_eda/core/tasking/__init__.py
+++ b/src/aap_eda/core/tasking/__init__.py
@@ -119,12 +119,9 @@ class DefaultWorker(_Worker):
         prepare_for_work: bool = True,
         serializer: Optional[SerializerProtocol] = None,
     ):
-        if job_class is None:
-            job_class = Job
-        if queue_class is None:
-            queue_class = Queue
-        if serializer is None:
-            serializer = JSONSerializer
+        job_class = Job
+        queue_class = Queue
+        serializer = JSONSerializer
 
         super().__init__(
             queues,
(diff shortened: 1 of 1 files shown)


hsong-rh (approved):

src/aap_eda/core/tasking/__init__.py
```diff
@@ -119,12 +119,9 @@ def __init__(
...
     ):
-        if job_class is None:
-            job_class = Job
-        if queue_class is None:
-            queue_class = Queue
-        if serializer is None:
-            serializer = JSONSerializer
+        job_class = Job
+        queue_class = Queue
+        serializer = JSONSerializer
```
jshimkus-rh, line 124: If these instantiation parameters are non-functional they should be eliminated from the __init__.   Doing so, however, could be an issue if the rq code instantiates using positional, rather than named, args (or does introspection on __init__'s signature).

Still, it should be checked if they can be eliminated.

jshimkus-rh: Comment applies to both specific worker classes.

jshimkus-rh: This implies, for compatibility under the assumption of keyword parameters, that the specific classes would need a **kwargs parameter and that this would need to be checked for job_class, queue_class and serializer and have those removed before invoking super's __init__.

That's not a stellar situation either, but it allows us to be more clear on what our classes do/don't accept at instantiation.

Alex-Izquierdo: You are right. I took the short path. `queues` is positional, rest are named.  Anyway, after further investigation, we don't need it, because django_rq takes our default values from the defined ones in the settings, so I think it is fine as it.

Alex-Izquierdo:
All these params are inyected by django_rq from the settings, so we don't really need this change.

PR #469, opened 2023-10-26
refactor: fix status when activation is completed, disabled or restarted

hsong-rh:


src/aap_eda/services/activation/manager.py
```diff
@@ -507,7 +494,7 @@ def _failed_policy(self):
         ):
             LOGGER.info(
                 f"Activation id: {self.db_instance.id} "
-                f"Restart policy is set to {self.db_instance.restart_policy}. ",
+                f"Restart policy is set to {self.db_instance.restart_policy}. "
                 "Has reached the maximum number of restarts. ",
             )
```
jshimkus-rh, line 499: Adding a definitive statement the activation failed to start is preferable to leaving the user to have to infer it (regardless of how "obvious" such an inference would be).

src/aap_eda/services/activation/manager.py
```diff
@@ -163,25 +163,31 @@ def _check_start_prerequirements(self) -> None:
...
 
         # serializator checks (dependent activation objects)
         is_valid, error = is_activation_valid(self.db_instance)
         if not is_valid:
-            msg = f"Activation {self.db_instance.id} can not be started. "
-            f"Reason: {error}"
+            msg = (
+                f"Activation {self.db_instance.id} can not be started. "
+                f"Reason: {error}"
+            )
```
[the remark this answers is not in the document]
hsong-rh: Agree, these formats mostly are caused by max line length linting.

src/aap_eda/services/activation/manager.py
hsong-rh: Agree

PR #475, opened 2023-11-01
Take over OS env variables in the GitExecutor

robinvalk:
With the current setup the GitExecutor does NOT take over the OS environment variables. Because of this it fails to pull the configured SCM URL as it doesn't take into account the http_proxy settings from in the environment variables.

This PR makes sure the GitExecutor runs with the same environment variables as the main process.

### DIFF
--- a/src/aap_eda/services/project/git.py
+++ b/src/aap_eda/services/project/git.py
@@ -144,9 +144,8 @@ class GitRepository:
 
 class GitExecutor:
     DEFAULT_TIMEOUT: Final = 30
-    ENVIRON: dict = {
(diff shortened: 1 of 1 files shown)


Alex-Izquierdo:
Hello @robinvalk We really appreciate your contribution but the current implementation was done on purpose to keep the control over the environment and prevent potential side effects. In other words, passing through the whole environment to the command is more like using a sledgehammer to crack a nut. 

The real solution to this issue should be support directly an optional http_proxy for projects and inject the respective environment variables or modify the git configuration accordingly. We already have a request feature for this and we will eventually implement it. Would you be interested in implementing this feature yourself?

PR #479, opened 2023-11-02
chore: remove --check flag for isort and black

Alex-Izquierdo:
formatters executed by pre-commit don't need to just advise to fix the issues manually. Let's do it at the same time.


Dostonbek1 (approved):
LGTM. Do we need to include pre-commit hooks changes in this PR as well?

Alex-Izquierdo:
@Dostonbek1 pre-commit runs task.

hsong-rh (approved):

PR #490, opened 2023-11-06
feat: implement system pull policy setting

Alex-Izquierdo:


mkanoor (approved):

src/aap_eda/services/activation/engine/podman.py
```diff
@@ -101,7 +101,9 @@ def start(self, request: ContainerRequest, log_handler: LogHandler) -> str:
             self._set_auth_json_file()
             self._login(request)
             LOGGER.info(f"Image URL is {request.image_url}")
-            self._pull_image(request, log_handler)
+            # TODO: implement IfNotPresent policy
```
mkanoor, line 104: @Alex-Izquierdo 
```
   def _image_exists(self, image_url:str) -> bool:
        for image in self.client.images.list():
               if image_url in image.tags:
                    return True
       return False
    ```

Alex-Izquierdo: Added the rest. Please, review @mkanoor

mkanoor (approved):

PR #492, opened 2023-11-07
[AAP-18000] chore: change validation errors format

hsong-rh:
The second `errors` is changed to `field_errors`. Now the validation error looks like:
```
{
  "errors": {
       "field_errors": "No controller token specified"   
  }
}
```

Resolves [AAP-18000](https://issues.redhat.com/browse/AAP-18000)


Alex-Izquierdo (approved):
LGTM but why are you renaming the key?

hsong-rh:
> LGTM but why are you renaming the key?

We thought `non_field_errors` was too confusing to the end users.

mkanoor (approved):

PR #493, opened 2023-11-07
refactor: improve orchestrator: dosctrings, exception handling

Alex-Izquierdo:
there was a bug and the job is not purged if the task fails. Also improves readability for the queue and adds docstrings

### DIFF
--- a/src/aap_eda/tasks/orchestrator.py
+++ b/src/aap_eda/tasks/orchestrator.py
@@ -14,79 +14,110 @@
 
 import logging
 
+import aap_eda.tasks.activation_request_queue as requests_queue
 from aap_eda.core import models
 from aap_eda.core.enums import ActivationRequest, ActivationStatus
 from aap_eda.core.tasking import unique_enqueue
 from aap_eda.services.activation.manager import ActivationManager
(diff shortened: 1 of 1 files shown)


src/aap_eda/tasks/orchestrator.py
```diff
@@ -14,79 +14,110 @@
...
+            try:
+                if request.request == ActivationRequest.START:
+                    manager.start()
+                elif request.request == ActivationRequest.STOP:
+                    manager.stop()
+                elif request.request == ActivationRequest.RESTART:
+                    manager.restart()
+                elif request.request == ActivationRequest.DELETE:
+                    manager.delete()
+            except Exception:
```
mkanoor, line 56: @Alex-Izquierdo Should we make this more granular?

Alex-Izquierdo: @mkanoor Can you elaborate? 
As it is now and as I understand, we want to pop the request in every error. The manager leaves the activation in the right status and next cycle of the monitor will handle it if is necessary. The only missing piece is to enqueue new delayed actions accordingly to the restart policy

PR #494, opened 2023-11-08
refactor: update activation status when stopping it without the instance

hsong-rh:
Update the status of activation from `Stopping` to `Stopped` when it's disabled without instances.


Alex-Izquierdo (approved):
good catch! @hsong-rh

PR #495, opened 2023-11-08
chore: split ws server in dev env

Alex-Izquierdo:
We are working on issues related with the WS server that requires to run daphne in a separated process to not block api requests.  @mkanoor will do the same for mac.


tools/docker/docker-compose-dev.yaml
```diff
@@ -5,7 +5,7 @@ x-environment:
   - DJANGO_SETTINGS_MODULE=${DJANGO_SETTINGS_MODULE:-aap_eda.settings.development}
   - EDA_ALLOWED_HOSTS=['*']
   - EDA_DEPLOYMENT_TYPE=${EDA_DEPLOYMENT_TYPE:-podman}
-  - EDA_WEBSOCKET_BASE_URL=${EDA_WEBSOCKET_BASE_URL:-ws://eda-api:8000}
+  - EDA_WEBSOCKET_BASE_URL=${EDA_WEBSOCKET_BASE_URL:-ws://eda-ws:8000}
```
mkanoor, line 8: @Alex-Izquierdo should this be 8001

Alex-Izquierdo: No, the process is listening at 8000, 8001 is only for access from the host.

Alex-Izquierdo: here we are using the internal dns "eda-ws" it means it makes the request directly to the host. If the dns would be "host.container.internal" then yes, it should be 8001

Alex-Izquierdo: yeah, that's why I leave on your hands that part ;)

Alex-Izquierdo: Anyway, it is weird, I think podman-machine should support internal resolution, its a very common feature.  Maybe its something that must be enabled manually.

mkanoor (dismissed):

Dostonbek1 (approved):
lgtm

mkanoor (approved):

PR #511, opened 2023-11-14
chore: enable custom port forward for minikube env

Alex-Izquierdo:
Allow to run for example `task minikube:fp:ui -- 9000` this is mostly for convenience and avoid potential collisions.


scripts/eda_kube.sh
```diff
@@ -227,8 +228,8 @@ case ${CMD} in
   "build") build-all "${VERSION}" ;;
   "clean") clean-deployment "${VERSION}";;
   "deploy") deploy "${VERSION}" ;;
-  "port-forward-api") port-forward-api 8000 ;;
-  "port-forward-ui") port-forward-ui 8080 ;;
+  "port-forward-api") port-forward-api ${PORT} ;;
+  "port-forward-ui") port-forward-ui ${PORT} ;;
```
Dostonbek1, line 232: Should this more specific? e.g. `API_PORT` & `UI_PORT`

Dostonbek1 (approved):

scripts/eda_kube.sh
Alex-Izquierdo: you will run only one command at a time.

Dostonbek1: What if you wanted to forward ports for both? Anyway you will have to do it in a new separate terminal but in the end I think explicit > implicit.

Alex-Izquierdo: the whole management of args is not a rebust implementation in this script I think, this is a small change to enable this, that's all. In fact I would like to remove it at some point in favor of ansible for various improvements for the minikube side.

hsong-rh (approved):

PR #512, opened 2023-11-14
refactor: change get_status to return container status object with status and message

hsong-rh:


Alex-Izquierdo (changes requested):

src/aap_eda/services/activation/engine/kubernetes.py
```diff
@@ -133,20 +132,21 @@ def start(self, request: ContainerRequest, log_handler: LogHandler) -> str:
...
         elif container_status.state.terminated:
             exit_code = container_status.state.terminated.exit_code
-            # should change 143 to ActivationStatus.STOPPED?
-            if exit_code in SUCCESSFUL_EXIT_CODES:
+            if exit_code == 0:
                 status = ActivationStatus.COMPLETED
                 LOGGER.info("Pod has successfully exited")
             else:
                 status = ActivationStatus.FAILED
+                message = container_status.state.terminated.message
```
Alex-Izquierdo, line 149: Hi @mkanoor @hsong-rh 
Here is an example of my concerns. 
Here we lose the control over the content of the message potentially returning to the user unclear or complex technical messages. I really tend to refuse to use this in the status_message since this field is exposed to the final user. 
In this case, we should take the exit code and return and well known message like "Container has failed with {exit_code} code" and that message should be placed in the messages.py file and used also by podman to have a consistent format in both engines. The specific message returned from the engine should be logged to allow the user or administrator to get more details for debugging.

src/aap_eda/services/activation/engine/kubernetes.py
```diff
@@ -158,8 +158,9 @@ def get_status(self, container_id: str) -> ActivationStatus:
                 "set status to error.",
             )
             status = ActivationStatus.ERROR
+            message = f"Pod {container_status} is not running or terminated"
```
Alex-Izquierdo, line 161: Same as my previous comments. I think all these messages should be placed in messages.py and shared by both engines for consistency.

Alex-Izquierdo (changes requested):

src/aap_eda/services/activation/engine/kubernetes.py
mkanoor: @Alex-Izquierdo Our software is not meant for business/home users most of our users are going to be Administrators/Operators who would need more information and they would be able to make sense of the message. Plus I think we are expecting podman/k8s to put a meaningful message in there I hope they don't put a stack trace. For now as we are developing it's painful to scroll thru logs to get the error message. For now let us keep it this way and we can bring it up in dev sync and discuss

Alex-Izquierdo: ok, I buy. But lets use a dataclass instead of tuple, please.

Alex-Izquierdo: For all not known messages, we should be consistent and reuse them between engines, like `"Pod {container_status} is not running or terminated"`

src/aap_eda/services/activation/engine/kubernetes.py
Alex-Izquierdo: Also I think it should be "is not running or **is** terminated"

mkanoor:
@hsong-rh In podman isn't there a boolean for OOMFailure?

hsong-rh:
> @hsong-rh In podman isn't there a boolean for OOMFailure?

Yes

src/aap_eda/services/activation/engine/podman.py
```diff
@@ -153,17 +153,22 @@ def start(self, request: ContainerRequest, log_handler: LogHandler) -> str:
...
-    def get_status(self, container_id: str) -> ActivationStatus:
+    def get_status(self, container_id: str) -> tuple[ActivationStatus, str]:
         if self.client.containers.exists(container_id):
             container = self.client.containers.get(container_id)
             if container.status == "exited":
                 exit_code = container.attrs.get("State").get("ExitCode")
+                error = container.attrs.get("State").get("Error", "")
                 if exit_code == 0:
-                    return ActivationStatus.COMPLETED
+                    return ActivationStatus.COMPLETED, ""
```
mkanoor, line 163: @hsong-rh maybe have a success message across both k8s and podman, we could use that constant here

Alex-Izquierdo (approved):
Minor comment, but LGTM

src/aap_eda/services/activation/engine/common.py
```diff
@@ -97,6 +97,11 @@ class ContainerRequest(BaseModel):
     extra_args: tp.Optional[dict] = None
 
 
+class ContainerStatus(BaseModel):
+    status: ActivationStatus
+    message: tp.Optional[str] = ""
```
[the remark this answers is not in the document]
hsong-rh: changed

PR #520, opened 2023-11-16
[AAP-11699] Add update_password command

Dostonbek1:
- Adds a new django-admin command `update_password` to update the given user's password

Solves: [AAP-11699](https://issues.redhat.com/browse/AAP-11699)

### DIFF
--- a/src/aap_eda/core/management/commands/update_password.py
+++ b/src/aap_eda/core/management/commands/update_password.py
@@ -1,0 +1,40 @@
+#  Copyright 2023 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from django.core.management import BaseCommand, CommandError
+from django.db import transaction
+
+from aap_eda.core import models
+
+
+class Command(BaseCommand):
+    help = "Update user password."
+
+    def add_arguments(self, parser):
+        parser.add_argument("-u", "--username", required=True)
+        parser.add_argument("-p", "--password", required=True)
+
+    @transaction.atomic
+    def handle(self, *args, **options):
+        username = options["username"]
+        password = options["password"]
+        try:
+            user = models.User.objects.get(username=username)
+            user.set_password(password)
+            user.save()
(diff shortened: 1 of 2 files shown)


jamesmarshall24 (dismissed):

mkanoor (commented):
@Dostonbek1 can we add tests

src/aap_eda/core/management/commands/update_password.py
```diff
@@ -0,0 +1,40 @@
...
+        parser.add_argument("-p", "--password", required=True)
+
+    @transaction.atomic
+    def handle(self, *args, **options):
+        username = options["username"]
+        password = options["password"]
+        try:
+            user = models.User.objects.get(username=username)
+            user.set_password(password)
+            user.save()
```
Alex-Izquierdo, line 34: I think we should print a message indicating the success of the operation.

src/aap_eda/core/management/commands/update_password.py
```diff
@@ -0,0 +1,40 @@
...
+    help = "Update user password."
+
+    def add_arguments(self, parser):
+        parser.add_argument("-u", "--username", required=True)
+        parser.add_argument("-p", "--password", required=True)
+
+    @transaction.atomic
+    def handle(self, *args, **options):
+        username = options["username"]
+        password = options["password"]
```
Alex-Izquierdo, line 30: We should check if required args were provided and raise a command error if not.

src/aap_eda/core/management/commands/update_password.py
```diff
@@ -0,0 +1,40 @@
...
+        try:
+            user = models.User.objects.get(username=username)
```
Alex-Izquierdo, line 32: I think user model has a check_password method that we could call before.

Alex-Izquierdo (changes requested):
This should have tests

src/aap_eda/core/management/commands/update_password.py
Dostonbek1: We have the message on line 38.

src/aap_eda/core/management/commands/update_password.py
Dostonbek1: `required=True` param in parser takes care of that for us.

src/aap_eda/core/management/commands/update_password.py
Alex-Izquierdo: Right, I commented before read the end. :)

src/aap_eda/core/management/commands/update_password.py
Dostonbek1: Thanks. Working on adding the tests now.

src/aap_eda/core/management/commands/update_password.py
Alex-Izquierdo: good point.

Alex-Izquierdo (approved):

mkanoor (approved):

PR #521, opened 2023-11-17
refactor: enable jinja2 in parsing rulebook

hsong-rh:
Enable Jinja2 to parse port information from the rulebook when it contains a template.

Resolves: [AAP-17870](https://issues.redhat.com/browse/AAP-17870)


tests/integration/services/activation/engine/test_ports.py
```diff
@@ -98,6 +108,63 @@ def test_ports_with_empty_plugin():
...
+---
+- name: Run a webhook listener service
+  hosts: all
+  sources:
+    - ansible.eda.webhook:
+        port: "{{ webhook_port | default(5000) }}"
+"""
+    context = {"webhook_port": 8000}
+    ports = find_ports(rulebook, context)
+    assert ports == [(None, 8000)]
```
mkanoor, line 128: @hsong-rh Is the first element of the tuple a host or port, can we have a test with the filled value

hsong-rh: @mkanoor it's the host, let me add tests for it

PR #525, opened 2023-11-20
chore: fix image of eda-ws pod in docker dev

Alex-Izquierdo:
Follow up: https://github.com/ansible/eda-server/pull/495

The service should use by default the local development image.


Dostonbek1 (approved):
lgtm

jshimkus-rh (approved):

PR #537, opened 2023-11-27
chore: allow arbitrary envvars in minikube

Alex-Izquierdo:
No need to define explicitly the envvars consumed by pods in minikube, since we create a custom config map we can inject it and allow to pass any envvar.


msmagnanijr (approved):
lgtm!

hsong-rh (approved):

PR #539, opened 2023-11-27
refactor: add safe plugins check list

hsong-rh:
https://issues.redhat.com/browse/AAP-17416


src/aap_eda/settings/default.py
```diff
@@ -399,3 +399,7 @@ def _get_secret_key() -> str:
 # ---------------------------------------------------------
 ANSIBLE_RULEBOOK_LOG_LEVEL = settings.get("ANSIBLE_RULEBOOK_LOG_LEVEL", "-v")
 ANSIBLE_RULEBOOK_FLUSH_AFTER = settings.get("ANSIBLE_RULEBOOK_FLUSH_AFTER", 1)
+
+SAFE_PLUGINS_FOR_PORT_FORWARD = settings.get(
+    "SAFE_PLUGINS_FOR_PORT_FORWARD", ["ansible.eda.webhook"]
```
Alex-Izquierdo, line 404: I think the default value should be the one to preserve the backward compatibility. It means support any plugin by default. Otherwise, this would need to be discussed with @josephpisciotta and remark in jira that it needs to update the documentation. @mkanoor @hsong-rh

mkanoor: @Alex-Izquierdo I don't think we support any plugin by default, the only plugin we support is ansible.eda.webhook we can't support ansible.eda.kafka. Because it doesn't work when you test with Kafka running locally. The only way to work around this issue when running Kafka locally is to only allow web hook plugin by default and the rest the user can override. 
```
2023-11-22 21:41:11,986 Container Start Error: 500 Server Error: Internal Server Error (something went wrong with the request: "listen tcp :9092: bind: address already in use\n")
```
This is the error we get when we run Kafka locally

PR #540, opened 2023-11-28
Add websocket connection authentication using jwt token

bzwei:
Use django-restframework-simplejwt for token generation and validation.

Original access token and renew token pair can be generated internally and pass as command line arguments when create an activation instance.

Provide api /auth/token/refresh that takes the renew token to generate renewed access token.
Jira: https://issues.redhat.com/browse/AAP-17774


src/aap_eda/wsapi/middleware.py
```diff
@@ -0,0 +1,68 @@
...
+        return await self.app(scope, receive, send)
+
+
+# maybe move to other location
+def create_jwt_token() -> tuple[str, str]:
+    """Create JWT access and refresh token pair.
+
+    They can be sent to rulebook clients through command line arguments.
+    """
+    user = User.objects.filter(username="admin").first()
```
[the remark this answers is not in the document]
bzwei: No.  They will be different tokens even from the same user and don't interrupt each other.

src/aap_eda/wsapi/middleware.py
```diff
@@ -0,0 +1,80 @@
...
+from rest_framework.request import Request
+from rest_framework.settings import api_settings
+from rest_framework_simplejwt.tokens import RefreshToken
+
+
+@database_sync_to_async
+def _get_authenticated_user(scope: dict) -> Optional[User]:
+    request = HttpRequest()
+    headers = {}
+    for item in scope["headers"]:
```
mkanoor, line 31: @bzwei can we use dict comprehension here to simplify this

mkanoor:
```
request.META = { http_key(k.decode()):v.decode() for (k,v) in scope['headers'] }
```

src/aap_eda/services/auth.py
```diff
@@ -29,3 +34,17 @@ def display_permissions(role_data: dict) -> dict:
...
+
+def create_jwt_token() -> tuple[str, str]:
+    """Create JWT access and refresh token pair.
+
+    They can be sent to rulebook clients through command line arguments.
+    """
+    user, _ = User.objects.get_or_create(
+        username="_service_user",
+        is_service_account=True,
+        defaults={"password": uuid.uuid4()},
```
mkanoor, line 47: @bzwei the python secrets package has functions for this https://docs.python.org/3/library/secrets.html#recipes-and-best-practices

src/aap_eda/wsapi/routes.py
```diff
@@ -8,8 +10,10 @@
...
 )
 
-router = URLRouter(
+wsapi_router = URLRouter(
     [
         path(f"{settings.API_PREFIX}/ws/", wsapi_router),
     ]
 )
+
+router = AllowedHostsOriginValidator(DrfAuthMiddlewareStack(wsapi_router))
```
mkanoor, line 19: @bzwei What would the recommended value for ALLOWED_HOSTS when the Activation container connects into the eda-server? Would that change between podman and Kubernetes?

bzwei: We will need to look into it for prod env. For my test in podman it works without any change. The default value is '*' for dev env.

src/aap_eda/core/models/user.py
```diff
@@ -31,6 +31,7 @@ class User(AbstractUser):
 
     roles = models.ManyToManyField("Role", related_name="users")
     modified_at = models.DateTimeField(auto_now=True, null=False)
+    is_service_account = models.BooleanField(default=False)
```
hsong-rh, line 34: We already have a boolean `is_superuser`, what's the difference between it with this new `is_service_account`? Can we use the existing `is_superuser`?

hsong-rh: For the superuser, its `is_service_account` will also have the default `false` value, right? how it can be excluded?

bzwei: The filter is `(is_service_account=False)`, so superuser and other regular users will be included in the list.

hsong-rh:
@bzwei can you sync the migrations?

Alex-Izquierdo (commented):
Shouldn't we have tests for the new endpoint? Maybe also to ensure that the ws doesn't accept unauthenticated connections?

Alex-Izquierdo (approved):
LGTM I can confirm, rulebook works as expected and connections without token are rejected.
 
@bzwei Maybe a dumb question. Have you tested if the activation works after 2 days (or the value of JWT_REFRESH_TOKEN_LIFETIME_DAYS) as it only supports days, it might be difficult to check .

bzwei:
> LGTM I can confirm, rulebook works as expected and connections without token are rejected.
> 
> @bzwei Maybe a dumb question. Have you tested if the activation works after 2 days (or the value of JWT_REFRESH_TOKEN_LIFETIME_DAYS) as it only supports days, it might be difficult to check .

@Alex-Izquierdo We could make the expiration unit to hours or minutes, but for practical use it will be a very large number. I haven't officially test an activation for days. I tested websocket connection after token expires.

The authentication is checked at connection. Once the connection is established, the websocket communication can continue beyond the token expiration. However since we see the eda-server will periodically drop the connection, the rulebook engine will not be able to reconnect after token expires. The rulebook will terminate at this point. A new activation instance will be required, ideally through our restart policy.

bzwei:
@benthomasson Are you ok with this PR? The ansible-rulebook part has been released in 1.0.4.

Alex-Izquierdo (approved):

Alex-Izquierdo (dismissed):

mkanoor (approved):

benthomasson (approved):

Alex-Izquierdo (approved):

PR #541, opened 2023-11-29
refactor: remove handling error at start in k8s

Alex-Izquierdo:
k8s should not handle exit code != 0 at starting the activation. If the pod has an exit code it means it was executed, so their state should be handled by the monitor, which knows how to apply the restart policy. `_wait_for_pod_to_start` should check only that the pod has been started, but not perform operations of its status after start. 

Otherwise, the manager treat it as error instead of failed:

Screenshot from 2023-11-29 14-44-03

### DIFF
--- a/src/aap_eda/services/activation/engine/kubernetes.py
+++ b/src/aap_eda/services/activation/engine/kubernetes.py
@@ -111,7 +111,7 @@ class Engine(ContainerEngine):
         # f"{request.id}-{uuid.uuid4()}" #noqa: E800
         try:
             log_handler.write("Creating Job")
-            log_handler.write(f"Image URL is {request.image_url}", True)
+            log_handler.write(f"Image URL is {request.image_url}", flush=True)
             self._create_job(request, log_handler)
             LOGGER.info("Waiting for pod to start")
             try:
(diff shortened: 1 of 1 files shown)


src/aap_eda/services/activation/engine/kubernetes.py
```diff
@@ -471,31 +471,10 @@ def _wait_for_pod_to_start(self, log_handler: LogHandler) -> None:
...
-                        reason = statuses[0].state.terminated.reason
-                        if exit_code != 0:
-                            error_msg = (
-                                f"Pod {pod_name} failed with "
-                                f"exit code {exit_code}"
-                            )
-                            LOGGER.error(error_msg)
-                            log_handler.write(error_msg, True)
-                            raise ContainerStartError(error_msg)
+                if pod_phase in ["Failed", "Succeeded", "Running"]:
```
hsong-rh, line 474: If we `break` without any output messages here when the pod fails, the manager cannot know the failure details, right?

hsong-rh: What's the output after you remove the handling of `exit_code != 0`?

Alex-Izquierdo: This the start operation. The manager only need to know if the pod has been started or not. That is the only kind of failure needed to catch, to set the activation as started or error. If there is an exit code it means the pod has been executed. The exit code or error message is handled later by the monitor task who calls get_status. Start method does not report the status of the pod, it is a binary matter: has been started or not. It has a pod_id or the pod has not been started (an exception)

PR #542, opened 2023-11-29
[AAP-15607] - Add pg sslmode verify-ca support to EDA

msmagnanijr:
See: https://issues.redhat.com/browse/AAP-15607


tools/docker/docker-compose-mac.yml
```diff
@@ -18,6 +18,7 @@ x-environment:
   - EDA_ANSIBLE_RULEBOOK_LOG_LEVEL=${EDA_ANSIBLE_RULEBOOK_LOG_LEVEL:-'-v'}
   - EDA_CONTAINER_NAME_PREFIX=${EDA_CONTAINER_NAME_PREFIX:-eda}
   - EDA_PODMAN_ENV_VARS=${EDA_PODMAN_ENV_VARS:-'@none None'}
+  - EDA_PGSSLMODE=${EDA_PGSSLMODE:-disable}
```
mkanoor, line 21: @msmagnanijr if we have to enable this for our local dev enviornment what would we have to change?

msmagnanijr: @mkanoor I realize that I should have included the variables in the 'compose' files and in Kubernetes deployments. I am wondering about the 'development.py' file.  Do you think this covers all the necessary aspects?

src/aap_eda/settings/default.py
```diff
@@ -176,10 +180,15 @@ def _get_secret_key() -> str:
         "USER": settings.get("DB_USER", "postgres"),
         "PASSWORD": settings.get("DB_PASSWORD"),
         "NAME": settings.get("DB_NAME", "eda"),
+        "OPTIONS": {
+            "sslmode": settings.get("EDA_PGSSLMODE", default="allow"),
```
Alex-Izquierdo, line 184: EDA prefix is automatically added by dynaconf. This must be just `PGSSLMODE` same for the rest

Alex-Izquierdo (changes requested):

tools/docker/docker-compose-mac.yml
Alex-Izquierdo: If we merge [this](https://github.com/ansible/eda-server/pull/537), we would not need to add anything for minikube

tools/docker/docker-compose-dev.yaml
```diff
@@ -17,6 +17,7 @@ x-environment:
   - EDA_ANSIBLE_RULEBOOK_LOG_LEVEL=${EDA_ANSIBLE_RULEBOOK_LOG_LEVEL:-'-v'}
   - EDA_CONTAINER_NAME_PREFIX=${EDA_CONTAINER_NAME_PREFIX:-eda}
   - EDA_PODMAN_ENV_VARS=${EDA_PODMAN_ENV_VARS:-'@none None'}
+  - EDA_PGSSLMODE=${EDA_PGSSLMODE:-disable}
```
Alex-Izquierdo, line 20: If the default value is "allow" we don't need to modify anything else and docker-compose would work as it is, don't?

src/aap_eda/settings/default.py
```diff
@@ -176,10 +180,15 @@ def _get_secret_key() -> str:
         "USER": settings.get("DB_USER", "postgres"),
         "PASSWORD": settings.get("DB_PASSWORD"),
         "NAME": settings.get("DB_NAME", "eda"),
+        "OPTIONS": {
+            "sslmode": settings.get("PGSSLMODE", default="allow"),
```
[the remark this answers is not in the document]
msmagnanijr: It makes sense, I will review this point.

tools/docker/docker-compose-dev.yaml
msmagnanijr: @Alex-Izquierdo yep, I believe I forgot to remove this when squashing the commits.

src/aap_eda/settings/default.py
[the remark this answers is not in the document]
msmagnanijr: If we use these two options without providing the certificates when starting the API server, we will encounter something like:  

 `psycopg2.OperationalError: connection to server at "localhost" (::1), port 5432 failed: server does not support SSL, but SSL was required. `

It seems clear enough to me.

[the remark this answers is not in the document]
msmagnanijr: @Alex-Izquierdo maybe it's better to use "prefer" which will fall back to non-TLS if TLS is not supported

msmagnanijr:
@Alex-Izquierdo @mkanoor I updated the PR just to demonstrate what I had in mind in order to enable this in the dev environment. It's not fully functional yet, but please let me know if that makes sense or not.  ( pls ignore the certs )

tools/docker/docker-compose-dev.yaml
```diff
@@ -166,6 +170,9 @@ services:
       - '5432:5432'
     volumes:
       - 'postgres_data:/var/lib/pgsql/data'
+      - './postgres_ssl_config/postgresql-cfg:/opt/app-root/src/postgresql-cfg'
```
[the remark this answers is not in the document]
msmagnanijr: @Alex-Izquierdo yep, it's necessary. I saw that yesterday in the sclorg/postgresql image doc, thanks. Unfortunately, I'm still facing some permission issues. After resolving them, almost everything will be ready.

`waiting for server to start....2023-12-01 12:40:08.573 UTC [35] FATAL:  could not load server certificate file "/opt/app-root/src/certs/fullchain.pem": SSL error code 2147483661`

mkanoor:
@msmagnanijr Should we split this PR into 2 one with just the settings changes and another one with all the docker compose changes for the dev env. Because I am guessing the Mac docker compose is going to be different from the linux docker compose when sharing the certs. Even there I am guessing we would need 1 PR for Mac docker compose 1 for Linux

msmagnanijr:
@mkanoor SSL Mode as "prefer" is ok, right? 
Screenshot from 2023-12-04 09-29-11

Alex-Izquierdo (approved):

msmagnanijr:
PR for dev env: https://github.com/ansible/eda-server/pull/562
PR for UI: https://github.com/ansible/ansible-ui/pull/1331

msmagnanijr:
@ansible/eda-maintainers Hey team, whenever you get a chance, could you take a look at this PR? Thanks

ttuffin (approved):

PR #551, opened 2023-11-30
refactor: rework internal management of activations

Alex-Izquierdo:
Refactor the internal management of activations as has been already discussed
Clients and e2e flows should not be affected. e2e api tests should pass. The only remarkable difference if the new option "max_running_activations" to limit how many activations can run simultaneously. 

High level changes:
- Implement task orchestrator/arbitrator
- Implement activation manager
- Implement container engine and log handler interfaces with concrete implementations for k8s and podman based on the existing code
- Update tests
- Update settings
- Update deps (only jinja for production)
- Some new fields and migrations in models
- Update views to use the new system. 

What doesn't include: 
- Still needs several tests for the new code to be added in later iterations
- Some dead code to be removed in later iterations. 

High level benefits:
- Remove the limitation of one worker per running activation
- Isolated layers and dependency injection 
- No orphan pods and better resilience. 
- Automatic recovering of the state after upgrades or restarts.
- Mitigation of race conditions. 

Jira: https://issues.redhat.com/browse/AAP-17297

Authored by:
@mkanoor @hsong-rh @bzwei @Alex-Izquierdo 
Screenshot 2023-12-01 at 5 18 18 PM


src/aap_eda/services/activation/manager.py
```diff
@@ -0,0 +1,1089 @@
...
+        except IntegrityError as exc:
+            msg = (
+                f"Activation {self.db_instance.id} failed to create "
+                f"activation instance. Reason: {exc}"
+            )
+            self._error_activation(msg)
+            raise exceptions.ActivationStartError(msg) from exc
+
+    def _build_container_request(self) -> ContainerRequest:
+        if self.db_instance.extra_var:
```
mkanoor, line 1037: @Dostonbek1 this is the check we are doing at runtime for the extra_vars once its checked during create we will remove this one

Dostonbek1 (commented):
Activations are stuck on `running` state when running the code locally, although all rules were fired and the logs show the activation was completed. 
Screenshot 2023-12-11 at 8 44 31 AM

Running the same rulebook activation in podman env completes fine and shows `completed` status. Is this expected behavior?

Alex-Izquierdo:
Hi @Dostonbek1 we would need more info, it might be an issue related with the local deployment or its configuration. The logs can give more info as well as details about how are you running all the components of the application.

Dostonbek1 (approved):

mkanoor (approved):

bzwei (approved):

PR #557, opened 2023-12-01
Experimental django-ansible-base integration (might be removed later)

Dostonbek1:
This PR adds experimental code to integrate django-ansible-base with EDA.

_NOTE:_ It might get removed later down the road.

### DIFF
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -234,8 +234,7 @@ REST_FRAMEWORK = {
     "DEFAULT_PAGINATION_CLASS": "aap_eda.api.pagination.DefaultPagination",
     "PAGE_SIZE": 20,
     "DEFAULT_AUTHENTICATION_CLASSES": [
-        # LDAP Integration for 2.4 https://issues.redhat.com/browse/AAP-16938
-        "ansible_base.authentication.session.SessionAuthentication",
+        "aap_eda.api.authentication.SessionAuthentication",
         "rest_framework.authentication.BasicAuthentication",
     ],
     "DEFAULT_PERMISSION_CLASSES": [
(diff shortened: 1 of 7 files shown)


src/aap_eda/settings/default.py
```diff
@@ -230,7 +234,8 @@ def _get_secret_key() -> str:
     "DEFAULT_PAGINATION_CLASS": "aap_eda.api.pagination.DefaultPagination",
     "PAGE_SIZE": 20,
     "DEFAULT_AUTHENTICATION_CLASSES": [
-        "aap_eda.api.authentication.SessionAuthentication",
+        # LDAP Integration for 2.4 https://issues.redhat.com/browse/AAP-16938
+        "ansible_base.authentication.session.SessionAuthentication",
```
mkanoor, line 238: @Dostonbek1 does this resolve the issue that the UI had found https://github.com/ansible/eda-server/blob/0d70bad6e523ceb680c3f23c91ddf77a5dc8b966/src/aap_eda/api/authentication.py#L9 is this pattern consistent with what has been implemented for controller.

Dostonbek1: This doesn't change anything functionally. It is a step towards the main goal of having ansible-base which is to export all common pieces of AWX, Hub and EDA to have more consistent code across all AAP components. We might not even need this change for this LDAP stuff so we could remove it for now to avoid potential issues in the future.

mkanoor: @Dostonbek1 https://github.com/ansible/django-ansible-base/blob/9c92bf26eaa205fd74269c1d7b47369f156b981b/ansible_base/authentication/session.py#L9 seems to be preserving the 401 so this is 👍

hsong-rh (dismissed):
LGTM

bzwei:
LGTM. But haven't really tested it

hsong-rh (approved):

ttuffin (approved):

PR #559, opened 2023-12-05
Fix port conflicts when running other Ansible dev environments

slemrmartin:
Following exposed docker ports conflicts with other AAP services:

- 8000:
  - conflicts with Gateway API
  - configurable by ENV `EDA_API_PORT` (`EDA_API_PORT=8010 go-task docker:up`)
- 5432:
  - conflicts with Gateway Postgres
  - configurable by ENV `EDA_PG_PORT`
- 6379 Redis
  - configurable by ENV `EDA_REDIS_PORT`
- 8001 WS
  - configurable by ENV `EDA_WS_PORT`
- 8443 UI
  - configurable by ENV `EDA_UI_PORT`
- 8888 Podman
  - configurable by ENV `EDA_PODMAN_PORT`

### DIFF
--- a/tools/docker/docker-compose-dev.yaml
+++ b/tools/docker/docker-compose-dev.yaml
@@ -69,7 +69,7 @@ services:
         && scripts/create_superuser.sh
         && aap-eda-manage runserver 0.0.0.0:8000
     ports:
-      - '8000:8000'
+      - '8010:8000'
     depends_on:
       redis:
         condition: service_healthy
@@ -166,7 +166,7 @@ services:
       POSTGRESQL_ADMIN_PASSWORD: secret
       POSTGRESQL_DATABASE: eda
     ports:
-      - '5432:5432'
+      - '5433:5432'
     volumes:
       - 'postgres_data:/var/lib/pgsql/data'
     healthcheck:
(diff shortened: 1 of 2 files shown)


Alex-Izquierdo (commented):
Having different ports between our docker-compose files might be confusing. Also it is a breaking change that might need modifications in other places. 

Why not adding an environment variable with a default value, so new devs can change the port as they want without breaking changes? 

eg. 
```
'${API_PORT_MAP:-8000}:8000'
```

```
'${PG_PORT_MAP:-5432}:5432'
```

In that way everybody can set those envvars to customize the port.

slemrmartin:
that is doable, okay, I'll change it

Alex-Izquierdo (dismissed):

Dostonbek1:
@Alex-Izquierdo Do you think we should also make the same changes in `docker-compose-mac.yml` file as well? And `docker-compose-stage.yml`?

Alex-Izquierdo:
@Dostonbek1 I would say yes for mac because it would be the same scenario than linux, in the case of stage I'm not sure

Dostonbek1 (approved):

Alex-Izquierdo (approved):

slemrmartin:
@Alex-Izquierdo added for mac, not added for stage (the potential port conflict is now dev problem only)

PR #562, opened 2023-12-06
Enable dev env on AAP-15607

msmagnanijr:
See: https://issues.redhat.com/browse/AAP-15607


tools/docker/docker-compose-dev.yaml
```diff
@@ -162,10 +166,14 @@ services:
       POSTGRESQL_PASSWORD: secret
       POSTGRESQL_ADMIN_PASSWORD: secret
       POSTGRESQL_DATABASE: eda
+      POSTGRES_HOST_AUTH_METHOD: trust
     ports:
       - '5432:5432'
     volumes:
       - 'postgres_data:/var/lib/pgsql/data'
+      - './postgres_ssl_config/postgresql-cfg:/opt/app-root/src/postgresql-cfg:Z'
+      - './postgres_ssl_config/certs:/opt/app-root/src/certs:Z'
```
mkanoor, line 175: @msmagnanijr should this be a ENV VAR since the paths are going to be different for each user

tools/docker/docker-compose-tls.yaml
```diff
@@ -0,0 +1,196 @@
...
+      POSTGRESQL_PASSWORD: secret
+      POSTGRESQL_ADMIN_PASSWORD: secret
+      POSTGRESQL_DATABASE: eda
+      POSTGRES_HOST_AUTH_METHOD: trust
+    ports:
+      - '5432:5432'
+    volumes:
+      - 'postgres_data:/var/lib/pgsql/data'
+      - "${EDA_PGSSL_PATHCONFIG:-./postgres_ssl_config/postgresql-cfg}:/opt/app-root/src/postgresql-cfg:Z"
+      - "${EDA_PGSSL_PATHCERT:-./postgres_ssl_config/certs}:/opt/app-root/src/certs:Z"
```
Alex-Izquierdo, line 175: I think this must be a lowercase "z"

Alex-Izquierdo:
Two questions:
- Why a new different docker-compose? it is just a new volume mounts, don't?
- Where are the certs? Why not create a wildcard selfsigned within the repo so everything can work OOB?

tools/docker/docker-compose-dev.yaml
msmagnanijr: @Alex-Izquierdo Exactly. I followed the doc image: https://github.com/sclorg/postgresql-container/tree/62a0e88ae0e5f19c3dad1cb6fea3778dfa2dbf2f/examples/enable-ssl

msmagnanijr:
> Two questions:
> 
> * Why a new different docker-compose? it is just a new volume mounts, don't?

I spoke to @mkanoor  about this, and the idea was not to mix up the settings. There was something else about MacOS that I can't remember now.

> * Where are the certs? Why not create a wildcard selfsigned within the repo so everything can work OOB?

I've compiled a document with a step-by-step guide on generating valid certificates using Red Hat's internal CA, but unfortunately, it still didn't work. If, in the end, we have to resort to using self-signed certificates, I agree to have everything prepared in the repo.

@mkanoor what do you think?

tools/docker/docker-compose-tls.yaml
Alex-Izquierdo: we already have the ":z" in the source code mountpoint without no issues. I suggest to do the same.

Alex-Izquierdo:
I think @mkanoor meant to do this in a separated PR and because the changes for mac will be different than for linux. But I don't see a reason to have a new docker-compose.

Regarding the certs. I think it is not worth to add complexity and/or manual steps to deploy the environment. The goal is to work in a setup with SSL to be closer to production, not to provide a real secure environment. A self-signed cert for the development/stage environment inside the repo is a common approach and completely acceptable to me.

tools/docker/docker-compose-tls.yaml
msmagnanijr: Sure!  let me update this.

mkanoor:
@Alex-Izquierdo We are trying to get certs from our internal CA and try this end to end with SSL for
eda-ui -> eda-api
eda-api -> postgres
worker -> postgres
ansible-rulebook -> Postgres (pg notify/listen)

While we  have certs we have to make sure that we test everything works with it, but there is a manual step to get the certs from our CA which takes about 1 day. Since not all developers would be using this from the get go it would take a while for everyone to get the certs. The docker-compose-dev-ssl.yml is the file which we would eventually keep once everyone has migrated over. Since we have to mount the certs locally from every developers machine to the containers, if we have a single docker compose we would have to make the mount conditional or keep a separate file. Keeping it separate with an ssl suffix will help in development without breaking everyone else's env.

Alex-Izquierdo:
Ok, that makes sense.

msmagnanijr:
@ansible/eda-maintainers  whenever you get a chance, could you please review this? tks

ttuffin:
@msmagnanijr can you resolve the conflict pls?

msmagnanijr:
> @msmagnanijr can you resolve the conflict pls?

Yep, I am working on it.

tools/docker/docker-compose-dev.yaml
```diff
@@ -199,4 +199,4 @@ volumes:
 
 networks:
   service-mesh:
-    name: service-mesh
+    name: service-mesh
```
Alex-Izquierdo, line 202: I think you don't have rebased it properly, this change is already on main. https://github.com/ansible/eda-server/blob/main/tools/docker/docker-compose-dev.yaml#L202

bzwei:
@msmagnanijr I don't have enough expertise to review this. How to use this work? Is it to do docker compose based on [tools/docker/docker-compose-tls.yaml](https://github.com/ansible/eda-server/pull/562/files#diff-bac30884b5533973918133dbe864dd0e9d64b46cb9326749e50d7cb39c9aa015)  then it will automatically enable the ssl connection to DB with all key and cert provided out of box? Otherwise can you document the procedure?

msmagnanijr:
> @msmagnanijr I don't have enough expertise to review this. How to use this work? Is it to do docker compose based on [tools/docker/docker-compose-tls.yaml](https://github.com/ansible/eda-server/pull/562/files#diff-bac30884b5533973918133dbe864dd0e9d64b46cb9326749e50d7cb39c9aa015) then it will automatically enable the ssl connection to DB with all key and cert provided out of box? Otherwise can you document the procedure?

Yes, Mahdu asked me to create this. Let me just give one last review and I'll share the doc.

tools/docker/docker-compose-dev.yaml
Alex-Izquierdo: have you look at this?

Alex-Izquierdo (commented):
Do we need a new docker-compose file? Would make sense to merge it with the existing https://github.com/ansible/eda-server/blob/main/tools/docker/docker-compose-dev-redis-tls.yaml ?

msmagnanijr:
> Do we need a new docker-compose file? Would make sense to merge it with the existing https://github.com/ansible/eda-server/blob/main/tools/docker/docker-compose-dev-redis-tls.yaml ?

I think it makes sense now.  I think we had thought about creating another one for some reason so as not to impact the compose-dev

msmagnanijr:
Closing since I'm dealing with this here:  https://github.com/ansible/eda-server/pull/907

PR #563, opened 2023-12-07
Update django-ansible-base to use latest from devel

Dostonbek1:
We need to update `django-ansible-base` to use latest from devel branch to stay consistent for builds across other AAP components. 

Discussion [ref](https://redhat-internal.slack.com/archives/C01ULNLT6AY/p1701859056092059)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -85,7 +85,7 @@ jobs:
         run: |
             set -ex
             sudo apt update
-            sudo apt install python3-dev libldap2-dev libsasl2-dev libssl-dev
+            sudo apt install python3-dev libldap2-dev libsasl2-dev libssl-dev libxmlsec1 libxmlsec1-dev
 
       - name: Install poetry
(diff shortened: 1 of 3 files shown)


ttuffin (approved):

mkanoor (approved):

.github/workflows/ci.yaml
```diff
@@ -85,7 +85,7 @@ jobs:
         run: |
             set -ex
             sudo apt update
-            sudo apt install python3-dev libldap2-dev libsasl2-dev libssl-dev
+            sudo apt install python3-dev libldap2-dev libsasl2-dev libssl-dev libxmlsec1 libxmlsec1-dev
```
Alex-Izquierdo, line 88: We should have a comment here indicating that libxmlsec1 libxmlsec1-dev are only for the experimental integration. In the future we might skip these things if in the future we want to remove them. 
Also the dev docs should indicate this is necessary. What do you think @Dostonbek1

PR #569, opened 2023-12-12
[AAP-18358] Add key/value pair validation for extra var

Dostonbek1:
* Add validation for extra var on create

Solves: [AAP-18358](https://issues.redhat.com/browse/AAP-18358)

### DIFF
--- a/src/aap_eda/core/validators.py
+++ b/src/aap_eda/core/validators.py
@@ -75,7 +75,9 @@ def is_extra_var_dict(extra_var: str):
     try:
         data = yaml.safe_load(extra_var)
         if not isinstance(data, dict):
-            raise serializers.ValidationError("yaml is not in object format")
+            raise serializers.ValidationError(
+                "Extra var is not in object format"
+            )
     except yaml.YAMLError:
         raise serializers.ValidationError(
             "Extra var must be in JSON or YAML format"
--- a/src/aap_eda/core/validators.py
+++ b/src/aap_eda/core/validators.py
@@ -13,6 +13,7 @@
 #  limitations under the License.
 import logging
 
+import yaml
 from rest_framework import serializers
 
 from aap_eda.core import models
@@ -68,3 +69,16 @@ def check_awx_tokens(user_id: int) -> int:
         )
 
     return user_id
+
+
+def is_extra_var_dict(extra_var: str):
+    try:
+        data = yaml.safe_load(extra_var)
+        if not isinstance(data, dict):
+            raise serializers.ValidationError(
+                "Extra var must be in JSON or YAML format."
+            )
+    except Exception:
+        raise serializers.ValidationError(
+            "Extra var must be in JSON or YAML format."
+        )
--- a/src/aap_eda/core/validators.py
+++ b/src/aap_eda/core/validators.py
@@ -76,7 +76,7 @@ def is_extra_var_dict(extra_var: str):
         data = yaml.safe_load(extra_var)
         if not isinstance(data, dict):
             raise serializers.ValidationError("yaml is not in object format")
-    except Exception:
+    except yaml.YAMLError:
         raise serializers.ValidationError(
             "Extra var must be in JSON or YAML format"
         )
--- a/src/aap_eda/core/validators.py
+++ b/src/aap_eda/core/validators.py
@@ -75,10 +75,8 @@ def is_extra_var_dict(extra_var: str):
     try:
         data = yaml.safe_load(extra_var)
         if not isinstance(data, dict):
-            raise serializers.ValidationError(
-                "Extra var must be in JSON or YAML format."
-            )
+            raise serializers.ValidationError("yaml is not in object format")
     except Exception:
         raise serializers.ValidationError(
-            "Extra var must be in JSON or YAML format."
+            "Extra var must be in JSON or YAML format"
         )
--- a/tests/integration/api/test_extra_var.py
+++ b/tests/integration/api/test_extra_var.py
@@ -68,3 +68,15 @@ def test_extra_var_invalid_data(client: APIClient, extra_var):
     }
     response = client.post(f"{api_url_v1}/extra-vars/", data=invalid_data)
     assert response.status_code == status.HTTP_400_BAD_REQUEST
+
+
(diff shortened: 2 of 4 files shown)


tests/integration/api/test_extra_var.py
```diff
@@ -56,3 +56,15 @@ def test_retrieve_extra_var(client: APIClient):
 def test_retrieve_extra_var_not_exist(client: APIClient):
     response = client.get(f"{api_url_v1}/extra-vars/42/")
     assert response.status_code == status.HTTP_404_NOT_FOUND
+
+
+@pytest.mark.parametrize(
+    "extra_var", ["John", "John, ", "[John, 3,]", '{"name": "John" - 2 }']
```
mkanoor, line 62: @Dostonbek1 Can you try these successful tests with both json and yaml strings.
e.g.
```
import yaml
import json
data = {'a': 1, 'b': 2}
json.dumps(data)
yaml.dump(data)
```

src/aap_eda/core/validators.py
```diff
@@ -68,3 +69,16 @@ def check_awx_tokens(user_id: int) -> int:
...
 
     return user_id
+
+
+def is_extra_var_dict(extra_var: str):
+    try:
+        data = yaml.safe_load(extra_var)
+        if not isinstance(data, dict):
+            raise serializers.ValidationError(
+                "Extra var must be in JSON or YAML format."
```
mkanoor, line 79: @Dostonbek1 I think this should the data should be in object format, with key=value format.
Also we might have to catch Parsing errors, that would be a different error. e.g a data like 2 which is an integer

src/aap_eda/core/validators.py
```diff
@@ -75,10 +75,8 @@ def is_extra_var_dict(extra_var: str):
     try:
         data = yaml.safe_load(extra_var)
         if not isinstance(data, dict):
-            raise serializers.ValidationError(
-                "Extra var must be in JSON or YAML format."
-            )
+            raise serializers.ValidationError("yaml is not in object format")
     except Exception:
```
mkanoor, line 79: @Dostonbek1 Should this Exception be yaml.YAMLError?

tests/integration/api/test_extra_var.py
```diff
@@ -56,3 +56,27 @@ def test_retrieve_extra_var(client: APIClient):
...
+def test_extra_var_invalid_data(client: APIClient, extra_var):
+    invalid_data = {
+        "extra_var": extra_var,
+    }
+    response = client.post(f"{api_url_v1}/extra-vars/", data=invalid_data)
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
+
+
+@pytest.mark.parametrize(
+    "extra_var", ["John: Doe", "John: 2", '{"name": "John"}', '"age": 20']
```
mkanoor, line 74: @Dostonbek1 This doesn't look like a valid Yaml string, is it a list of key value pairs?

Dostonbek1: @mkanoor It is part of how pytest parametrize works - iterates through a list of objects to run the unit test individually.

hsong-rh (approved):
LGTM

tests/integration/api/test_extra_var.py
```diff
@@ -56,3 +56,27 @@ def test_retrieve_extra_var(client: APIClient):
...
+@pytest.mark.parametrize(
+    "extra_var", ["John", "John, ", "[John, 3,]", '{"name": "John" - 2 }']
+)
+@pytest.mark.django_db
+def test_extra_var_invalid_data(client: APIClient, extra_var):
+    invalid_data = {
+        "extra_var": extra_var,
+    }
+    response = client.post(f"{api_url_v1}/extra-vars/", data=invalid_data)
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
```
mkanoor, line 70: @Dostonbek1 Can we test what the error message looks like for validation error, this would be needed by @lgalis to display in the UI.

src/aap_eda/core/validators.py
```diff
@@ -68,3 +69,14 @@ def check_awx_tokens(user_id: int) -> int:
         )
 
     return user_id
+
+
+def is_extra_var_dict(extra_var: str):
+    try:
+        data = yaml.safe_load(extra_var)
```
mkanoor, line 76: @Dostonbek1 How is an empty string handled? Does the UI send in an empty string or None?

Dostonbek1: Serializer has it set to `required=True` so passing an empty string or `null` will return the following error messages
```
{
  "extra_var": [
    "This field may not be blank."
  ]
}
```
and 
```
{
  "extra_var": [
    "This field may not be null."
  ]
}
```

src/aap_eda/core/validators.py
```diff
@@ -68,3 +69,14 @@ def check_awx_tokens(user_id: int) -> int:
...
+        if not isinstance(data, dict):
+            raise serializers.ValidationError("yaml is not in object format")
```
Alex-Izquierdo, line 78: It might be a json too, I think it is more accurate and consistent with the other validation error: 
```suggestion
            raise serializers.ValidationError("Extra var is not in object format")
```

src/aap_eda/core/validators.py
mkanoor: @lgalis Is the UI always sending some data in the extra_var?

lgalis: @Dostonbek1 , @mkanoor - if the Variables field is empty in the Activation form, the /extra_vars POST will not be called.

mkanoor (approved):

Alex-Izquierdo (approved):
LGTM but 
Should we remove here the current validation performed at the manager activation level? Otherwise we should add a TODO comment and create a ticket for it to not lose track of this necessary change.

Dostonbek1:
> LGTM but Should we remove here the current validation performed at the manager activation level? Otherwise we should add a TODO comment and create a ticket for it to not lose track of this necessary change.

@Alex-Izquierdo Removed validation from the manager level.

Alex-Izquierdo (approved):

PR #570, opened 2023-12-12
chore: use centos9 image and add deps for experimental ldap

Alex-Izquierdo:


ttuffin (dismissed):

hsong-rh (approved):

tools/docker/Dockerfile
```diff
@@ -16,7 +16,9 @@ RUN DNF=dnf \
     git-core \
     java-17-openjdk-headless \
     # Experimental LDAP Integration https://issues.redhat.com/browse/AAP-16938
-    openldap-devel" \
+    openldap-devel libxml2-devel xmlsec1-devel xmlsec1-openssl-devel libtool-ltdl-devel" \
```
Dostonbek1, line 19: I don't think `xmlsec1-openssl-devel` is necessary. I don't see django-ansible-base using it.

Alex-Izquierdo: Have you test it without it? Is a requirement to build xmlsec regardless of it is used by django-ansible-base

Alex-Izquierdo: If would prefer to follow python-xmlsec indications unless you are really sure we don't need it.

Dostonbek1: Yes, I did. These three packages are the only ones that were required for successful build of the image: `openldap-devel xmlsec1-devel libtool-ltdl-devel`. I also test ran eda-server instance, didn't see any errors. Having said that, I'm fine with keeping all of those packages in there since it was documented as required in [Pypi](https://pypi.org/project/xmlsec/).

Dostonbek1 (approved):

tools/docker/Dockerfile
Alex-Izquierdo: updated packages accordingly. Please review again @Dostonbek1 @ttuffin @hsong-rh @mkanoor

Dostonbek1 (approved):

ttuffin (approved):

PR #578, opened 2023-12-14
[AAP-19019] Fan-out: Add endpoint to list sources from rulebook

hsong-rh:
Add the endpoint `api/eda/v1/rulebooks/{id}/sources/` to list the sources defined in the rulebook.

Resolves: [AAP-19019](https://issues.redhat.com/browse/AAP-19019)

### DIFF
--- a/src/aap_eda/services/rulebook.py
+++ b/src/aap_eda/services/rulebook.py
@@ -14,6 +14,43 @@
 
 from aap_eda.core import models
 
+DEFAULT_SOURCE_NAME_PREFIX = "EDA_AUTO_SOURCE_"
+
+
+def build_source_list(rulesets: list[dict]) -> list[dict]:
+    results = []
+    if rulesets is None:
+        return results
+
+    i = 1
+    for ruleset in rulesets:
+        duplicate_names = {}
+        srcs = ruleset.get("sources", [])
+        for src in srcs:
+            src_record = {}
+            src_record["name"] = f"{DEFAULT_SOURCE_NAME_PREFIX}{i}"
+            for src_key, src_val in src.items():
+                if src_key == "name":
+                    if src_val in duplicate_names.keys():
+                        duplicate_names[src_val] += 1
+                        src_record[
+                            "name"
+                        ] = f"{src_val} #{duplicate_names[src_val]}"
+                    else:
+                        src_record["name"] = src_val
+
+                if src_key not in ["name", "filters"]:
+                    src_record["type"] = src_key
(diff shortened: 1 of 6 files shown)


src/aap_eda/services/rulebook.py
```diff
@@ -15,6 +15,32 @@
...
+    results = []
+    if rulesets is None:
+        return results
+
+    i = 1
+    for ruleset in rulesets:
+        srcs = ruleset.get("sources", [])
+        for src in srcs:
+            src_record = {}
+            src_record["name"] = f"source_{i}"
```
mkanoor, line 28: @hsong-rh Can we move this source_ to be constant in the class as DEFAULT_SOURCE_NAME_PREFIX?

tests/integration/services/test_rulebook.py
```diff
@@ -0,0 +1,287 @@
...
+- name: Test sample 001
+  hosts: all
+  sources:
+    - name: my source
+      ansible.eda.range:
+        limit: 5
+        delay: 1
+      filters:
+        - noop:
+    - name: my source
```
mkanoor, line 175: @hsong-rh In this scenario we have the same named source which we should not allow but since we have never enforced we might want to just change the name to be my source #2 since its a duplicate.

tests/integration/services/test_rulebook.py
```diff
@@ -0,0 +1,287 @@
...
+
+
+def test_multiple_sources():
+    sources = """
+---
+- name: Test sample 001
+  hosts: all
+  sources:
+    - name: my source
+      ansible.eda.range1:
```
jshimkus-rh, line 112: It'd be nice if the types (`range1` and `range2`) were valid types.
Since they are types they can both be the same.

tests/integration/services/test_rulebook.py
jshimkus-rh: Blue-skying... maybe when generating a name for a source that does not provide one that we generate the name based on the type? For example the name (for otherwise not named) ansible.eda.range would be ansibile.eda.range-1, ansible.eda.range-2, etc. The -<number> would be per-type so each type would counted separately.

Could be seen as redundant w/ the type for those sources.

Alex-Izquierdo:
We can not merge the PoC on main yet, it would be shipped for 1.0.X downstream, otherwise we need first an agreement about this.

hsong-rh:
Changed to DRAFT, wait for further discussions and decisions.

PR #579, opened 2023-12-15
[AAP-19017]Fan-out: add source model

hsong-rh:
Resolves: [AAP-19017](https://issues.redhat.com/browse/AAP-19017)


src/aap_eda/core/models/source.py
```diff
@@ -0,0 +1,50 @@
...
+class Source(models.Model):
+    class Meta:
+        db_table = "core_source"
+        indexes = [
+            models.Index(fields=["uuid"], name="ix_source_uuid"),
+            models.Index(fields=["name"], name="ix_source_name"),
+        ]
+
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4)
+    type = models.TextField(null=False, default="", blank=True)
```
Alex-Izquierdo, line 33: Why this field is not mandatory without defaults. Type can not be empty never.

src/aap_eda/core/models/source.py
```diff
@@ -0,0 +1,50 @@
...
+    name = models.TextField(null=False, unique=True)
+    args = models.TextField(null=True, default=None)
```
Alex-Izquierdo, line 35: Why this is not a json field?

src/aap_eda/core/models/source.py
```diff
@@ -0,0 +1,50 @@
...
+
+class Source(models.Model):
+    class Meta:
+        db_table = "core_source"
+        indexes = [
+            models.Index(fields=["uuid"], name="ix_source_uuid"),
+            models.Index(fields=["name"], name="ix_source_name"),
+        ]
+
+    uuid = models.UUIDField(primary_key=True, default=uuid.uuid4)
```
mkanoor, line 32: @hsong-rh Is the default computed every time or just once?

src/aap_eda/settings/default.py
```diff
@@ -422,3 +422,5 @@ def _get_secret_key() -> str:
     "ansible_base.authentication.backend.AnsibleBaseAuth",
     "django.contrib.auth.backends.ModelBackend",
 ]
+
+EDA_PG_NOTIFY_DSN = "host=localhost dbname=eda user=postgres password=secret"
```
Alex-Izquierdo, line 426: The values should be taken from the DB config. It is a low effort and provides flexibility in future changes and tests.

tests/integration/api/test_source.py
```diff
@@ -0,0 +1,223 @@
...
+    client: APIClient, init_db, check_permission_mock: mock.Mock
+):
+    activation = init_db.activation
+
+    source = models.Source.objects.create(
+        uuid=uuid.uuid4(),
+        name="test-source-1",
+        type="ansible.eda.range",
+        args={"limit": 5, "delay": 1},
+        activation=activation,
```
mkanoor, line 92: @hsong-rh I am thinking this will happen later the user won't be passing in an activation we will create it behind the scenes.

src/aap_eda/core/models/source.py
Alex-Izquierdo: just once, as primary key it is executed at the creation of the record.

tests/integration/api/test_source.py
mkanoor: We can leave it as Null and have the post create task that creates and starts the activation fill it in.

Alex-Izquierdo: Agree

tests/integration/api/test_source.py
```diff
@@ -0,0 +1,223 @@
...
+    assert result["user"] == "test.admin"
+
+    check_permission_mock.assert_called_once_with(
+        mock.ANY, mock.ANY, ResourceType.SOURCE, Action.CREATE
+    )
+
+
+@pytest.fixture
+def init_db():
+    test_rulesets_sample = """
```
mkanoor, line 132: @hsong-rh Is this needed if we are not creating an activation before creating the source. The Activation is post create

tests/integration/api/test_source.py
```diff
@@ -0,0 +1,223 @@
...
+
+@pytest.mark.django_db
+def test_create_source(
+    client: APIClient, init_db, check_permission_mock: mock.Mock
+):
+    data_in = {
+        "name": "test_source",
+        "type": "ansible.eda.generic",
+        "args": {"limit": 1, "delay": 5},
+        "activation_id": init_db.activation.id,
```
mkanoor, line 115: @hsong-rh this could go too

src/aap_eda/core/models/source.py
hsong-rh: As I was told, TextField (yaml format) has better performance in DB.

src/aap_eda/core/models/source.py
```diff
@@ -0,0 +1,50 @@
...
+import uuid
+
+from django.db import models
+
+from aap_eda.core.enums import RestartPolicy
+
+__all__ = "Source"
+
+
+class Source(models.Model):
```
Alex-Izquierdo, line 24: if we are going to replicate the activation model, we need too the is_enabled field.

src/aap_eda/settings/default.py
hsong-rh: we can move this line to the right place later

src/aap_eda/core/models/source.py
Alex-Izquierdo: The serializer operates in the context of a request. We can support text from the view and validate/process it in the serializer, but the type of this field is an object and it's intrinsic to the model. We are not going to treat this internally as text but as object. Doesn't make sense to perform conversions all the time. I think it should be a json field.

mkanoor: @Alex-Izquierdo it was a JSON field but @benthomasson had mentioned that there were performance issues with JSON conversions and storage. And we wanted to follow the extra_var which gets stored as string.

src/aap_eda/core/models/source.py
mkanoor: By default is_enabled=True we don't have a place to start activation for the source, we have to start it right away. It will eventually become Activation eventually.

src/aap_eda/settings/default.py
mkanoor: @Alex-Izquierdo Why would it come from DB CONFIG if they have 2 different databases with different config options why would you want to tie it our DB. One DB could be using SSL other could be using password.

src/aap_eda/core/models/source.py
Alex-Izquierdo: Can we elaborate these performance issues? @benthomasson we are not going to perform queries on the content or partial writes. Either we expect large objects. I don't see any potential issue in our usecase. On the contrary, we are converting a dictionary to yaml only to convert it back again later (the case of listener_args), that is an unnecessary use of cycles.

Alex-Izquierdo: At least we could add a property or methods to work with this field as a dict out of the box.

src/aap_eda/settings/default.py
Alex-Izquierdo: I'm not discussing about the place, the settings seems the right place. I'm talking about not having hardcoded the values, this would not even work in docker-compose where the hostname is different. It must be taken from the variables.

src/aap_eda/core/models/source.py
Alex-Izquierdo: ok.

src/aap_eda/core/models/source.py
jshimkus-rh: Doesn't seem like you can attach a particular serializer (such as the YAML one) to a single field.
If it "must" be yaml, maybe override the appropriate de-/serialization methods and "manually" decode/encode to/from yaml.

src/aap_eda/settings/default.py
Alex-Izquierdo: @mkanoor precisely because we might have a second DB that would be configured just for this, it must be taken from that DB config. I assume that right now we are going to work with the same DB. We can add a second one later. But it shoud be taken from the variables, as it is now not even would work in minikube/docker-compose dev envs because the hostname is different.

Alex-Izquierdo: We can add a new database for this in `DATABASES` (with a fallback into our current default DB) or set it to allow to be configurable by the user:
```
_PG_NOTIFY_DSN = f"{our string composed from DATABASES}"
PG_NOTIFY_DSN = settings.get("PG_NOTIFY_DSN", _PG_NOTIFY_DSN)
```  

but at least for now, as minimal, we need to compose the variable from the current DB settings:
```
PG_NOTIFY_DSN = f"host={DATABASES["default"]["HOST"] etc etc..."
```

Note that we don't need the "EDA" prefix, that would be only for environment variables and is added automatically by dynaconf.

src/aap_eda/api/views/source.py
```diff
@@ -0,0 +1,111 @@
...
+                description="Invalid data to create source."
+            ),
+        },
+    )
+    def create(self, request):
+        context = {"request": request}
+
+        # convert 'args' into yaml format
+        in_args = request.data.get("args", {})
+        request.data["args"] = yaml.dump(in_args)
```
Alex-Izquierdo, line 90: I'm confusing here, what is the type we expect and we want to support? Because it is defined in the model and in the serializator as text and the openapi would indicate that this field must be a string. So we don't need to do here any conversion, just a validation in the model. If we want to be an object or both, we should change the serializer.

src/aap_eda/core/models/source.py
benthomasson: JSONB decomposes the string and indexes it.  We are fine with using the JSON type.

src/aap_eda/api/serializers/source.py
```diff
@@ -0,0 +1,91 @@
...
+from aap_eda.core import models
+
+
+class SourceSerializer(serializers.ModelSerializer):
+    decision_environment_id = serializers.IntegerField(
+        required=False,
+        allow_null=True,
+    )
+
+    user = serializers.HiddenField(default=serializers.CurrentUserDefault())
```
Alex-Izquierdo, line 28: We need also the `args = serializers.CharField()`

src/aap_eda/api/serializers/source.py
```diff
@@ -0,0 +1,91 @@
...
+                    "The 'args' field must be a YAML object (dictionary)"
+                )
+
+            return parsed_args
+        return value
+
+
+class SourceOutSerializer(serializers.ModelSerializer):
+    user = serializers.SerializerMethodField()
+    args = serializers.SerializerMethodField()
```
Alex-Izquierdo, line 65: When we override a field with a MethodField we need as well to define it for the openapi schema. Example: https://github.com/ansible/eda-server/blob/8cecffdd5856f4deff062212afb51e0132dda77f/src/aap_eda/api/serializers/rulebook.py#L253

src/aap_eda/api/views/source.py
mkanoor: is that old code

Alex-Izquierdo: Correct.

src/aap_eda/core/models/source.py
Alex-Izquierdo: updated to json field

src/aap_eda/settings/default.py
Alex-Izquierdo: solved

mkanoor:
@Alex-Izquierdo @lgalis Please review

Alex-Izquierdo (approved):

lgalis (approved):
LGTM

PR #584, opened 2023-12-20
Added sources to activations

mkanoor:
https://issues.redhat.com/browse/AAP-19018

Links Activations to zero or more sources


src/aap_eda/api/serializers/activation.py
```diff
@@ -92,6 +95,11 @@ def to_representation(self, activation):
             activation.ruleset_stats
         )
 
+        sources = []
+        if activation.sources:
+            for source in activation.sources.all():
```
jshimkus-rh, line 100: Is there a specific reason to avoid a list comprehension?  E.g.,

> sources = [
    SourceOutSerializer(source).data for source in activation.sources.all()
]

jshimkus-rh (approved):
Looks good to me.  My question re list comprehension not withstanding.

lgalis (approved):
LGTM

src/aap_eda/api/serializers/activation.py
mkanoor: @jshimkus-rh updated to use list comprehension

src/aap_eda/api/serializers/activation.py
```diff
@@ -130,6 +138,12 @@ class ActivationCreateSerializer(serializers.ModelSerializer):
     )
     user = serializers.HiddenField(default=serializers.CurrentUserDefault())
 
+    sources = serializers.ListField(
```
Alex-Izquierdo, line 141: The openapi spec doesn't reflect the correct schema for the request:
```
                    "sources": {
                        "type": "array",
                        "items": {},
                        "nullable": true
                    }
```

Alex-Izquierdo (changes requested):

src/aap_eda/api/serializers/activation.py
jshimkus-rh: @mkanoor Want to be certain...
If it's not necessary, great; but I assumed the presence of `if activation.sources:` in the previous commit was necessary before using that value.

mkanoor: @jshimkus-rh I checked it locally in my testing the check is not needed since the ManyToMany Manager that sits in between the 2 tables returns an empty array if it doesn't have a relationship

jshimkus-rh (approved):

mkanoor:
@Alex-Izquierdo @lgalis @jshimkus-rh Please review fixed openapi spec

src/aap_eda/api/serializers/activation.py
```diff
@@ -141,6 +152,7 @@ class ActivationCreateSerializer(serializers.ModelSerializer):
     sources = serializers.ListField(
         required=False,
         allow_null=True,
+        child=serializers.IntegerField(),
```
jshimkus-rh, line 155: I presume the `child` is an integer rather than a serialized representation 'cause `ActivationCreateSerializer`  is internal and we only need the `id`?

Alex-Izquierdo (approved):

PR #586, opened 2023-12-22
feat: support multiple awx tokens

Alex-Izquierdo:
Add support for multiple awx tokens and allow to choose the token when creating the activation.
- awx token is an optional field of the activation
- Inspect the content of the rulebook to determine if the token is required
- Populate awx token for existing records in migrations
- Tests accordingly. 
- Disable docstring linting for tests. 
Jiras:
- https://issues.redhat.com/browse/AAP-17327
- https://issues.redhat.com/browse/AAP-13385


src/aap_eda/core/models/rulebook.py
```diff
@@ -36,11 +37,26 @@ class Meta:
...
     # TODO: this field should not have a default value.
+    # TODO: should the content of this field be validated?
+    # https://issues.redhat.com/browse/AAP-19202
     rulesets = models.TextField(null=False, default="")
     project = models.ForeignKey("Project", on_delete=models.CASCADE, null=True)
     created_at = models.DateTimeField(auto_now_add=True, null=False)
     modified_at = models.DateTimeField(auto_now=True, null=False)
 
+    # For instrospection purposes we need to return a dict
+    def get_rulesets_data(self) -> list:
```
mkanoor, line 48: @Alex-Izquierdo Should this be list[dict] instead of list?

src/aap_eda/core/models/rulebook.py
```diff
@@ -36,11 +37,26 @@ class Meta:
     name = models.TextField(null=False)
     description = models.TextField(null=True, default="")
     # TODO: this field should not have a default value.
+    # TODO: should the content of this field be validated?
+    # https://issues.redhat.com/browse/AAP-19202
     rulesets = models.TextField(null=False, default="")
     project = models.ForeignKey("Project", on_delete=models.CASCADE, null=True)
     created_at = models.DateTimeField(auto_now_add=True, null=False)
     modified_at = models.DateTimeField(auto_now=True, null=False)
```
mkanoor, line 46: @Alex-Izquierdo Do we need a attribute in the Rulebook model that has a boolean field that stores token_required.
token_required = models.BooleanField()

Which can be set once we parse the rulebook.

The UI would need this when it detects that a rulebook needs a token, it can prompt the user to pick a token or if the rulebook doesn't need the token it can hide the token selection.

src/aap_eda/core/models/rulebook.py
Alex-Izquierdo: That's correct. Updated.

hsong-rh (approved):
LGTM

tests/integration/api/test_activation.py
```diff
@@ -776,35 +785,109 @@ def test_is_activation_valid():
...
+    fks = create_activation_related_data()
+    test_activation = TEST_ACTIVATION.copy()
+    test_activation["is_enabled"] = True
+    test_activation["decision_environment_id"] = fks["decision_environment_id"]
+    test_activation["rulebook_id"] = new_rulebook_instance_with_job_template.id
+    test_activation["extra_var_id"] = fks["extra_var_id"]
+    test_activation["awx_token_id"] = new_admin_awx_token_instance.id
+
+    response = client.post(f"{api_url_v1}/activations/", data=test_activation)
+    assert response.status_code == status.HTTP_201_CREATED
```
mkanoor, line 893: @Alex-Izquierdo Now that we have a field for the token would we still have non_field_errors in this test file. I think @hsong-rh had added the non_field_errors because token didn't have a column.

src/aap_eda/core/models/rulebook.py
Alex-Izquierdo: I'm not aware of such a requirement. I understood that the check would be reactive. I agree it can improve the UX but this would add more effort in the implementation for the UI and for the API (where we have to change as well the import/sync logic and the migrations). 

Let's discuss it with @eclarizio and @lgalis to be everybody aligned on this or even consider to implement it in later iterations.

tests/integration/api/test_activation.py
hsong-rh: @mkanoor That logic to handle non_field_errors especially had been removed. UI will do the work and display the appropriate messages to the end users.

src/aap_eda/api/serializers/activation.py
```diff
@@ -130,25 +147,26 @@ class ActivationCreateSerializer(serializers.ModelSerializer):
     )
     user = serializers.HiddenField(default=serializers.CurrentUserDefault())
 
+    awx_token_id = serializers.IntegerField(
+        allow_null=True,
+        validators=[validators.check_if_awx_token_exists],
+        required=False,
```
hsong-rh, line 153: @Alex-Izquierdo if the required is `false` and it can be nullable, do we still need validators?

mkanoor:
@Alex-Izquierdo do we need to change this line in the websocket consumer when it tries to [fetch](https://github.com/ansible/eda-server/blob/01746bec37107c64bd0e11a58adb6600015e6951/src/aap_eda/wsapi/consumers.py#L332) and send the token to ansible-rulebook.

Alex-Izquierdo:
> @Alex-Izquierdo do we need to change this line in the websocket consumer when it tries to [fetch](https://github.com/ansible/eda-server/blob/01746bec37107c64bd0e11a58adb6600015e6951/src/aap_eda/wsapi/consumers.py#L332) and send the token to ansible-rulebook.

Good catch @mkanoor websockets updated.

src/aap_eda/wsapi/consumers.py
```diff
@@ -111,14 +112,15 @@ async def handle_workers(self, message: WorkerMessage):
...
-            url=settings.EDA_CONTROLLER_URL,
-            token=await self.get_awx_token(message),
-            ssl_verify=settings.EDA_CONTROLLER_SSL_VERIFY,
-        )
-
         await self.send(text_data=rulebook_message.json())
-        await self.send(text_data=controller_info.json())
+        awx_token = await self.get_awx_token(message)
+        if awx_token:
+            controller_info = ControllerInfo(
```
mkanoor, line 118: @Alex-Izquierdo We would have test the removal of ControllerInfo with ansible-rulebook so it doesn't fail. It has [guards](https://github.com/ansible/ansible-rulebook/blob/c72a42087508e2aa03c0a183cf140dcc2b5aa093/ansible_rulebook/websocket.py#L76) for it but would be nice to know that existing DE's would keep working.

Alex-Izquierdo: @mkanoor In the migration we populate the activations with the existing token, activations without token works fine. I think there are no more cases.

mkanoor (approved):

PR #590, opened 2024-01-04
refactor[fanout] WIP proposal proposal of Abstract base model for activations/sources

Alex-Izquierdo:
Since the source is a new flavor of activation (understanding an activation as a background rulebook process) implement a new BaseActivation abstract model with common logic and two separated models for rulebook activations and source activations. 

All the names are preserved to keep backward compatibility, only the OneToMany relationship with the activation in the source model is removed.


src/aap_eda/core/models/base.py
jshimkus-rh, line 1: The file name is too generic and is easily confused as containing a root class for the entire eda server object hierarchy.
It is is activation specific a name reflecting that should be used.

Alex-Izquierdo: I would agree. Is it fine `base-activation`?

Alex-Izquierdo: Anyway I would like to discuss the final implementation because I'm not sure if an Abstract model is what we need instead of multitable inheritance.

jshimkus-rh: > I would agree. Is it fine `base-activation`?

I'm ok with that.

jshimkus-rh: > Anyway I would like to discuss the final implementation because I'm not sure if an Abstract model is what we need instead of multitable inheritance.

I want to be clear on what you're asking particularly regarding multiple inheritance. The following is my perspective...

I prefer a single-inheritance (subclass descends from one superclass) approach (unless otherwise required) except for the situation where the class will be implementing what other languages call an interface or protocol; i.e., if it acts like a duck, it's a duck.  In those situations the abstraction is the interface/protocol and unrelated classes can be used as long as they adhere to the semantic behavior required of the interface.

Such an approach may be overly general for some situations; i.e., if every subclass in a hierarchy does X, Y and Z in the exact same way with, at most, some minor tweaks but fundamentally dependent on common processing.  Making that a general abstraction may be going further than necessary.  Should the need arise to create an abstraction it would be fairly easy to separate the abstraction out and modify the superclass to inherit from the abstraction while leaving the subclasses alone.

src/aap_eda/core/models/base.py
```diff
@@ -0,0 +1,142 @@
...
+        null=True,
+        default=None,
+        on_delete=models.SET_NULL,
+        related_name="+",
+    )
+
+    class Meta:
+        abstract = True
+
+    def save(self, *args, **kwargs):
```
benthomasson, line 76: Do not override the save function

benthomasson (changes requested):

src/aap_eda/core/models/base.py
benthomasson: Save function is not called on update and doesn't enforce.

Alex-Izquierdo:
Approach discarded

PR #591, opened 2024-01-05
chore: update django-ansible-base dependency installation

Dostonbek1:
Following the PR https://github.com/ansible/django-ansible-base/pull/55, we are now able to install DAB-related packages based on each feature (authentication, filtering, etc.). This PR changes DAB installation to reflect those changes. 

NOTE: If we want to use other features such as `filtering` from DAB in the future, we will need update pyproject.toml to include those features in extras.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -600,22 +600,25 @@ develop = false
 [package.dependencies]
 cryptography = "*"
 Django = ">=4.2.5,<4.3.0"
-django-auth-ldap = "*"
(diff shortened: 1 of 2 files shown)


john-westcott-iv (approved):
Not really knowing poetry syntax I think this looks ok.

Alex-Izquierdo (approved):

hsong-rh (approved):

PR #592, opened 2024-01-08
fix[ws]: handle unsupported paths and message types

Alex-Izquierdo:
We are not handling WS requests for unsupported paths having a 500 unhandled exception
```
eda-ws_1                 | 2024-01-08 13:50:00,821 daphne.server ERROR    Exception inside application: No route found for path 'api/eda/ws/ansible-rulebookk'.
eda-ws_1                 | Traceback (most recent call last):
eda-ws_1                 |   File "/app/venv/lib64/python3.9/site-packages/django/contrib/staticfiles/handlers.py", line 101, in __call__
eda-ws_1                 |     return await self.application(scope, receive, send)
eda-ws_1                 |   File "/app/venv/lib64/python3.9/site-packages/channels/routing.py", line 62, in __call__
eda-ws_1                 |     return await application(scope, receive, send)
eda-ws_1                 |   File "/app/venv/lib64/python3.9/site-packages/channels/routing.py", line 134, in __call__
eda-ws_1                 |     raise ValueError("No route found for path %r." % path)
eda-ws_1                 | ValueError: No route found for path 'api/eda/ws/ansible-rulebookk'.
eda-ws_1                 | 2024-01-08 13:50:00,822 django.channels.server INFO     WebSocket DISCONNECT /api/eda/ws/ansible-rulebookk [10.89.0.20:52444]
```
Also I have found a bug where we are not handling unsupported message types. 
Jira: https://issues.redhat.com/browse/AAP-19294


src/aap_eda/wsapi/consumers.py
```diff
@@ -72,12 +72,31 @@ class Event(Enum):
...
     async def receive(self, text_data=None, bytes_data=None):
         data = json.loads(text_data)
         logger.debug(f"AnsibleRulebookConsumer received: {data}")
 
-        msg_type = MessageType(data.get("type"))
+        try:
+            msg_type = MessageType(data.get("type"))
+        except ValueError:
+            logger.error(f"Unsupported message type: {data}")
+            await self.send('{"error": "unsupported message type"}')
```
jshimkus-rh, line 97: Including the actual type in the response could be useful even though it can be determined from the sent payload.  If we can make the consumer's job easier it's something to consider.

src/aap_eda/wsapi/consumers.py
```diff
@@ -72,12 +72,31 @@ class Event(Enum):
...
+            await self.close()
```
Alex-Izquierdo, line 98: @mkanoor tell me what do you think, I'm not sure if here we want to close the connection.

jshimkus-rh (dismissed):

src/aap_eda/wsapi/consumers.py
Alex-Izquierdo: updated.

src/aap_eda/wsapi/consumers.py
```diff
@@ -90,11 +90,13 @@ async def receive(self, text_data=None, bytes_data=None):
         data = json.loads(text_data)
         logger.debug(f"AnsibleRulebookConsumer received: {data}")
 
+        msg_type = data.get("type")
         try:
             msg_type = MessageType(data.get("type"))
```
jshimkus-rh, line 95:
```suggestion
            msg_type = MessageType(msg_type)
```

jshimkus-rh (dismissed):

jshimkus-rh (dismissed):

src/aap_eda/wsapi/consumers.py
```diff
@@ -72,12 +72,33 @@ class Event(Enum):
...
         logger.debug(f"AnsibleRulebookConsumer received: {data}")
 
-        msg_type = MessageType(data.get("type"))
+        msg_type = data.get("type")
+        try:
+            msg_type = MessageType(msg_type)
+        except ValueError:
+            logger.error(f"Unsupported message type: {data}")
+            payload = {"error": f"unsupported message type: {msg_type}"}
+            await self.send(text_data=json.dumps(payload))
```
mkanoor, line 99: @Alex-Izquierdo What are we sending back?

mkanoor: I think we should just terminate the session and not send anything back

Alex-Izquierdo: updated

src/aap_eda/wsapi/consumers.py
```diff
@@ -72,12 +72,31 @@ class Event(Enum):
...
+class DefaultConsumer(AsyncWebsocketConsumer):
+    """Default consumer for websocket connections.
+
+    This is the consumer to handle all the unexpected paths, it will close the
+    connection with an error message.
+    """
+
+    async def connect(self):
+        await self.accept()
+        await self.send('{"error": "invalid path"}')
```
mkanoor, line 84: @Alex-Izquierdo Should we reveal this, it might be better to just close the connection

src/aap_eda/wsapi/routes.py
```diff
@@ -1,15 +1,28 @@
...
+default_router = URLRouter(
+    [
+        default_path,
+    ],
+)
+
 wsapi_router = URLRouter(
-    [path("ansible-rulebook", consumers.AnsibleRulebookConsumer.as_asgi())]
+    [
+        path("ansible-rulebook", consumers.AnsibleRulebookConsumer.as_asgi()),
```
mkanoor, line 18: @Alex-Izquierdo @bzwei has a PR which will put authentication on this path https://github.com/ansible/eda-server/pull/540

Alex-Izquierdo:
> @Alex-Izquierdo Could HandleRouteNotFoundMiddleware suggested in [django/daphne#165](https://github.com/django/daphne/issues/165) be a more elegant solution?

Hi @bzwei, I agree. Middleware seems like a good approach too, right now I don't fully understand how that middleware works, and even if we copy it, it probably should have its own tests. Since you are more familiar with this, if you really trust on that code, I have no problem using it, otherwise I would think the current approach is also simpler and effortless.

bzwei:
> > @Alex-Izquierdo Could HandleRouteNotFoundMiddleware suggested in [django/daphne#165](https://github.com/django/daphne/issues/165) be a more elegant solution?
> 
> Hi @bzwei, I agree. Middleware seems like a good approach too, right now I don't fully understand how that middleware works, and even if we copy it, it probably should have its own tests. Since you are more familiar with this, if you really trust on that code, I have no problem using it, otherwise I would think the current approach is also simpler and effortless.

If this is not urgently needed, I can try out the middleware solution.

Alex-Izquierdo:
@bzwei Yeah, I don't think it is critical since ansible-rulebook is the only real client of the websockets. Just I noticed that and I thought I could address it easily.

ttuffin:
Closing due to inactivity.

PR #593, opened 2024-01-09
refactor[fanout]: WIP proposal of multitable inheritance for activations/sources

Alex-Izquierdo:
Al alternative approach to https://github.com/ansible/eda-server/pull/590 using multiable inheritance.


src/aap_eda/core/models/source.py
```diff
@@ -16,36 +16,16 @@
...
 
-from aap_eda.core.enums import RestartPolicy
+from aap_eda.core.models.activation import Activation
 
-__all__ = "Source"
+__all__ = ["Source"]
 
 
-class Source(models.Model):
+class Source(Activation):
```
benthomasson, line 24: Sources are not activations.   They are something else.   Maybe both sources and activations are processes.

benthomasson (changes requested):

Alex-Izquierdo:
Approach discarded

PR #594, opened 2024-01-09
[AAP-15757] - [API] Improve our development env for OpenShift

msmagnanijr:
Please see: https://issues.redhat.com/browse/AAP-15757


msmagnanijr:
I still need to handle this in the playbook:  https://github.com/ansible/eda-server/pull/594/commits/66a3007cc31fa86445997340cdd3cc5cc47f2ea3#diff-d7ecd0564dd20df7257c250e200a664fe59b3149a5c9e447cb768059a1589fc6R59

bzwei:
@msmagnanijr In the docs folder can you add step by step instructions how to run this?

msmagnanijr:
> @msmagnanijr In the docs folder can you add step by step instructions how to run this?

Yep!  Please check again. I created a README.

automation/roles/eda-deploy-operator/tasks/main.yaml
```diff
@@ -0,0 +1,35 @@
...
+        name: "{{ eda_namespace }}"
+  when: k8s_info_result is succeeded
+
+- name: Generate EDA Operator kustomization file
+  template:
+    src: kustomization-eda-operator.yaml.j2
+    dest: "{{ eda_operator_files_path }}/kustomization.yaml"
+
+- name: Apply EDA Operator kustomization file
+  shell: |
```
ttuffin, line 33: Since we're already using the `kubernetes.core` collection above, can we make use of that here instead of using `shell`?

automation/group_vars/all/vars.yaml
```diff
@@ -0,0 +1,74 @@
...
+      requests: {}
+
+eda_server_kustomization:
+  eda_database_configuration_host: eda-postgres-13
+  eda_database_configuration_port: 5432
+  eda_database_configuration_database: eda
+  eda_database_configuration_username: eda
+  eda_database_configuration_password: Ansible123!
+  eda_database_configuration_type: managed
+  eda_admin_password_password: Ansible123!
```
ttuffin, line 74: Is the double "password" a typo? Also wouldn't it be better to use the same password (testpass) that we use for our docker deployments? https://github.com/ansible/eda-server/blob/85c43985834b0bdcdbe641005c94bff8e1feaf92/scripts/create_superuser.sh#L23

automation/roles/eda-deploy-operator/tasks/main.yaml
msmagnanijr: @ttuffin it makes total sense. done.

automation/group_vars/all/vars.yaml
msmagnanijr: @ttuffin sure! I will update it.

msmagnanijr:
> @msmagnanijr The `files` folders were missed in both `roles/eda-deploy-operator/` and `roles/eda-deploy-server/`, which caused `ansible-playbook` to fail.

Fixed! thanks. I forgot that it's an empty directory... I'll think about how to deal with that later.

automation/group_vars/all/vars.yaml
msmagnanijr: @ttuffin updated to testpass. About the passwords, eda_database_configuration_password  is used in the secret and the eda_admin_password_password in the configmap.

msmagnanijr:
> I still need to handle this in the playbook: [66a3007#diff-d7ecd0564dd20df7257c250e200a664fe59b3149a5c9e447cb768059a1589fc6R59](https://github.com/ansible/eda-server/commit/66a3007cc31fa86445997340cdd3cc5cc47f2ea3#diff-d7ecd0564dd20df7257c250e200a664fe59b3149a5c9e447cb768059a1589fc6R59)

Fixed!

automation/group_vars/all/vars.yaml
msmagnanijr: @ttuffin hahaha got it! I was typo  :D thanks.

dhaustein:
@msmagnanijr Hi, I have a couple of questions:
1. Is this going to replace the old `task minikube:` commands? (using `minikube` as an example)
2. Could we switch the `ansible_python_interpreter:` to `/usr/bin/env python` so it works with poetry/venv instead of calling the system-wide interpreter?
3. Is the user expected to manage the rest of the lifecycle of the cluster using the tooling of their choice? In other words, if I use `minikube` am I expected then to destroy the cluster once I am done working with it myself instead of running some other playbook?

ttuffin (commented):
I'm noticing that PV claims are getting stuck in pending state in OCP. Im still troubleshooting the cause. I didn't see the issue in minikube.

ttuffin:
The reason it was failing in OCP is because I was using a dev cluster instead of running OCP locally and thus the postgres PVC was failing to create due to missing `crc-csi-hostpath-provisioner`. Since the scope for this PR is to setup a dev environment locally I think we can ignore this error, unless we do want to support deploying on OCP clusters.

automation/group_vars/all/vars.yaml
```diff
@@ -0,0 +1,75 @@
+---
+ansible_python_interpreter: /usr/bin/python3
```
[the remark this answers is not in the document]
msmagnanijr: @ttuffin sure! this makes sense.

automation/ansible.cfg
```diff
@@ -0,0 +1,8 @@
+[defaults]
+host_key_checking = False
+retry_files_enabled = False
+callback_whitelist = profile_tasks
+
+[ssh_connection]
+ssh_args = -o ControlMaster=auto -o ControlPersist=600s
+pipelining = True
```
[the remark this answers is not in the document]
msmagnanijr: @ttuffin good catch,  I'll remove it, this was leftover from another test.

automation/playbook.yaml
```diff
@@ -0,0 +1,8 @@
+- hosts: localhost
+  roles:
+
+    - name: eda-deploy-operator
+      when: (env_type == "minikube" or env_type == "openshift") and eda_deploy_operator == "true"
```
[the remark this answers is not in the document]
msmagnanijr: @ttuffin Yes, I was using this trick when there were still some differences in the roles but now I think it really won't be necessary anymore.

msmagnanijr:
> @msmagnanijr Hi, I have a couple of questions:
> 
> 1. Is this going to replace the old `task minikube:` commands? (using `minikube` as an example)
> 2. Could we switch the `ansible_python_interpreter:` to `/usr/bin/env python` so it works with poetry/venv instead of calling the system-wide interpreter?
> 3. Is the user expected to manage the rest of the lifecycle of the cluster using the tooling of their choice? In other words, if I use `minikube` am I expected then to destroy the cluster once I am done working with it myself instead of running some other playbook?

> 1. Is this going to replace the old `task minikube:` commands? (using `minikube` as an example)
Maybe in the future but not now. I'm guided by the team's ideas. We can implement whatever the team thinks is best :) 

> 2. Could we switch the `ansible_python_interpreter:` to `/usr/bin/env python` so it works with poetry/venv instead of calling the system-wide interpreter?
Yes, that makes perfect sense.

> 3. Is the user expected to manage the rest of the lifecycle of the cluster using the tooling of their choice? In other words, if I use `minikube` am I expected then to destroy the cluster once I am done working with it myself instead of running some other playbook?
Yes, I've been considering automating everything feasible, including this cleanup task. I already have some ideas and plan to incorporate additional roles. I would appreciate your feedback on them

automation/group_vars/all/vars.yaml
```diff
@@ -0,0 +1,71 @@
...
+  operator_literal: eda
+  resource_url: github.com/ansible/eda-server-operator/config/default?ref=0.0.11
+  image_name: quay.io/ansible/eda-server-operator
+  image_new_tag: 0.0.11
+
+eda_operator_files_path: roles/eda-deploy-operator/files
+eda_server_files_path:  roles/eda-deploy-server/files
+
+eda_server:
+  pvc_name: eda-postgres-13-volume
```
rooftopcellist, line 24: Is this being used at all?

It looks to me like a PVC is being created on the fly by the name of `postgres-13-eda-demo-postgres-13-0` using the default storageclass on whatever cluster is used.  

I don't see this `pvc_name` variable referenced anywhere else in this PR.

automation/roles/eda-deploy-server/templates/eda-server-pvc.yaml.j2
```diff
@@ -0,0 +1,12 @@
+apiVersion: v1
```
rooftopcellist, line 1: If my comment above is true, then we can remove this file and the reference to it.

automation/group_vars/all/vars.yaml
```diff
@@ -0,0 +1,71 @@
...
+  access_modes: ReadWriteOnce
+  storage_request: 500Mi
```
rooftopcellist, line 26: If you want to limit the storage request of the dynamically created PVC (when the stateful set is applied), you can set this variable, which would normally be set onthe EDA CR spec.

```
storage_requirements:
  requests:
    storage: 20Gi
```

automation/roles/eda-deploy-server/tasks/main.yaml
```diff
@@ -0,0 +1,80 @@
...
+  set_fact:
+    ingress_type: "{{ 'ingress' if env_type == 'minikube' else 'Route' }}"
+
+- name: Define the TLS termination Mechanism
+  set_fact:
+    tls_termination_mechanism: "{{ '' if env_type == 'minikube' else 'Edge' }}"
+
+- name: Define the PostgreSQL storage class
+  set_fact:
+    postgres_storage_class: "{{ 'standard' if env_type == 'minikube' else 'crc-csi-hostpath-provisioner' }}"
```
rooftopcellist, line 60: Maybe we could have a variable that allows users to override the "smart determination" of the value for postgres_storage_class?  

That way if I want to use this with a cluster that does not have `crc-csi-hostpath-provisioner`, but is also not minikube, I can specify my own storageclass?

Also, have you tried just not specifying storage class at all? iirc, it will use the default storage class for the cluster if none is provided.

msmagnanijr:
@ansible/eda-maintainers  Hey team, whenever you get a chance, could you take a look at this PR? Thanks

automation/group_vars/all/vars.yaml
msmagnanijr: thanks Fixed!

automation/roles/eda-deploy-server/tasks/main.yaml
msmagnanijr: Good idea, I'll do this in the next iteration, I intend to automate the gateway too.

ttuffin:
Closing due to inactivity and the associated JIRA ticket has been closed.

PR #600, opened 2024-01-12
chore: add rulebook id to ruleset serialization

jshimkus-rh:


src/aap_eda/api/serializers/rulebook.py
```diff
@@ -63,7 +63,7 @@ class RulesetSerializer(serializers.ModelSerializer):
     class Meta:
         model = models.Ruleset
         fields = "__all__"
-        read_only_fields = ["id", "created_at", "modified_at"]
+        read_only_fields = ["id", "created_at", "modified_at", "rulebook_id"]
```
mkanoor, line 66: @jshimkus-rh What is the use case for this? We abandoned storing the rulesets in the Database. Now we just store a blob. The old approach was to parse and store everything in the database, but that was abandoned.

jshimkus-rh: @mkanoor The api presents descriptions of rulebooks, rulesets and rules. The rules description includes references to both the rulesets and the rulebooks.  The lack of a reference to a rulebook in the ruleset description is inconsistent with the other representations. No particular use case, but a consistency of representation.

jshimkus-rh: @mkanoor On reflection I believe I misinterpreted your question and this particular line of code mod is not needed.

Alex-Izquierdo: Maybe we should delete rulesets and rules if they are no longer needed.

mkanoor: I feel the same, those classes should be deleted. I don't think we are doing the introspection anymore. Plus the fact that the data might get stale as soon a next Project Sync happens and even if we implement this differencing the rulesets, rules action it is going to be a cumbersome process. For now we are treating a rulebook as a blob.

jshimkus-rh: @mkanoor @Alex-Izquierdo and I were chatting about it.  Were they kept (and do we need to keep them) for backwards compatibility?  Only in that case does it make sense that they were, and possibly to be, kept.
Else, keeping them is confusing noise,  particularly as there doesn't appear to be any mention of why they were kept.

Alex-Izquierdo: I would say they were kept there just because we had other priorities. It sounds to me that @ttuffin asked to removed them too.

ttuffin: https://issues.redhat.com/browse/AAP-16530

PR #607, opened 2024-01-17
feat: honor http_proxy envvar for git command

Alex-Izquierdo:
A quick workaround to support http_proxies at system level until decide/implement a more robust solution for the RFE: https://issues.redhat.com/browse/AAPRFE-835

### DIFF
--- a/src/aap_eda/services/project/git.py
+++ b/src/aap_eda/services/project/git.py
@@ -184,6 +184,8 @@ class GitExecutor:
         if timeout is None:
             timeout = self.DEFAULT_TIMEOUT
 
+        self._set_http_proxy()
+
         try:
             return subprocess.run(
                 [GIT_COMMAND, *args],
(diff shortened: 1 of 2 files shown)


src/aap_eda/services/project/git.py
```diff
@@ -210,3 +212,15 @@ def __call__(
             if e.stderr:
                 usr_msg += f" Error: {e.stderr}"
             raise GitError(usr_msg) from None
+
+    def _set_http_proxy(self) -> None:
```
mkanoor, line 216: @Alex-Izquierdo If these env vars are set in the vm's systemd service file wouldn't they automatically flow to the child process. If we are running these as containers where would the user set these env vars.

Alex-Izquierdo: Hi @mkanoor 
In the case of vm's there are multiple ways to set the envar. In case of containers I'm not sure %100, @rooftopcellist can confirm, but I think the operator allows to add arbitrary environment variables.

benthomasson (changes requested):
This quick fix adds addition burden on the installer to support these environment variables.   We need to determine the whole user experience for setting these variables in VMs and containers before we merge this.

PR #609, opened 2024-01-17
chore: bump minimal python version to 3.11

Alex-Izquierdo:
Make python3.11 the minimal supported version of python, required for incoming features and dependencies. 
Jira: https://issues.redhat.com/browse/AAP-19615

PR #610, opened 2024-01-17
Python311 test

Alex-Izquierdo:
same as https://github.com/ansible/eda-server/pull/609, just trying to figure out why CI don't run.


Alex-Izquierdo:
This was only a test. Closing, the good one is https://github.com/ansible/eda-server/pull/609

PR #612, opened 2024-01-18
feat: add relationships between Credential and Activation 

hsong-rh:
This PR adds:
1. a new credential type: `Vault`
2. Add relationships between Activation and Credentials:
- each activation may have zero or more credentials
- each activation may have a system credential to save the vault password
- system credentials are not displayed to the user

Resolves: [AAP-19418](https://issues.redhat.com/browse/AAP-19418), [AAP-19419](https://issues.redhat.com/browse/AAP-19419)


mkanoor:
@hsong-rh Can we add tests

src/aap_eda/core/models/activation.py
```diff
@@ -102,6 +102,16 @@ class Meta:
         null=True,
         default=None,
     )
+    credentials = models.ManyToManyField(
+        "Credential", related_name="activations", default=None
```
Alex-Izquierdo, line 106: Now that we have a devel branch, we don't need the fanout branch anymore. I feel it would be better if the we merge fanout branch here before, because the credential should have a relation with the source too. Thoughts? @hsong-rh @mkanoor

src/aap_eda/core/models/credential.py
```diff
@@ -38,5 +38,6 @@ class Meta:
     )
     username = models.TextField(null=True)
     secret = EncryptedTextField(null=False)
+    key = models.TextField(null=True)
```
Alex-Izquierdo, line 41: I feel "key" is a little generic. Why not "variable_name" or "extra_var_key"?

Alex-Izquierdo: This value is mandatory depending on the credential_type. I think we should check this constraint for integrity. 
https://docs.djangoproject.com/en/4.2/ref/forms/validation/
https://docs.djangoproject.com/en/4.2/ref/validators/

src/aap_eda/core/enums.py
```diff
@@ -89,6 +89,8 @@ class CredentialType(DjangoStrEnum):
     REGISTRY = "Container Registry"
     GITHUB = "GitHub Personal Access Token"
     GITLAB = "GitLab Personal Access Token"
+    EXTRA_VAR = "Extra Var"
```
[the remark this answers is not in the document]
Alex-Izquierdo: On a second thought the "encrypted" value could suggest that the value must be encrypted so it might be confusing, but I feel we should use the full name, something "extra variable" to be consistent with the rest of the options.

hsong-rh:
> @hsong-rh Can we add tests

This PR only focuses on DB changes. I will add tests later when real logic is added

src/aap_eda/core/enums.py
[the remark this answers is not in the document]
hsong-rh: Like what we did in the demo, a more descriptive title means we need more keyins to create it

src/aap_eda/core/models/credential.py
hsong-rh: > This value is mandatory depending on the credential_type. I think we should check this constraint for integrity. https://docs.djangoproject.com/en/4.2/ref/forms/validation/ https://docs.djangoproject.com/en/4.2/ref/validators/

Yes, the validator will be added in the next PR

src/aap_eda/core/enums.py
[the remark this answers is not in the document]
mkanoor: Since the UI is going to be displaying it the text won't matter like we did in swagger.
The Extra Var terminology is familiar to ansible users so we need that.Maybe Vaulted Extra Var

src/aap_eda/core/models/activation.py
hsong-rh: can we focus on vault support in this devel branch now and merge fanout later?

Alex-Izquierdo: Fanout branch was created to not create conflicts in main same as devel. The devel branch is for all the changes for the next release not only for vault. Unless we decide that fanout is not ready to be merged as it is, I think we should merge them. All the work for the next release should start from and be merged on the same branch.

mkanoor: Is the fanout modeling going to change do we have to wait for it to be merged before we start the vault work. if the fanout model is still in a state of flux we should let this one be separate.

src/aap_eda/core/models/credential.py
mkanoor: We have a credential model which supports different types and each type has a different set of required attribute names, are we going to share some of the attribute names across these different types. Otherwise this table will have the attributes for all types in a single table

mkanoor:
@hsong-rh For the Ansible Vault Password we should change it to Vault to align with this
image (2)

src/aap_eda/api/serializers/credential.py
```diff
@@ -30,20 +32,55 @@ class Meta:
...
+    def validate(self, data):
+        key_required_types = [
+            CredentialType.EXTRA_VAR,
+            CredentialType.VAULT_PASSWORD,
+        ]
+        credential_type = data.get("credential_type", CredentialType.REGISTRY)
+        # TODO: may need to change `key` to `variable_name` later
+        key = data.get("key")
+
+        if credential_type in key_required_types and key is None:
```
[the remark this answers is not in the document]
mkanoor: Multiple classes are getting stored in Credential Table, if you take every attribute from every class and jam it into 1 table imagine the columns in that class so there is some reuse happening in that table.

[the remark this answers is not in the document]
hsong-rh: For `vault password` type, `key` is used to store vault label; for `extra var` type, it's used to store variable name.

src/aap_eda/core/models/activation.py
```diff
@@ -102,6 +102,16 @@ class Meta:
...
+    )
+    system_credential = models.OneToOneField(
```
mkanoor, line 108: @hsong-rh @Alex-Izquierdo We would need this field to encrypt the EDA_PG_NOTIFY_DSN for the pg_notify action. We cannot depend on the user creating a Vault password for this we would have to add one for the activation behind the scenes. The Activations that are using the "sources" we would be adding the EDA_PG_NOTIFY_DSN as a vaulted string.

hsong-rh:
@mkanoor @Alex-Izquierdo The `EXTRA_VAR` is removed from Credential types

src/aap_eda/core/enums.py
```diff
@@ -89,6 +89,7 @@ class CredentialType(DjangoStrEnum):
     REGISTRY = "Container Registry"
     GITHUB = "GitHub Personal Access Token"
     GITLAB = "GitLab Personal Access Token"
```
mkanoor, line 91: @hsong-rh Should this be just Vault

tests/integration/api/test_credential.py
```diff
@@ -50,12 +52,31 @@ def test_create_credential(client: APIClient):
...
         "credential_type": CredentialType.REGISTRY,
         "id": id_,
+        "identifier": None,
     }
     obj = models.Credential.objects.filter(pk=id_).first()
     assert obj.username == "me"
     assert obj.secret == "mypassword"
 
 
+@pytest.mark.django_db
```
mkanoor, line 62: @hsong-rh Can we please add a test for Vault without an identifier since the identifier can be empty.
Screenshot 2024-01-23 at 4 55 02 PM

src/aap_eda/core/enums.py
hsong-rh: ok

tests/integration/api/test_credential.py
mkanoor: @hsong-rh Can the credential be updated after its created, initially its created without the identifier but then the user decides to add an identifier during an update

hsong-rh: @mkanoor added

src/aap_eda/core/models/credential.py
hsong-rh: "key" is changed to "identifier"

src/aap_eda/api/views/activation.py
```diff
@@ -88,6 +95,20 @@ def create(self, request):
             status=status.HTTP_201_CREATED,
         )
 
+    def perform_create(self, serializer):
+        instance = serializer.save()
+
+        if instance.credentials:
```
mkanoor, line 101: @hsong-rh Can this be made into a function since we would be calling it only if we have some internal encrypted fields to use. E.g in fanout we have an extra var called EDA_PG_NOTIFY_DSN which has a secret embedded in it in that scenario we will want to create a system credential and then encrypt the EDA_PG_NOTIFY_DSN

hsong-rh: @mkanoor Based on the discussion, this logic is removed from here and will be added to the fanout project.

mkanoor (dismissed):

mkanoor (dismissed):

src/aap_eda/core/models/credential.py
```diff
@@ -38,5 +38,6 @@ class Meta:
     )
     username = models.TextField(null=True)
     secret = EncryptedTextField(null=False)
+    identifier = models.TextField(null=True)
```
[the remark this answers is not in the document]
hsong-rh: Changed to `vault_identifier`

mkanoor (dismissed):

src/aap_eda/api/views/credential.py
```diff
@@ -127,6 +120,27 @@ def get_serializer_class(self):
...
+            status.HTTP_200_OK: OpenApiResponse(
+                serializers.CredentialSerializer(many=True),
+                description="Return a list of credential.",
+            ),
+        },
+    )
+    def list(self, request):
+        credentials = models.Credential.objects.exclude(
+            credential_type=CredentialType.VAULT,
+            vault_identifier=EDA_SERVER_VAULT_LABEL,
```
Alex-Izquierdo, line 135: - It seems that we introduce a new "internal credential" type identified by `credential.credential_type == CredentialType.VAULT and credential.vault_identifier == EDA_SERVER_VAULT_LABEL` We already have a similar logic for destroy and I wonder if it would be more consistent and reusable through methods in the model. A method that returns a reusable queryset for non-internal credentials and a `is_internal` method that tell us if a credential is internal. 

- If we are not going to allow a deletion of a internal credential and we are not going to return them in the list view, does make sense to allow to fetch it by GET? It looks like the GET verb should have a similar logic to prevent returning an internal credential, don't?

- The previous two points make me wonder if it would not be better to have this internal credential in a separated table instead of taking care of excluding it in multiple places. @hsong-rh @benthomasson @mkanoor

Alex-Izquierdo (commented):
I have suggested to move some logic to the model and to consider a different table for the internal credential. Otherwise LGTM

src/aap_eda/api/views/credential.py
hsong-rh: @Alex-Izquierdo what's your concern? I feel the current way is straightforward.

Alex-Izquierdo (approved):

src/aap_eda/api/views/credential.py
mkanoor: @Alex-Izquierdo We can fix the GET but for the rest of the items can we punt now and do it later when the Gateway Credentials show up, we need the system credential for FanOut.

Alex-Izquierdo: approved.

hsong-rh: @mkanoor @Alex-Izquierdo The GET for internal credentials is hidden.

mkanoor (approved):

Alex-Izquierdo (approved):

PR #613, opened 2024-01-18
fix: add dynamic settings from django-ansible-base & revert to use latest devel branch

Dostonbek1:
* define ansible base features to be used
* import ansible base dynamic settings
* revert to using latest from devel branch of django-ansible-base

### DIFF
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -69,8 +69,11 @@ PODMAN_MOUNTS - A list of dicts with mount options. Each dict must contain
                              "type": "bind"}]'
 
 """
+import os
+
 import dynaconf
 from django.core.exceptions import ImproperlyConfigured
+from split_settings.tools import include
 
 default_settings_file = "/etc/eda/settings.yaml"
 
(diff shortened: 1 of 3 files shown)


src/aap_eda/settings/default.py
```diff
@@ -434,3 +437,15 @@ def _get_secret_key() -> str:
     "ansible_base.authentication.backend.AnsibleBaseAuth",
     "django.contrib.auth.backends.ModelBackend",
 ]
+ANSIBLE_BASE_FEATURES = {
+    "AUTHENTICATION": True,
+    "FILTERING": False,
+    "SWAGGER": False,
+}
```
john-westcott-iv, line 444: FYI, this will change with the split.

Dostonbek1: I initially wanted to include the split changes in this PR but I thought I would do in another PR after since this is more urgent to unblock QE pipelines.

jshimkus-rh (approved):

bzwei (approved):

PR #617, opened 2024-01-19
feat: enable sources and activations for orchestrator/manager

Alex-Izquierdo:
This is the first PR to enable sources and activations for the orchestrator and the activation manager. Implements the concept of the "process parent" as union[source, activation] for all the common logic, a double relation with sources and activations in the activation instance (future "rulebook process") and a proxy pattern to adapt the existing machinery to work with both models. 

Includes:
- Replication of the source model as an activation
- StatusHandlerModelMixin and ProcessParentValitatorMixin for code reuse
- ProcessParentProxy for common operations for both sources and activations
- Migrations
- Changes in the orchestrator and manager to work with the new ProcessParentProxy

NOTES:
- many things (names and logs) are not renamed yet. This is done on purpose to not have a big PR and create unnecessary noise in the review. At this point the names are secondary and can be addressed later, this PR is focused on the functionality. 
- The PR should work fine as it is. E2E tests pass. This PR is pending yet to update failing unit tests.

Next iterations/PR's:
- run sources
- finish the renaming. 

Jira: https://issues.redhat.com/browse/AAP-19384


src/aap_eda/tasks/orchestrator.py
```diff
@@ -13,78 +13,85 @@
...
-            f"Processing monitor request for activation {activation_id}",
+            f"Processing monitor request for parent {process_parent.id}",
         )
-        ActivationManager(activation).monitor()
+        ActivationManager(process_parent).monitor()
 
 
 def _run_request(
-    activation: models.Activation,
+    parent_process: ProcessParentProxy,
```
Alex-Izquierdo, line 77: Rename variable to process_parent

src/aap_eda/tasks/orchestrator.py
```diff
@@ -99,52 +106,57 @@ def _run_request(
...
             f"Failed to process request {request.request} for "
-            f"activation {activation.id}. Reason {str(e)}",
+            f"activation {parent_process.id}. Reason {str(e)}",
         )
     return True
 
 
-def _can_start_new_activation(activation: models.Activation) -> bool:
-    num_running_activations = models.Activation.objects.filter(
+def _can_start_new_activation(parent_process: ProcessParentProxy) -> bool:
```
Alex-Izquierdo, line 114: rename variable to process_parent

src/aap_eda/tasks/orchestrator.py
```diff
@@ -99,52 +106,57 @@ def _run_request(
...
+        _manage,
+        process_parent_data=process_parent.to_dict(),
+    )
 
 
-def start_activation(activation_id: int) -> None:
-    """Create a request to start the activation with the given id."""
-    _make_user_request(activation_id, ActivationRequest.START)
+def start_activation(process_parent: ProcessParentProxy) -> None:
+    """Create a request to start the process parent with the given id."""
```
Alex-Izquierdo, line 143: remove "with the given id."

src/aap_eda/tasks/orchestrator.py
```diff
@@ -157,13 +169,29 @@ def monitor_activations() -> None:
...
     # monitor running instances
-    for activation in models.Activation.objects.filter(
+    for instance in ProcessParentProxy.filter(
         status=ActivationStatus.RUNNING,
     ):
-        job_id = _manage_activation_job_id(activation.id)
-        unique_enqueue("activation", job_id, _manage, activation.id)
+        process_parent = ProcessParentProxy(instance)
+        job_id = _manage_activation_job_id(process_parent)
+        # Rename the activation queue to "process_parent"
```
Alex-Izquierdo, line 187: Remove this duplicated comment as it is already in the queue settings.

src/aap_eda/core/models/activation.py
```diff
@@ -98,153 +88,88 @@ class Meta:
...
         default=ActivationStatus.PENDING,
     )
     git_hash = models.TextField(null=False, default="")
-    activation = models.ForeignKey("Activation", on_delete=models.CASCADE)
+
+    # Source and activation will be completely different models
+    # Since for now are only two, two foreign keys are enough
+    # If more are added, a generic relation might be considered
+    # In this way the relation is clear and simple without the
+    # tradeoffs of a generic relation
```
Alex-Izquierdo, line 128: Add a link to the docs of django generic relations for reference.

src/aap_eda/core/models/proxies.py
```diff
@@ -0,0 +1,110 @@
...
+        """Return a ProcessParentProxy from a dictionary representation.
+
+        Used for deserialization for redis rq tasks.
+        """
+        if data["is_activation"]:
+            return ProcessParentProxy(Activation.objects.get(id=data["id"]))
+        return ProcessParentProxy(Source.objects.get(id=data["id"]))
+
+    @staticmethod
+    def get_all() -> CombinedQuerySet:
```
Alex-Izquierdo, line 97: double check if this method is used or needed.

src/aap_eda/core/models/activation.py
```diff
@@ -98,153 +88,88 @@ class Meta:
...
+        # populate latest_instance when creating
+        if self._state.adding:
+            parent = getattr(self, "activation", None) or getattr(
+                self,
+                "source",
+                None,
+            )
+            if not parent:
+                raise ValueError(
+                    "ActivationInstance must have either activation or source"
```
Alex-Izquierdo, line 167: Change this. Parent can be a property and the integrity is ensured by ProcessParentValitatorMixin

src/aap_eda/core/models/queue.py
```diff
@@ -12,20 +12,45 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+
```
Alex-Izquierdo, line 15: remove empty line

src/aap_eda/core/models/mixins.py
```diff
@@ -0,0 +1,129 @@
...
+            raise UnknownStatusError(f"Status [{self.status}] is invalid")
+
+    def _is_valid_status(self):
+        try:
+            ActivationStatus(self.status)
+        except ValueError as error:
+            raise UnknownStatusError(error)
+
+
+class ProcessParentValitatorMixin:
```
mkanoor, line 118: @Alex-Izquierdo Typo Validator

src/aap_eda/services/activation/manager.py
```diff
@@ -102,11 +102,12 @@ def __init__(
...
             container_engine: The container engine to use.
         """
-        self.db_instance = db_instance
+        self.db_instance = proxy_instance.instance
+        self.proxy_instance = proxy_instance
         if container_engine:
             self.container_engine = container_engine
         else:
-            self.container_engine = new_container_engine(db_instance.id)
+            self.container_engine = new_container_engine(proxy_instance.id)
```
mkanoor, line 110: @Alex-Izquierdo Will this id be unique across both the Activation and Source class. Otherwise the engine will create labels that might match multiple objects

Alex-Izquierdo: @mkanoor My first approach is to keep them separate to respect the requirement of treating them as completely different objects, which allows for the possibility of adding more types in the future or implementing the relationships between the child objects through generic relations. In this way, the manager needs to be updated to create the activation instance according to the type of the ProcessParent. This will be addressed in the next PR to implement the management and execution of sources.

I think implementing a unique id among sources and activations is possible in different ways, but at this point, I feel it's not a good idea. Anyway, I don't have enough arguments in favor or against, so we can discuss it in more detail if you want.

src/aap_eda/core/models/mixins.py
Alex-Izquierdo: thanks, updated.

src/aap_eda/services/activation/manager.py
mkanoor: @Alex-Izquierdo I am talking about stuff like this
https://github.com/ansible/eda-server/blob/01746bec37107c64bd0e11a58adb6600015e6951/src/aap_eda/services/activation/engine/kubernetes.py#L95
If the 2 ids are the same this code is going to have issues.

Alex-Izquierdo: Yes, I'm aware, we need to change this to deal with sources and activations, is part of the scope for the next PR. We were using the string "activation" to name our resources, now we will have to use activation or source. 
I thought the container request might contain this information.

Alex-Izquierdo: I have not looked at that in deep but I think we are passing the activation_id to the engine only for this type of thing. I want to look if we really need to receive the id there.

src/aap_eda/core/models/activation.py
```diff
@@ -98,152 +90,92 @@ class Meta:
...
-        if self.status == ActivationStatus.PENDING and not self.is_enabled:
-            self.status_message = "Activation is marked as disabled"
-
-    def _get_default_status_message(self):
-        try:
-            return ACTIVATION_STATUS_MESSAGE_MAP[self.status]
-        except KeyError:
-            raise UnknownStatusError(f"Status [{self.status}] is invalid")
-
-    def _is_valid_status(self):
```
mkanoor, line 156: @Alex-Izquierdo Where did this move to?

Alex-Izquierdo: In the mixin: https://github.com/ansible/eda-server/pull/617/files#diff-23367b3fab1d54eb173437e73363840615b77ca82575d88ede9a996520d4cc67R111

src/aap_eda/core/models/proxies.py
```diff
@@ -0,0 +1,106 @@
...
+        Used for deserialization for redis rq tasks.
+        """
+        if data["is_activation"]:
+            return ProcessParentProxy(Activation.objects.get(id=data["id"]))
+        return ProcessParentProxy(Source.objects.get(id=data["id"]))
+
+    @staticmethod
+    def filter(**kwargs) -> CombinedQuerySet:  # noqa: A003
+        """Filter Source and Activation objects."""
+        return CombinedQuerySet(
```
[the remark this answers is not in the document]
Alex-Izquierdo: I could not say right now. I think when the necessity comes up with specific usecases we will see if we can adapt our current code, change it or adapt ansible-base, it currently has lines like this one: https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/rest_filters/utils.py#L26

mkanoor (approved):

PR #618, opened 2024-01-22
ci: set GH token when running scheduled tests

ttuffin:
Set the EDAQA_GH_TOKEN environment variable when running the scheduled eda-qa test suite. Fixes potential failures due to rate limiting.


ttuffin:
We'll do this in the test suite instead by using encrypted vars: https://github.com/ansible/eda-qa/pull/288

PR #619, opened 2024-01-23
fix: [AAP-19694] Fix pod missing logs intermittently

hsong-rh:
When we use [Podman](https://github.com/containers/podman-py/blob/main/podman/domain/containers.py#L262)/[Kubernetes](https://github.com/kubernetes-client/python/blob/ae7b5ddd219fe09b6ed0be715dcca3377a029584/kubernetes/docs/CoreV1Api.md#read_namespaced_pod_log) APIs to update logs based on the timestamp (`since` for Podman, `since_seconds` for Kubernetes), it only accepts the number of seconds. It will cause log entries to be missing intermittently if logging to DB stops in the middle of the second.

This PR adds a `log_timestamp` field to activation_instance_log. It will store the log timestamps and use them to tell whether the line is duplicated.

Example of timestamp returned by Podman API, it doesn't contains macro seconds:
```
2024-01-29T13:49:38-05:00 2024-01-29 18:49:38,240 rq.worker INFO     activation: aap_eda.tasks.orchestrator._manage(16) (activation-16)
2024-01-29T13:49:38-05:00 2024-01-29 18:49:38,296 aap_eda.tasks.orchestrator WARNING  Activation 16 no longer exists, activation manager task will not be processed
2024-01-29T13:49:38-05:00 2024-01-29 18:49:38,301 rq.worker INFO     activation: Job OK (activation-16)
```
But in Kubernetes, the returned timestamp contains macro seconds:
```
2024-01-29T20:39:08.144053185Z 2024-01-29 20:39:08,143 aap_eda.core.tasking INFO     Enqueing unique job: monitor_project_tasks
2024-01-29T20:39:08.146846976Z 2024-01-29 20:39:08,146 rq.worker INFO     default: Job OK (5f4b2fe6-0c5c-482c-b346-b97a9c3dad13)
2024-01-29T20:39:08.146918017Z 2024-01-29 20:39:08,146 rq.worker INFO     Result will never expire, clean up result key manually
```
To avoid the following `since_time` issue, the real `since_seconds` number is always plus one when it's sent to the API call to make sure no log entries are missing, and the log entries written at log_read_at inside DB will be deleted to make sure no duplicates.

Resolves: [AAP-19694](https://issues.redhat.com/browse/AAP-19694)
Resolves: [AAP-19845](https://issues.redhat.com/browse/AAP-19845)

In Podman they support **since** which is the number of seconds since epoch and works even when the clocks between the 2 processes are not in sync.

Problems in Kubernetes API https://github.com/kubernetes/kubernetes/issues/108498
Since we use **since_seconds** and not **since_time** which is documented but not available in the openapi clients we loose information if the clocks are not in sync.


mkanoor:
@hsong-rh does this account for the line number, when we delete the old lines the line number will have to be reset

hsong-rh:
> @hsong-rh does this account for the line number, when we delete the old lines the line number will have to be reset

No, this PR only focuses on missing logs. I have asked @ttuffin to open another ticket for line number issue.

src/aap_eda/services/activation/db_log_handler.py
```diff
@@ -47,6 +47,7 @@ def write(
         lines: Union[list[str], str],
         flush: bool = False,
         timestamp: bool = True,
+        log_timestamp: int = 0,
```
Alex-Izquierdo, line 50: This field is not used and doesn't look like we need to keep the state of this, don't?

hsong-rh: The Podman and Kubernetes engines need it when inserting log lines into DB.

Alex-Izquierdo: where is used? 
I don't see any reference to LogHandler.log_timestamp, always is passed through function arguments which makes sense to me.

Alex-Izquierdo: nevermind, the GH view has confused to me and I thought it is in the __init__.

Alex-Izquierdo, line 50: Is correct that default value? I think we expect always a value.

hsong-rh: 0 should be ok. It only happens in the beginning when the lines are added by us.

Alex-Izquierdo (approved):

bzwei (approved):

PR #622, opened 2024-01-24
[AAP-19306] Add DAB JWT settings

Dostonbek1:
This PR adds changes to enable JWT authentication in EDA from django-ansible-base.

### DIFF
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -252,6 +252,7 @@ REST_FRAMEWORK = {
     "DEFAULT_AUTHENTICATION_CLASSES": [
         "aap_eda.api.authentication.SessionAuthentication",
         "rest_framework.authentication.BasicAuthentication",
+        "ansible_base.jwt_consumer.eda.auth.EDAJWTAuthentication",
     ],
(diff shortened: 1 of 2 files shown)


src/aap_eda/settings/default.py
```diff
@@ -445,3 +446,13 @@ def _get_secret_key() -> str:
     os.path.dirname(dynamic_config.__file__), "dynamic_settings.py"
 )
 include(dab_settings)
+
+# ---------------------------------------------------------
+# DJANGO ANSIBLE BASE JWT SETTINGS
+# ---------------------------------------------------------
+ANSIBLE_BASE_JWT_VALIDATE_CERT = settings.get(
```
mkanoor, line 453: @Dostonbek1 Is the JWT support in DAB going to support getting refresh tokens. EDA is using Simple JWT for authenticating web sockets connection and getting refresh tokens.

bzwei: @Dostonbek1 same question. Can you explain why we need to set these two settings?

bzwei (approved):

hsong-rh (approved):
LGTM

PR #623, opened 2024-01-25
feat: [AAP-19536] Add yaml tag support for vaulted data

hsong-rh:
Resloves: [AAP-19536](https://issues.redhat.com/browse/AAP-19536)


tests/integration/api/test_extra_var.py
```diff
@@ -41,6 +51,21 @@ def test_create_extra_var(client: APIClient):
...
 
 
+@pytest.mark.django_db
+def test_create_vault_extra_var(client: APIClient):
+    data_in = {
+        "extra_var": TEST_VAULT_EXTRA_VAR,
+    }
+    response = client.post(f"{api_url_v1}/extra-vars/", data=data_in)
+    assert response.status_code == status.HTTP_201_CREATED
+    id_ = response.data["id"]
```
mkanoor, line 61: @hsong-rh Is there a reason for a trailing underscore?

mkanoor: @hsong-rh Can we fetch the limit from the REST API call and ensure it made a round trip.

mkanoor:
@lgalis this is to support the Extra Var having an embedded vaulted string.

lgalis (approved):
LGTM

tests/integration/api/test_extra_var.py
hsong-rh: changed

tests/integration/services/data/project-04/extensions/eda/rulebooks/demo_vault_str.yml
```diff
@@ -0,0 +1,18 @@
...
+  sources:
+    - name: vaulted 
+      ansible.eda.range:
+        limit: !vault |
+          $ANSIBLE_VAULT;1.1;AES256
+          36376437373138336362373430393138396139376631643039356330386235633536336666653433
+          3262333434663238663235353933646437633430616466310a323739313739356338663038356438
+          35623138663333633638323666633861373361663831613262383630343738373033323338326230
+          3935386638343137370a626161373036383764646633656163613734343362623434366165393063
+          3736
```
mkanoor, line 13: @hsong-rh Just for testing purposes we might have to use this same rulebook in the e2e test and we won't remember what vault password that was used to encrypt this string. Do you want to add a comment to indicate what secret was used to create this vaulted string. I am guessing on the server side for the E2E test we would have to create 1 test with a bad vault credential and one with a good vault credential to test the behavior

mkanoor:
@hsong-rh Should we try adding e2e tests for this in the eda-qa [repo](https://github.com/ansible/eda-qa) and add a rulebook into the @Alex-Izquierdo sample repo. We should be able to test this with a bad and good credential to count the number of rules that get triggered.

bzwei (approved):

Alex-Izquierdo (approved):

PR #626, opened 2024-01-29
feat: [AAP-19417] Enable to send vault credentials to rulebook

hsong-rh:
This PR adds the logic to send vault-type credentials to the rulebook engine via WebSocket channels

Resolves: [AAP-19417](https://issues.redhat.com/browse/AAP-19417)

### DIFF
--- a/src/aap_eda/wsapi/consumers.py
+++ b/src/aap_eda/wsapi/consumers.py
@@ -11,6 +11,7 @@ from django.conf import settings
 from django.db import DatabaseError
 
 from aap_eda.core import models
+from aap_eda.core.enums import CredentialType
 
 from .messages import (
     ActionMessage,
@@ -21,6 +22,8 @@ from .messages import (
     HeartbeatMessage,
     JobMessage,
     Rulebook,
+    VaultCollection,
+    VaultPassword,
     WorkerMessage,
 )
 
@@ -121,6 +124,12 @@ class AnsibleRulebookConsumer(AsyncWebsocketConsumer):
                 ssl_verify=settings.EDA_CONTROLLER_SSL_VERIFY,
             )
             await self.send(text_data=controller_info.json())
+
+        vault_data = await self.get_vault_passwords(message)
+        if vault_data:
+            vault_collection = VaultCollection(data=vault_data)
+            await self.send(text_data=vault_collection.json())
+
         await self.send(text_data=EndOfResponse().json())
 
         # TODO: add broadcasting later by channel groups
@@ -333,3 +342,35 @@ class AnsibleRulebookConsumer(AsyncWebsocketConsumer):
         awx_token = activation_instance.activation.awx_token
 
         return awx_token.token.get_secret_value() if awx_token else None
+
+    @database_sync_to_async
+    def get_vault_passwords(
+        self, message: WorkerMessage
+    ) -> tp.List[VaultPassword]:
+        """Get vault info from activation."""
+        activation_instance = models.RulebookProcess.objects.get(
+            id=message.activation_id,
+        )
+        activation = activation_instance.activation
+        vault_passwords = []
+
+        if activation.system_vault_credential:
+            vault = models.Credential.objects.filter(
+                id=activation.system_vault_credential.id
+            )
+        else:
+            vault = models.Credential.objects.none()
+
+        for credential in activation.credentials.filter(
+            credential_type=CredentialType.VAULT
+        ).union(vault):
+            vault_passwords.append(
+                VaultPassword(
+                    label=credential.vault_identifier,
+                    # TODO: Use pydantic secret feature (available > 2.0)
+                    # https://docs.pydantic.dev/latest/examples/secrets/
+                    password=credential.secret.get_secret_value(),
+                )
+            )
+
+        return vault_passwords
--- a/src/aap_eda/wsapi/messages.py
+++ b/src/aap_eda/wsapi/messages.py
@@ -85,6 +85,17 @@ class ControllerInfo(BaseModel):
     ssl_verify: str
 
 
+class VaultPassword(BaseModel):
+    type: str = "VaultPassword"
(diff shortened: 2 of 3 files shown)


src/aap_eda/wsapi/consumers.py
```diff
@@ -121,6 +124,13 @@ async def handle_workers(self, message: WorkerMessage):
                 ssl_verify=settings.EDA_CONTROLLER_SSL_VERIFY,
             )
             await self.send(text_data=controller_info.json())
+
+        vault_data = await self.get_vault_passwords(message)
+        if vault_data:
+            vault_collection = VaultCollection(data=vault_data)
+            logger.error(vault_collection.json())
```
mkanoor, line 131: @hsong-rh This line should be removed it will leak passwords

src/aap_eda/wsapi/consumers.py
```diff
@@ -208,9 +218,12 @@ def insert_event_related_data(self, message: AnsibleEventMessage) -> None:
 
     @database_sync_to_async
     def insert_audit_rule_data(self, message: ActionMessage) -> None:
-        job_id = message.job_id
-        job_instance = models.JobInstance.objects.filter(uuid=job_id).first()
-        job_instance_id = job_instance.id if job_instance else None
+        job_instance_id = None
+        if message.job_id:
```
mkanoor, line 222: @hsong-rh I think this should be in a separate PR. I think the newer versions of pydantic does the uuid validation I am not sure if the version we are using is doing the validation.

src/aap_eda/wsapi/consumers.py
```diff
@@ -330,3 +343,35 @@ def get_awx_token(self, message: WorkerMessage) -> tp.Optional[str]:
...
+                )
+            )
+
+        if activation.system_vault_credential:
+            password = (
+                activation.system_vault_credential.secret.get_secret_value()
+            )
+            vault_passwords.append(
+                VaultPassword(
+                    label=activation.system_vault_credential.key,
```
mkanoor, line 373: @hsong-rh Is this key or vault_credential?

tests/integration/wsapi/test_consumer.py
```diff
@@ -9,6 +9,7 @@
 from pydantic.error_wrappers import ValidationError
 
 from aap_eda.core import models
+from aap_eda.core.enums import CredentialType
```
mkanoor, line 12: @hsong-rh can we add a test with system vault credential. I am hoping the label gets set when we add it in.

src/aap_eda/wsapi/messages.py
```diff
@@ -85,6 +85,17 @@ class ControllerInfo(BaseModel):
     ssl_verify: str
 
 
+class VaultPassword(BaseModel):
+    type: str = "VaultPassword"
```
Alex-Izquierdo, line 89: - What is the purpose of a type str? I think it should be a enum: https://docs.pydantic.dev/2.3/usage/types/enums/
- Do we really need a type? What are the other types? because the field "password" is a mandatory string.

src/aap_eda/wsapi/messages.py
```diff
@@ -38,14 +38,14 @@ class ActionMessage(Message):
...
-    message: Optional[str]
-    kind: Optional[str]
+    playbook_name: Optional[str] = ""
+    job_template_name: Optional[str] = ""
+    organization: Optional[str] = ""
+    job_id: Optional[str] = ""
+    rc: Optional[int] = 0
+    delay: Optional[float] = 0.0
+    message: Optional[str] = ""
+    kind: Optional[str] = ""
```
Alex-Izquierdo, line 48: Why do we need empty default values when the type is already optional?

tests/integration/wsapi/test_consumer.py
hsong-rh: Will add it later when db migrations are merged

src/aap_eda/wsapi/messages.py
hsong-rh: They will be required when pydantic package upgrades

src/aap_eda/wsapi/messages.py
mkanoor: @Alex-Izquierdo Every message sent on the WebSocket Channel has a type they are all listed [here](https://github.com/ansible/eda-server/blob/eed9db3923c1c2b1674fdddd335142beb55d4768/src/aap_eda/wsapi/messages.py#L21)

tests/integration/wsapi/test_consumer.py
hsong-rh: @mkanoor tests are added

mkanoor (approved):

src/aap_eda/wsapi/messages.py
hsong-rh: Removed the changes required by pydantic 2.5.

tests/integration/wsapi/test_consumer.py
```diff
@@ -427,6 +523,172 @@ def get_job_instance_event_count():
     return models.JobInstanceEvent.objects.count()
 
 
+@database_sync_to_async
+def _prepare_activation_instance_with_vault_credentials():
```
bzwei, line 527: can you consolidate three _prepare functions into one with controlling parameters?
Similarly you may be able to combine three tests into one parameterized test.

Alex-Izquierdo (approved):

bzwei:
This PR has dependency on https://github.com/ansible/ansible-rulebook/pull/651, which has been merged.

tests/integration/wsapi/test_consumer.py
hsong-rh: These are the async function tests, each test case is marked as `transaction=True`. I don't think we can separate them as  parameterized tests.

Alex-Izquierdo (approved):

PR #628, opened 2024-01-29
chore: add docker-compose-mac-dev.yml

jshimkus-rh:
Makes docker-compose-mac.yml largely identical to docker-compose-dev.yaml.  Hopefully this will make it easier to incorporate future updates between (in both directions) the files easier.

### DIFF
--- a/docs/mac_development.md
+++ b/docs/mac_development.md
@@ -53,20 +53,28 @@ ps aux | grep "ssh -fnNT" |grep -v color
 3. export DOCKER_HOST=unix:///tmp/podman.sock
 
 
-## Start the app with available images in quay.io
+## There are three options for working with the app
+
+### 1. Use only images available from quay.io
 
 1. podman login quay.io
 2. export EDA_CONTROLLER_URL="your_controller_url"
 3. export EDA_CONTROLLER_SSL_VERIFY=no
 4. docker-compose -f ./tools/docker/docker-compose-mac.yml up
 
-## Start the app with your local changes
-
-You can build an image with your changes and use that image in docker-compose
+### 2. Use a local EDA image and everything else from quay.io
 
 1. podman build -t localhost/myserver -f tools/docker/Dockerfile .
 2. export EDA_IMAGE=localhost/myserver:latest
 3. docker-compose -f ./tools/docker/docker-compose-mac.yml up
+
+### 3. Use a closely matching environment of ./tools/docker/docker-compose-dev.yaml
+
+*   *Step 2 is optional as docker-compose-mac-dev.yaml defaults to using localhost/aap-eda*
+
+1. podman build -t localhost/aap-eda -f tools/docker/Dockerfile .
+2. [optional] export EDA_IMAGE=localhost/aap-eda :latest
+3. podman-compose -f ./tools/docker/docker-compose-mac-dev.yaml up
 
 You can now access the UI at <https://localhost:8443/eda/> with default login username and password(admin/testpass).
 
--- a/tools/docker/docker-compose-mac-dev.yaml
+++ b/tools/docker/docker-compose-mac-dev.yaml
@@ -1,0 +1,194 @@
+x-environment:
+  &common-env
+  - EDA_DB_HOST=postgres
+  - EDA_MQ_HOST=redis
+  - DJANGO_SETTINGS_MODULE=${DJANGO_SETTINGS_MODULE:-aap_eda.settings.development}
+  - EDA_ALLOWED_HOSTS=['*']
+  - EDA_DEPLOYMENT_TYPE=${EDA_DEPLOYMENT_TYPE:-podman}
+  - EDA_WEBSOCKET_BASE_URL=${EDA_WEBSOCKET_BASE_URL:-ws://eda-ws:8000}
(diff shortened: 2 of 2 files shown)


tools/docker/docker-compose-mac.yml
```diff
@@ -58,27 +83,41 @@ services:
       timeout: 5s
       retries: 10
     volumes:
-      - ${EDA_HOST_PODMAN_SOCKET_URL:-/run/user/501/podman/podman.sock}:/run/podman/podman.sock:z
+      - ${EDA_HOST_PODMAN_SOCKET_URL}:/run/podman/podman.sock:z
```
Alex-Izquierdo, line 86: This makes this environment variable mandatory for users. Is that what you want? It would be a breaking change and probably should be documented.

jshimkus-rh: @Alex-Izquierdo `EDA_HOST_PODMAN_SOCKET_URL` is defaulted in to `/run/user/501/podman/podman.sock` in `common-env` to eliminate the redundancy in its previous uses.

Alex-Izquierdo: I think that won't work, at least with docker-compose v1 (which we have to support) 

Common env is for the environment variables injected into the pods. Here, the scope is the docker-compose process itself, not pods. Docker-compose takes the value from the user environment. If this environment variable is not defined docker-compose would fail.

jshimkus-rh: I'd be inclined to agree if it weren't that I don't set that variable and both docker-compose (v1.29.2) and podman-compose (v1.0.6) build and run without exception that I've encountered.  If others have different experiences I'd be interested in hearing.

Alex-Izquierdo: Well, if it works, I'm wrong. :)

mkanoor:
@jshimkus-rh Are there pre-reqs to this PR
When I run it with my current setup the pods don't startup
```
WARNING: The EDA_HOST_PODMAN_SOCKET_URL variable is not set. Defaulting to a blank string.
Creating volume "docker_podman_data" with default driver
Pulling podman-pre-setup (quay.io/containers/podman:v4)...
7eaf774f9ce1: Download complete
c8d184852bb2: Download complete
34fa8cad82a0: Download complete
74eee9c10a6d: Download complete
cb70ad9471e1: Download complete
4aeaa23602b2: Download complete
d367bcc8d193: Download complete
39d83fc23602: Download complete
80f84ed70059: Download complete
1017ffc03448: Download complete
Pulling eda-scheduler (localhost/aap-eda:)...
ERROR: connection refused
```

jshimkus-rh:
@mkanoor That's what @Alex-Izquierdo was concerned about.  Try as I might I couldn't make that happen; don't know why and it's not worth it right now to figure it out.  I've gone back to the previous handling (see my latest commit) of `EDA_HOST_PODMAN_SOCKET_URL` and try it with that; presumably it should not be a problem, now.

tools/docker/docker-compose-mac.yml
```diff
@@ -33,7 +54,10 @@ services:
         condition: service_healthy
 
   eda-api:
-    image: "${EDA_IMAGE:-quay.io/ansible/eda-server:main}"
+    image: "${EDA_IMAGE:-localhost/aap-eda}"
```
mkanoor, line 57: @jshimkus-rh Does this mean now we have to provide an EDA_IMAGE env var to get to use the default image from quay.io. Shouldn't this default the other way around. By default we use quay unless the developer wants to override it. If we have a UI developer who wants to use podman they have to change the EDA_IMAGE to point to quay.io/ansible/eda-server:main because they will never build the eda-server locally.

jshimkus-rh: The question I infer from this is what is the purpose of the different compose files?
As  a developer of the "engine" (i.e., something functional even w/o a UI; the desirability of which, product-wise, appears to be somewhat unclear based on things I've heard) who uses a Mac I'm interested in default establishment of an "engine" development environment.  The differences between the *-mac.yaml and *-dev.yml files which have accumulated making the two development environments differ is a "bad" thing and my motivation in more closely aligning them.
That is not to say there isn't a need for what you describe, though rather than having a single file that has to pull double-duty I think it is preferable to have a specific, shall we say, "canned" compose for that purpose.  With changes that have been made in relation to podman support it appears that docker-compose-stage.yaml provides that.  If not, we could easily create such.

mkanoor: @Alex-Izquierdo We have 2 kinds of developers 1 developing the UI and another developing the API Server. If I am a UI developer I don't want to build the API Server image and if I am an API Server I don't want to build the UI image. If this tool needs to be embraced by both kinds of developers the default needs to be fetch standard quay images and a developer can override their local ones to use the image of their choice. In Mac there is only 1 docker file unlike the dev and stage in Llnux. If we are using local builds we just override the EDA_IMAGE and everything else falls into place. When we started the EDA-SERVER development we used to build the UI and server images locally and use them locally. Now the patterns are changing we use standard quay images as much as possible and even for local PR testing we have been pushing images to our private quay repos and making the images public so we can test without waiting for a final approval on the PR. Using quay images improves the speed of development

PR #633, opened 2024-01-30
fix: importing package from ansible_base.lib

bzwei:
Due to recent restructure of dependency library django-ansible-base, `lib` was inserted into the package path.


Dostonbek1 (approved):
Were you able to test this locally to make sure we are not hitting any errors?

bzwei:
yes. it is running in my local

PR #634, opened 2024-01-30
refactor: process_fqcn and process_id POC

Alex-Izquierdo:
A PoC of process_fqcn and process_id for the manager and rulebook_process


src/aap_eda/core/models/activation.py
```diff
@@ -193,22 +195,36 @@ class Meta:
...
+    activation = models.ForeignKey(
+        "Activation", on_delete=models.CASCADE, null=True
+    )
     started_at = models.DateTimeField(auto_now_add=True, null=False)
     updated_at = models.DateTimeField(null=True)
     ended_at = models.DateTimeField(null=True)
     activation_pod_id = models.TextField(null=True)
     status_message = models.TextField(null=True, default=None)
     log_read_at = models.DateTimeField(null=True)
+    parent_id = models.TextField(null=False)
```
bzwei, line 209: shouldn't it be an integer?

src/aap_eda/core/models/activation.py
```diff
@@ -193,22 +195,36 @@ class Meta:
...
+    parent_fqcn = models.TextField(null=False)
+
+    def get_parent(self):
+        module_name, class_name = self.parent_fqcn.rsplit(".", 1)
```
bzwei, line 213: This is can be done by `from django.utils.module_loading import import_string`
```
parent_model = import_string(self.parent_fqcn)
```

bzwei: do this only if `self.parent_fqcn` is not None, otherwise return `self.activation`

src/aap_eda/services/activation/manager.py
```diff
@@ -1018,12 +1019,14 @@ def update_logs(self):
             return
 
     def _create_activation_instance(self):
+        fqcn = get_fully_qualified_name(self.db_instance)
```
mkanoor, line 1022: @Alex-Izquierdo We also need a get_parent method that does something like this

```
from django.core.exceptions import ObjectDoesNotExist
from django.utils.module_loading import import_string
def get_parent(self):
      try:
             klass = import_string(self.parent_fqcn)
            return klass.objects.get(id=self.parent_id)
    except ObjectDoesNotExist:
        LOGGER.exception(
            f"Parent id  {self.parent_id} for class {self.parent_fqcn} no longer exists, "
            )
 ````

src/aap_eda/core/models/activation.py
mkanoor: @Alex-Izquierdo This is still a valid function in Django 5.0
https://docs.djangoproject.com/en/5.0/ref/utils/#django.utils.module_loading.import_string

src/aap_eda/core/models/activation.py
Alex-Izquierdo: changed

PR #636, opened 2024-01-31
feat: Add EventStream model

Alex-Izquierdo:
- Add EventStream model as an almost replica of the activation. 
- Move the logic to manage the status/status_message fields into the StatusHandlerModelMixin to share it between Activation and EventStream. 
- Update tests accordingly. 

Jira: https://issues.redhat.com/browse/AAP-20050


src/aap_eda/core/models/event_stream.py
```diff
@@ -0,0 +1,92 @@
+#  Copyright 2022 Red Hat, Inc.
```
mkanoor, line 1: @Alex-Izquierdo should this be 2024?

mkanoor:
@Alex-Izquierdo Are you missing the RBAC stuff which was added for Sources with this [PR](https://github.com/ansible/eda-server/pull/587/files)

Alex-Izquierdo:
@mkanoor, thanks, good catch! Added.

src/aap_eda/core/models/event_stream.py
Alex-Izquierdo: should it be the year when the application was licensed? I'm not sure. Otherwise we should update the year in all the places, don't?

Alex-Izquierdo: changed

hsong-rh (approved):

mkanoor (approved):

PR #637, opened 2024-01-31
refactor: orchestrator to support both Activation and EventStream

bzwei:
Hasn't been integrated with #634. Will need some extra work once #634 is merged.

https://issues.redhat.com/browse/AAP-20031


src/aap_eda/api/views/activation.py
```diff
@@ -224,7 +225,7 @@ def enable(self, request, pk):
                 "modified_at",
             ]
         )
-        start_activation(activation_id=pk)
+        start_activation(KLASS_NAME, pk)
```
mkanoor, line 228: @bzwei do you want to use kwargs style as opposed to positional args for your functions. I am guessing earlier it was a kwargs style function.

src/aap_eda/core/models/queue.py
```diff
@@ -23,8 +23,14 @@ class Meta:
         ordering = ["id"]
 
     request = models.TextField(null=False, choices=ActivationRequest.choices())
+    process_parent_fqcn = models.TextField(
+        null=True, default="aap_eda.core.models.Activation"
+    )
+    process_parent_id = models.BigIntegerField(null=True)
```
mkanoor, line 29: @bzwei Should we allow NULLs can we use the same migration technique as in PR #634 which sets all the data in the new columns and leaves the old one in place.

src/aap_eda/tasks/orchestrator.py
```diff
@@ -26,65 +29,67 @@
 LOGGER = logging.getLogger(__name__)
 
 
-def _manage_activation_job_id(activation_id: int) -> str:
+def _manage_activation_job_id(process_parent_type: str, id: int) -> str:
     """Return the unique job id for the activation manager task."""
-    return f"activation-{activation_id}"
+    return f"{process_parent_type}-{id}"
```
mkanoor, line 34: @bzwei Will this be tricky when we are in the process of migration the names are going to change? Or it won't matter because when we stop all our services and restart we won't have RQ jobs with the old job_id. Is this job_id stored in the DB?

src/aap_eda/api/views/activation.py
bzwei: should not matter, unless we have a policy to always use kwargs

src/aap_eda/core/models/queue.py
bzwei: yes, we can

src/aap_eda/tasks/orchestrator.py
bzwei: job_id is not stored in our DB. I think a restart will clear everything

mkanoor (approved):

src/aap_eda/api/views/activation.py
```diff
@@ -120,7 +124,10 @@ def perform_destroy(self, activation):
         activation.status = ActivationStatus.DELETING
         activation.save(update_fields=["status"])
         logger.info(f"Now deleting {activation.name} ...")
-        delete_activation(activation_id=activation.id)
+        delete_activation(
```
mkanoor, line 127: @bzwei Lets change the start/stop/delete/restart activation to start/stop/delete/restart rulebook_process.

Alex-Izquierdo (approved):

src/aap_eda/services/activation/restart_helper.py
```diff
@@ -23,25 +23,35 @@
 LOGGER = logging.getLogger(__name__)
 
 
-def system_restart_activation(activation_id: int, delay_seconds: int) -> None:
+def system_restart_activation(
+    process_parent_type: str, id: int, delay_seconds: int
```
jshimkus-rh, line 27: I have reservations about referring to a `process_parent_type`.  While I grasp that the intention seems to be to refer to X (where X is the thing to process) versus X-Instance (a specific processing) the processing of X via an X-Instance is incidental (architecture/design-wise, not in an auditable sense) to what we want to accomplish which is processing an X. 
I think the name should reflect this higher-level conceptual idea rather than implicitly exposing an aspect of the implementation.

bzwei: @jshimkus-rh Do you have a recommended name to replace `process_parent_type`? Here the `process` is a short for `rulebook_process`

jshimkus-rh: I think simply dropping `parent` from the name  would suffice.
This would also apply more broadly to every other use of `parent` in the change excepting anything that actually has a parent-child relation, of course.

PR #645, opened 2024-02-03
Combine event streams with activations

hsong-rh:


benthomasson (changes requested):
Don't use BigIntegerField for parent_id.  Use separate foreign key fields for activation and event stream.   In fact you have OneToOneField relations on activation and event stream so you already have the backward reference.

tests/integration/api/test_event_stream.py
```diff
@@ -0,0 +1,207 @@
+#  Copyright 2023 Red Hat, Inc.
```
mkanoor, line 1: @hsong-rh 2024?

src/aap_eda/api/views/event_stream.py
```diff
@@ -0,0 +1,323 @@
+#  Copyright 2023 Red Hat, Inc.
```
mkanoor, line 1: @hsong-rh 2024?

Alex-Izquierdo:
Do we want to continue working in this branch?

hsong-rh:
> Do we want to continue working in this branch?

no, will move all changes to the main branch. Here is a reference.

PR #646, opened 2024-02-04
Generic relations

mkanoor:
This PR shows the changes needed to use generic relations
    instead of using our own implementation. I have left out
    the migration of getting the old activation instances and
    setting up the relationship. Once we have a direction
    that can be implemented.


jshimkus-rh (commented):
I thought the use of generics had been, more-or-less, declared as a non-starter?
Perhaps this is meant as an example only?

benthomasson:
Generic relations are a bad database architecture pattern that we should not use.   This complicates all the code that has to deal with tables that have generic relations and removes benefits of database indexes for queries.

PR #648, opened 2024-02-05
[AAP-18826] Integrate Resources API

Dostonbek1:
* implement resource_registry app from django-ansible-base
* update django-ansible-base to latest commit from devel to import resource_registry changes

JIRA: [AAP-18826](https://issues.redhat.com/browse/AAP-18826)

NOTE: will keep it in draft for now until we merge devel branch into main, so that this PR will go into main branch.


AlanCoding:
I don't have approval permission here, but :+1: from me.

We are adding teams & organizations in a separate branch. I was already saying it makes sense to enable this first with only users. Letting @newswangerd know about this approach, don't expect there will be any disagreement.

Dostonbek1:
Replaced by this PR https://github.com/ansible/eda-server/pull/652

PR #653, opened 2024-02-07
Create an `Orchestrator ABC` and use for request management.

jshimkus-rh:
I started out with relocating `services/activation/engine` to `services/process/engine` as part of separating its functionality from specific uses.  In the long-term this is still applicable, but not the main thrust of this change.  For that reason ignore the relocation (for the time being) and focus on the creation of `Orchestrator` as an `ABC` and the realized `ActivationOrchestrator`.

`Orchestrator` defines an `ABC` from which concrete subclasses are defined.  Each of these subclasses realizes the orchestrator functionality for a particular request type (e.g., activation).  Via the definition in each subclass of `@staticmethod manage` and the use of same when enqueuing requests the type of the request is implicit; there is no need to pass identifying information along with the request.  The guts of each `manage` method can be realized in an analogous manner to that currently used for activations.

Although not required, as implemented there is a presumption of separate queues and thus separate worker(s) for each request type for complete compartmentalization of each type.  As stated this is not a requirement; the same queue and worker(s) can be utilized for all request types.  The only practical effect is how max running requests may be enforced (per-type or in aggregate); regardless, each can be supported.


eclarizio:
Is there a JIRA issue that is related to this work? If so, can you update the description to include a link to the issue? If not, maybe we can put this on hold and discuss later.

PR #654, opened 2024-02-07
feat: enable relationship with parents in rulebook process

Alex-Izquierdo:
- Two optional foreign keys for event_streams and activations
- A type field populated from the parent defined in an enum
- get_parent function
- model validation to ensure the parent and type
- Update tests accordingly
Jira: https://issues.redhat.com/browse/AAP-20503


mkanoor:
@Alex-Izquierdo do we need a migration to set the parent_type in existing rows in the DB and set it to Activation

Alex-Izquierdo:
Look carefully at the changes in the RulebookProcess model, since I moved it to a new file, you miss the diff.

src/aap_eda/core/migrations/0021_rulebookprocess_event_stream_and_more.py
```diff
@@ -0,0 +1,48 @@
...
+        ),
+        migrations.AddField(
+            model_name="rulebookprocess",
+            name="parent_type",
+            field=models.TextField(
+                choices=[
+                    ("activation", "activation"),
+                    ("event_stream", "event_stream"),
+                ],
+                default=aap_eda.core.enums.ProcessParentType["ACTIVATION"],
```
mkanoor, line 34: @Alex-Izquierdo does this take care of existing rows or is it just using a default.

mkanoor (approved):

src/aap_eda/core/migrations/0021_rulebookprocess_event_stream_and_more.py
Alex-Izquierdo: The field is mandatory, so the default is applied to all records. Since the activation was previously mandatory, we don't need any other action in the migration.

hsong-rh (approved):

bzwei (approved):

src/aap_eda/core/migrations/0021_rulebookprocess_event_stream_and_more.py
mkanoor: ok

PR #657, opened 2024-02-07
feat: combine event streams with activations

hsong-rh:
https://issues.redhat.com/browse/AAP-20144
https://issues.redhat.com/browse/AAP-20036
https://issues.redhat.com/browse/AAP-19799

This is the first attempt at getting a basic fanout working with the latest model changes. We can create event streams attach them to activations, report on the events for each one separately, get the logs for each one separately. With the basic stuff working with complete UI support we should be able to continue parallel work in QE and Dev to harden some of the internal plumbing needed to get fanout to closure. There are unit tests for most of the changes. We would start writing the first E2E tests once we get a client generated from this.

### Testing this PR

To start up the test environment we need the 2 images

You can build the image from this PR or use the ones below
* export EDA_IMAGE=quay.io/hsong_rh/eda-server:final
* export EDA_UI_IMAGE=quay.io/lgalis/eda-ui:latest_arm
   On Linux if using amd chip use
* export EDA_UI_IMAGE=quay.io/lgalis/eda-ui:latest_amd

Start up the server using docker_compose or minikube
Log into the EDA Server
### Steps
1. Create a Decision Environment  (e.g. **quay**) pointing to 
```
quay.io/ansible/ansible-rulebook:main
```
2. Create a new Project with the Name set to  **demo** and the **SCM URL** set to
``` 
 https://github.com/mkanoor/eda-project
```
3. Create an Event Stream
     * Name
     ``` 
       Testing Generic
    ```
     * Source type
     ```
         ansible.eda.generic
     ```
     * Decision environment
     ```
       quay
     ```
     * Args:
```yaml
startup_delay: 60
payload:
   - i : 1
   - i : 2
   - name: Fred
```       

4. Create a new Rulebook Activation pick the project **demo** and the rulebook **[single_rule.yml](https://github.com/mkanoor/eda-project/blob/main/rulebooks/single_rule.yml)** and attach the "**Testing Generic**" EventStream to it
5. Watch for results in the Logs or in the Rule Audit, you will see 2 **myrule1** which is the rule triggered when the condition matches.


src/aap_eda/api/serializers/activation.py
```diff
@@ -200,8 +239,35 @@ def create(self, validated_data):
         validated_data["rulebook_rulesets"] = rulebook.rulesets
         validated_data["git_hash"] = rulebook.project.git_hash
         validated_data["project_id"] = rulebook.project.id
+        if validated_data.get("event_streams"):
+            validated_data["rulebook_rulesets"] = self._updated_ruleset(
+                validated_data
+            )
         return super().create(validated_data)
 
+    def _updated_ruleset(self, validated_data):
```
mkanoor, line 248: @hsong-rh self is not being used can we make it a static function and just call it.

src/aap_eda/settings/default.py
```diff
@@ -478,3 +478,17 @@ def _get_secret_key() -> str:
 ANSIBLE_BASE_JWT_KEY = settings.get(
     "ANSIBLE_BASE_JWT_KEY", "https://localhost"
 )
+
+# TODO: fix to get db address for activation
+ACTIVATION_DB_HOST = "host.containers.internal"
```
mkanoor, line 483: @Alex-Izquierdo How do we solve this for the installer, operator, Minikube and podman. The current setting works for podman. For minikube or operator it might have to be a different value. We should track this separately.

mkanoor: @hsong-rh I think this should be
```
ACTIVATION_DB_HOST = settings.get("ACTIVATION_DB_HOST", "host.containers.internal")
```

Since we default the deployment to podman this would align well and in the file https://github.com/ansible/eda-server/blob/rework-activation/tools/deploy/environment.properties.example we add an entry for K8S to indicate that a value is needed for minikube. Not sure what that default value for ACTIVATION_DB_HOST should be.
@msmagnanijr this might also effect the work you are doing for the operator based deployment

src/aap_eda/api/exceptions.py
```diff
@@ -95,3 +95,18 @@ class InvalidWebsocketHost(APIException):
...
+    status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
+    default_detail = (
+        "Configuration Error: Event stream template rulebook not found"
+    )
+
+
+class MissingEventStreamRulebookKeys(APIException):
+    status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
+    default_detail = (
+        "Configuration Error: Event stream template rulebook missed required "
```
mkanoor, line 110: @hsong-rh "missed required" should be "is missing required"

src/aap_eda/api/serializers/activation.py
```diff
@@ -11,18 +11,55 @@
...
+    for event_stream_id in validated_data["event_streams"]:
+        event_stream = models.EventStream.objects.get(id=event_stream_id)
+
+        if event_stream.rulebook:
+            rulesets = yaml.safe_load(event_stream.rulebook.rulesets)
+        else:
+            rulesets = yaml.safe_load(PG_NOTIFY_TEMPLATE_RULEBOOK_DATA)
+        extra_vars = rulesets[0]["sources"][0]["extra_vars"]
+        extra_vars = substitute_extra_vars(
+            event_stream.__dict__, extra_vars, [], "password"
```
mkanoor, line 53: @hsong-rh Can you put a TODO for the password once @bzwei has implemented the ansible-rulebook with valut support we will fill in the correct password and encrypt the data with it using ansible-vault encrypt feature.

src/aap_eda/api/serializers/activation.py
```diff
@@ -357,6 +434,7 @@ class PostActivationSerializer(serializers.ModelSerializer):
         validators=[validators.check_if_extra_var_exists],
     )
     awx_token_id = serializers.IntegerField(
+        required=False,
```
mkanoor, line 437: @Alex-Izquierdo @jshimkus-rh This is a nasty one it happens because we are passing in the EventStream object as Activation object. Once we resolve https://issues.redhat.com/browse/AAP-20044 and call the appropriate validate function in EventStream or[ activation here](https://github.com/ansible/eda-server/blob/eb2df468cb2288fda08772cf83ec4464bb4b10d0/src/aap_eda/services/activation/manager.py#L202) it should resolve itself.

src/aap_eda/api/serializers/event_stream.py
```diff
@@ -0,0 +1,218 @@
...
+                settings.PG_NOTIFY_TEMPLATE_RULEBOOK,
+            )
+            raise MissingEventStreamRulebook
+
+        required_keys = ["type", "name", "args"]
+        rulesets = yaml.safe_load(rulebook.rulesets)
+        for ruleset in rulesets:
+            sources = ruleset.get("sources", [])
+            for source in sources:
+                complementary_source = source.get("complementary_source")
```
mkanoor, line 59: @hsong-rh if complementary_source is missing we should raise an Exception.

src/aap_eda/api/serializers/activation.py
```diff
@@ -11,18 +11,55 @@
...
+    sources_info = []
+
+    for event_stream_id in validated_data["event_streams"]:
+        event_stream = models.EventStream.objects.get(id=event_stream_id)
+
+        if event_stream.rulebook:
+            rulesets = yaml.safe_load(event_stream.rulebook.rulesets)
+        else:
+            rulesets = yaml.safe_load(PG_NOTIFY_TEMPLATE_RULEBOOK_DATA)
+        extra_vars = rulesets[0]["sources"][0]["extra_vars"]
```
Alex-Izquierdo, line 51: We need a ticket to implement validation of the content of the rulesets @mkanoor @hsong-rh 
We should not accept that doesn't meet the expected schema (the new keys, only one ruleset, etc...)

Alex-Izquierdo: until implement a validation of the content, I think all these direct accesses to the rulesets content should be within a try blocks

src/aap_eda/api/serializers/activation.py
```diff
@@ -111,6 +162,10 @@ def to_representation(self, activation):
             CredentialSerializer(credential).data
             for credential in activation.credentials.all()
         ]
+        event_streams = [
```
Alex-Izquierdo, line 165: This would return a list of objects. Do we want this for the list view? Should be just the list of id's?

src/aap_eda/api/serializers/activation.py
```diff
@@ -215,6 +282,8 @@ class Meta:
             "git_hash",
             "status_message",
             "activation_id",
+            "event_stream_id",
+            "parent_type",
```
Alex-Izquierdo, line 286: I think parent_type is internal and we don't want to expose this field to the public api. Not even makes sense because it will be always the only same value.

src/aap_eda/api/serializers/activation.py
```diff
@@ -297,7 +367,8 @@ def to_representation(self, activation):
             else None
         )
         activation_instances = models.RulebookProcess.objects.filter(
-            activation_id=activation.id
+            parent_type=ProcessParentType.ACTIVATION,
+            id=activation.id,
```
Alex-Izquierdo, line 371: Are you using the activation id to filter the id of the rulebook process. This would return wrong results. I think this line doesn't need to be changed. You want the rulebook_processes related with the activation id.

src/aap_eda/api/serializers/activation.py
jshimkus-rh: Making this a static function solely on `self` not being referenced is not a sound basis; `self` is only visible to the implementation as that is how `python` implemented object handling.
If there is a case where it is known that this will be reused then it's reasonable.

src/aap_eda/settings/default.py
Alex-Izquierdo: This hostname is going to be used for both, event_streams and activations, right? I feel the name might be confusing. 

I see multiple ways to handle it. 

For custom values of PG_NOTIFY_DSN we must state that the hostname must be reachable and resolvable by the containers. 

For our default we could handle it based on the value provided in DATABASES. Probably not even we need a new variable for this. I think the logic can be simple: we only want "host.containers.internal" when the deployment type is podman and the database is localhost. For k8s this value would be never localhost but a service or external.

src/aap_eda/api/serializers/activation.py
Alex-Izquierdo: A TODO comment would be good

src/aap_eda/api/serializers/activation.py
jshimkus-rh: I should add that making it a static method of the class rather than an external function would be a preferable approach, barring the need for reuse.

src/aap_eda/api/serializers/activation.py
mkanoor: @Alex-Izquierdo We were having UI issues and were not sure if the UI should be checking for NULL values and would it need a type check.

src/aap_eda/api/serializers/activation.py
mkanoor: @Alex-Izquierdo The UI needs the object not the ID since it gets displayed.

mkanoor: @hsong-rh @lgalis mentioned she just needs the id and name so we can trim the object

src/aap_eda/api/serializers/activation.py
Alex-Izquierdo: What are the issues? The UI should not need to do anything with the parent_type, it is internal. The endpoint should return the right type (activations in this case) For the UI, the type is implicit in the endpoint (activation_instances or event_stream_instances)

src/aap_eda/services/activation/manager.py
```diff
@@ -228,9 +228,13 @@ def _check_latest_instance_and_pod_id(self) -> None:
...
 
     def _check_non_finalized_instances(self) -> None:
-        instances = models.RulebookProcess.objects.filter(
-            activation=self.db_instance,
-        )
+        args = {"parent_type": self.db_instance_type}
+        if self.db_instance_type == ProcessParentType.EVENT_STREAM:
+            args["event_stream_id"] = self.db_instance.id
+        else:
+            args["activation_id"] = self.db_instance.id
```
Alex-Izquierdo, line 235: This code and the similar in https://github.com/ansible/eda-server/pull/657/files#diff-767f95d0b4918e197417276be62147629e68416233ba137c48b8c171b570f73cR1054
can simpler if you directly use `self.db_instance_type`. The value of the enum already matches the name of the field. Not even you need to pass the parent_type. 
```
args = {f"{self.db_instance_type}": self.db_instance}
```
We don't want to fill the code with these conditionals and we don't need it.

mkanoor:
UI PR https://github.com/ansible/ansible-ui/pull/1571

mkanoor:
Setting up Decision Environment
DecisionEnvionment

Setting up Event Stream
EventStream

Setting up Activation

Activation


Rule Audit
RuleAudit

mkanoor (approved):

bzwei (approved):

bzwei:
Tested and worked fine

PR #658, opened 2024-02-08
[RBAC] Add relation to Organization in models

Dostonbek1:
* Add Organization as foreign key to object models exposed on the API (exception: Users have M2M relation)
* Add migration script to create a default org and add all existing data to default org
* Update `save` method in models to add them to default org if they are not part of any other orgs


src/aap_eda/core/migrations/0022_activation_organization_auditaction_organization_and_more.py
```diff
@@ -0,0 +1,139 @@
...
+        migrations.AlterField(
+            model_name="team",
+            name="organization",
+            field=models.ForeignKey(
+                null=True,
+                on_delete=django.db.models.deletion.CASCADE,
+                to="core.organization",
+            ),
+        ),
+    ]
```
AlanCoding, line 139: The `0021_organization_team` migration is the last one that runs before DAB RBAC migrations. I think it makes it the easiest for us to put these kind of changes in that migration file or before. This means that there is no possibility that DAB RBAC will initialize without these models, and we don't need to worry any corner cases that might result from that. Just adding these steps to that file seems like the simplest thing to do right now.

src/aap_eda/core/migrations/0023_add_default_organization.py
```diff
@@ -0,0 +1,76 @@
...
+from django.utils import timezone
+
+
+def create_default_org(apps, schema_editor):
+    Organization = apps.get_model("core", "Organization")  # noqa: N806
+
+    db_alias = schema_editor.connection.alias
+
+    now = timezone.now()
+    Organization.objects.using(db_alias).create(
```
AlanCoding, line 12: might be good to call `get_or_create` just in case you or I want to reverse and re-apply the migration for testing.

src/aap_eda/core/models/user.py
```diff
@@ -32,6 +34,17 @@ class User(AbstractUser):
     roles = models.ManyToManyField("Role", related_name="users")
     modified_at = models.DateTimeField(auto_now=True, null=False)
     is_service_account = models.BooleanField(default=False)
+    organizations = models.ManyToManyField(
```
AlanCoding, line 37: I don't want to add organizations to the User model, because that is tracked with an organization role. That role doesn't exist yet, the idea is that a migration will create it. DAB RBAC has some utilities for this but I don't plan to use it for EDA because it doesn't use the Django Meta to specify permissions. So plan is for those roles to be the same as current roles, python / dict-defined

src/aap_eda/core/models/team.py
```diff
@@ -1,5 +1,8 @@
 from ansible_base.lib.abstract_models import AbstractTeam
+from django.db import models
 
 
 class Team(AbstractTeam):
-    pass
+    organization = models.ForeignKey(
+        "Organization", on_delete=models.CASCADE, null=True
```
AlanCoding, line 7: The `AbstractTeam` should already define the `organization` field in DAB. I would prefer to remove this field and see what makemigrations does.

src/aap_eda/core/migrations/0023_add_default_organization.py
```diff
@@ -0,0 +1,76 @@
...
+        modified_on=now,
+    )
+
+
+def delete_default_org(apps, schema_editor):
+    Organization = apps.get_model("core", "Organization")  # noqa: N806
+    db_alias = schema_editor.connection.alias
+
+    Organization.objects.using(db_alias).filter(
+        name=settings.DEFAULT_ORGANIZATION_NAME
```
AlanCoding, line 25: umm... wouldn't this cascade delete everything?

src/aap_eda/core/models/activation.py
```diff
@@ -113,6 +117,12 @@ class Meta:
         related_name="+",
     )
 
+    def save(self, *args, **kwargs):
+        super().save(*args, **kwargs)
+        if not self.organization:
+            self.organization = Organization.objects.get_default()
+            super().save(update_fields=["organization"])
```
AlanCoding, line 124: Interesting approach. It should unblock some other work before serializer changes happen.

AlanCoding (changes requested):
Recap:
 - I'd like to move the migration steps to an earlier migration before DAB RBAC
 - remove organization field from team model
 - remove organizations from user model
 
 I'd like to get those critical changes in and then merge into the feature branch :+1:

src/aap_eda/core/models/user.py
Dostonbek1: I think moving it to the organization model would essentially be the same thing and I guess we wouldn't want that:
```
class Organization(AbstractOrganization):
    users = models.ManyToManyField(
        settings.AUTH_USER_MODEL,
        related_name="organizations",
        help_text="The list of users in this organization.",
    )

```

AlanCoding: I'm saying don't move it to the organization model. Just delete this field. In the data migration for creating roles (`RoleDefinition` in dab_rbac) we will create a role for what Gateway sees as the `Organization.users` relationship.

Then in the EDA auth class for JWT consumption, the org-user mapping will be applied to the applicable RoleDefinition. This already exists in a partial form, with the existing EDA `Role` model here.

https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/jwt_consumer/eda/auth.py

src/aap_eda/core/migrations/0023_add_default_organization.py
Dostonbek1: Yeah it would.. That's another thing I'd like clarity on: how should org foreign key field be defined in models, in terms on_delete, or having CASCADE delete there is fine but we fix this reverse function to loop through all existing objects to set their org to null then delete the default org.

AlanCoding: For the reverse migration, I suspect it might work better to do nothing. Then it drops the field, but that doesn't cascade delete anything.

AlanCoding (approved):

AlanCoding (approved):

PR #665, opened 2024-02-09
[AAP-9727] Remove /tasks endpoint

Dostonbek1:
* remove /tasks endpoints
* remove /inventory model

JIRA: [AAP-9727](https://issues.redhat.com/browse/AAP-9727)


Alex-Izquierdo (commented):
LGTM, but, what about https://github.com/ansible/eda-server/blob/main/src/aap_eda/api/serializers/project.py#L129 ?

src/aap_eda/api/serializers/project.py
```diff
@@ -126,7 +124,6 @@ def to_representation(self, project):
             "credential": credential,
             "import_state": project["import_state"],
             "import_error": project["import_error"],
-            "import_task_id": project["import_task_id"],
```
Dostonbek1, line 129: @Alex-Izquierdo It is removed here. You might looking at the wrong branch, maybe?

Alex-Izquierdo (approved):

hsong-rh (approved):

mkanoor:
@Dostonbek1 @Alex-Izquierdo @hsong-rh There are references to the import_task_id in ansible/ui repo, there is also an openapi.json in that repo we would have to check with @lgalis  on this
The E2E tests are also failing because of this change. https://github.com/ansible/eda-server/actions/runs/7846940638/job/21414768516?pr=657

Alex-Izquierdo:
The main problem right now is the generation of the new openapi spec. @Dostonbek1 is looking it. I understood that he did the double check with @lgalis

PR #666, opened 2024-02-12
ci: ensure application runs and validate openapi spec

Alex-Izquierdo:
Ensure the application starts and validate the openapi spec. 
Jira: https://issues.redhat.com/browse/AAP-20673


Alex-Izquierdo:
The CI should not pass because right now the openapi spec is broken, this is expected and one of reasons to create this PR. But if you agree we can merge with no CI pipeline passing.

msmagnanijr (approved):

PR #669, opened 2024-02-12
refactor: [AAP-20044] implement interface for "containerable" entities

jshimkus-rh:
https://issues.redhat.com/browse/AAP-20044
https://issues.redhat.com/browse/AAP-20022
https://issues.redhat.com/browse/AAP-20019


src/aap_eda/services/activation/manager.py
```diff
@@ -1059,54 +1054,51 @@ def _create_activation_instance(self):
...
-        else:
-            context = {}
+        try:
+            container_params = self.db_instance.get_container_parameters()
+        except ContainerableInvalidError:
+            msg = (
+                f"Activation {self.db_instance.id} not valid, "
+                "container request cannot be built."
+            )
+            LOGGER.exception(msg)
```
mkanoor, line 1064: @jshimkus-rh should this be LOGGER.error as opposed to LOGGER.exception otherwise it will raise an exception and kill the worker.

jshimkus-rh: I essentially copy/pasted the exception handling in `_build_cmdline` which was already using `LOGGER.exception`; I would think both should have the same behavior.  Being `error` is probably better for both, unless there's something special about the command line creation?

tests/integration/services/activation/test_activation.py
```diff
@@ -0,0 +1,125 @@
...
+
+import pytest
+
+from aap_eda.core import models
+
+PROJECT_GIT_HASH = "684f62df18ce5f8d5c428e53203b9b975426eed0"
+
+
+@pytest.fixture
+def default_rulebook() -> models.Rulebook:
```
Alex-Izquierdo, line 25: What if all these fixtures are located in a conftest.py for reusability?

Alex-Izquierdo (commented):
why not just make the mixin returning directly the ContainerRequest?

tests/integration/services/activation/test_activation.py
jshimkus-rh: Created conftest.py for common fixtures.

jshimkus-rh:
@Alex-Izquierdo 
The mixin cannot create a ContainerRequest as part of the URL for ws is an attribute of ActivationManager.

Alex-Izquierdo (commented):
IMO, there is a lot of unnecessary code, including implementing getters for fields that don't change across models, as well as unnecessary steps like composing dicts to build the container request when it could be built directly. I think we don't need such an early abstraction, and I feel it's a bit of overengineering. The mixin could just compose and return the objects, and we can implement getters only for the properties that are different between activations and event streams. We can add them incrementally, as we need them.

I don't want to block the PR. If other people disagree and are fine with this approach, it's fine by me.

Alex-Izquierdo (dismissed):

Alex-Izquierdo (approved):

mkanoor (dismissed):

src/aap_eda/services/activation/engine/common.py
```diff
@@ -102,8 +110,8 @@ class ContainerRequest(BaseModel):
     name: str  # f"eda-{activation_instance.id}-{uuid.uuid4()}"
     image_url: str  # quay.io/ansible/ansible-rulebook:main
     cmdline: AnsibleRulebookCmdLine
-    activation_instance_id: str
-    activation_id: str
+    rulebook_process_id: str
+    resource_id: str
```
bzwei, line 114: `resource_id` is too generic, it has no clue the true meaning. In model RulebookProcess we call it parent or parent_type. In model ActivationRequestQueue we call it process_parent_id. To be consistent we can call it `process_parent_id` too here, because it refers to the same thing.

bzwei: Both id types are `int`, not `str`.

src/aap_eda/services/auth.py
```diff
@@ -16,7 +16,7 @@
 
 from rest_framework_simplejwt.tokens import RefreshToken
 
-from aap_eda.core.models import User
+from aap_eda.core.models.user import User
```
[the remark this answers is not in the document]
jshimkus-rh: It was changed to avoid a circular import issue involving `models/__init__.py` (`Activation -> ContainerableMixin -> create_jwt_token -> User`.
It's not ideal but, short-term, it solves the issue without having to perform a more involved location restructuring.

src/aap_eda/services/activation/engine/common.py
jshimkus-rh: Nice catch.  It was wrong even before me. 😉

jshimkus-rh:
The e2e failure results from the eda-qa tests not yet having been updated (there is a PR: [https://github.com/ansible/eda-qa/pull/298]) for the removal of ruleset endpoints.

src/aap_eda/services/activation/engine/common.py
```diff
@@ -113,6 +121,162 @@ class ContainerRequest(BaseModel):
...
+            "mem_limit": settings.PODMAN_MEM_LIMIT,
+            "mounts": settings.PODMAN_MOUNTS,
+            "process_parent_id": self.id,
+            "rulebook_process_id": self.latest_instance.id,
+            "cmdline": self._build_cmdline(),
+        }
+
+    def get_container_request(self) -> ContainerRequest:
+        """Return ContainerRequest used for creation."""
+        params = self.get_container_parameters()
```
Alex-Izquierdo, line 187: What is the purpose of a method to build a dict to later call it here and compose the ContainerRequest from that dict when you could just build the ContainerRequest, directly? It looks like a redundant step that also impact to tests. Same for _build_cmdline function.

Alex-Izquierdo (approved):

src/aap_eda/services/activation/engine/common.py
jshimkus-rh: AAP-20044 specifies the get* methods as part of the interface. Once those exist they act as the single source of truth for the respective values.

bzwei (approved):

src/aap_eda/services/activation/engine/common.py
Alex-Izquierdo: Is maybe the ticket outdated/incorrect? @mkanoor I think doesn't make sense to have a get_container_parameters with a dict when all is contained in the dataclass ContainerRequest. That dataclass is the interface for their consumers, otherwise we would have it duplicated.  I would say only the ContainerRequest is needed. Problably not even we need to get the CmdLine alone as it is part of the ContainerRequest itself.

jshimkus-rh: Oh. That was your point? The way it was worded came across, to me, as a critique of not just directly accessing the attributes when creating a ContainerRequest (which would be non-stellar software engineering). 

What you explained here makes more sense.

jshimkus-rh: Resolving solely to get the code in. I’ll pursue this after returning from PTO. Feel free to discuss whether Alejandro’s suggestion meets everyone’s use cases and, if so, make the necessary changes, if desired.

PR #673, opened 2024-02-13
[WIP] [RBAC] Add views for organizations and teams

Dostonbek1:
* add views for organizations and teams
* add org id field to all resource serializers - pending
* add tests - pending


src/aap_eda/api/filters/team.py
```diff
@@ -0,0 +1,29 @@
...
+class TeamFilter(django_filters.FilterSet):
+    name = django_filters.CharFilter(
+        field_name="name",
+        lookup_expr="istartswith",
+        label="Filter by team name.",
+    )
+
+    class Meta:
+        model = Team
+        fields = ["name"]
```
AlanCoding, line 29: Can you easily add a filter for `organization` as a related field? The would be the obvious thing to add.

Also, maybe description would be desired to be filterable? Both cases, for org and team.

AlanCoding (approved):
It's very helpful to have this structure in place to finish the work. I don't see anything that's even a little bit surprising here.

src/aap_eda/api/filters/team.py
Dostonbek1: Would it be ok to do filter by organization_id or do we wanna do it by organization_name? In the team list view, we just return the organization_id, not the whole object.

AlanCoding: I would just leave it at the organization_id for now. It gives the UI an option if they need it.

PR #675, opened 2024-02-13
fix: avoid race condition in max running processes

Alex-Izquierdo:
Having set the max_running_activations to 1 in the environment, I create several activations. As expected, as I create the activations, the first one starts running an the rest remain in pending state. However, if after that I delete the one that is running, then the rest of the pending rulebook activations all start running at the same time (ignoring the max_running_activations parameter)
Jira: https://issues.redhat.com/browse/AAP-20719


src/aap_eda/services/activation/manager.py
```diff
@@ -1041,6 +1041,17 @@ def _create_activation_instance(self):
             if hasattr(self.db_instance, "git_hash")
             else ""
         )
+        with transaction.atomic():
```
bzwei, line 1044: We should place line 1063 https://github.com/ansible/eda-server/pull/675/files#diff-767f95d0b4918e197417276be62147629e68416233ba137c48b8c171b570f73cR1063 inside the transaction block.

tests/integration/tasks/test_orchestrator.py
```diff
@@ -194,3 +223,41 @@ def test_monitor_rulebook_processes(
...
+    start_mock.side_effect = side_effect
+
+    max_running_processes[0].delete()
+
+    queue.push(
+        ProcessParentType.ACTIVATION, activation.id, ActivationRequest.START
+    )
+    orchestrator._manage(ProcessParentType.ACTIVATION, activation.id)
+    assert start_mock.call_count == 1
+    running_processes = models.RulebookProcess.objects.filter(
```
bzwei, line 260: use `filter().count()`

src/aap_eda/services/activation/manager.py
Alex-Izquierdo: added.

tests/integration/tasks/test_orchestrator.py
Alex-Izquierdo: changed

bzwei (dismissed):

mkanoor (approved):

Alex-Izquierdo:
Needed rebase. @mkanoor @bzwei please, review again.

bzwei (approved):

hsong-rh (approved):

PR #676, opened 2024-02-13
chore: allow for minikube deployments to use quay images

mkanoor:
Previously the minikube workflow required the UI and eda server images to be built locally before they can be used. Which is cumbersome, this PR adds a new task called
 * quay-deploy

which allows us to run directly from quay images and it also allows for the images to be overridden for UI and eda server local images using the following env vars

* EDA_UI_IMAGE=localhost/eda-ui:latest
* EDA_UI_IMAGE_PULL_POLICY=Never
* EDA_IMAGE=localhost/eda-server:latest
* EDA_IMAGE_PULL_POLICY=Never

To run the minikube deployment run the following command
```
    task minikube:quay-deploy
```
 Once all the pods are up and running do a port forwarding

```
task minikube:fp:ui 
```
Now you should be able to access the running containers using the following URL **https://localhost:8443/**

To cleanup
```
task minikube:clean
```
Given that we might be switching to use the ansible-playbook deployment this is a short term solution.

### DIFF
--- a/scripts/eda_kube.sh
+++ b/scripts/eda_kube.sh
@@ -22,6 +22,11 @@ source "${SCRIPTS_DIR}"/common/utils.sh
 
 trap handle_errors ERR
 
+export EDA_IMAGE=${EDA_IMAGE:-'quay.io/ansible/eda-server:main'}
+export EDA_UI_IMAGE=${EDA_UI_IMAGE:-'quay.io/ansible/eda-ui:main'}
+export EDA_IMAGE_PULL_POLICY=${EDA_IMAGE_PULL_POLICY:-'Always'}
+export EDA_UI_IMAGE_PULL_POLICY=${EDA_UI_IMAGE_PULL_POLICY:-'Always'}
+
 handle_errors() {
   log-err "An error occurred on or around line ${BASH_LINENO[0]}. Unable to continue."
   exit 1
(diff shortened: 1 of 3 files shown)


scripts/eda_kube.sh
```diff
@@ -22,6 +22,11 @@ source "${SCRIPTS_DIR}"/common/utils.sh
 
 trap handle_errors ERR
 
+export EDA_IMAGE=${EDA_IMAGE:-'quay.io/ansible/eda-server:main'}
+export EDA_UI_IMAGE=${EDA_UI_IMAGE:-'quay.io/ansible/eda-ui:main'}
+export EDA_IMAGE_PULL_POLICY=${EDA_IMAGE_PULL_POLICY:-'Always'}
+export EDA_UI_IMAGE_PULL_POLICY=${EDA_UI_IMAGE_PULL_POLICY:-'Always'}
```
jshimkus-rh, line 28: Conceivably these could be made to automatically set `Never` if no specific value was set and the associated image starts with `localhost`.

mkanoor: @jshimkus-rh when I do build using 
```
task minikube:build:api
```
The image is called 
```
docker.io/library/aap-eda:latest
```

jshimkus-rh: Never mind, then.

jshimkus-rh (approved):

hsong-rh (approved):

PR #678, opened 2024-02-14
chore: enforce version constraints

Alex-Izquierdo:
Set correct version constraints and update poetry.lock accordingly.  Set by default only upgrades for PATCH version. 
Jira: https://issues.redhat.com/browse/AAP-20786


Alex-Izquierdo:
Discarded after more discussions. The current constraints are based on older version, for the most recent it will be different.

PR #680, opened 2024-02-14
feat: Enable to encrypt and decrypt a string by ansible-vault

hsong-rh:
https://issues.redhat.com/browse/AAP-19426

### DIFF
--- a/src/aap_eda/api/vault.py
+++ b/src/aap_eda/api/vault.py
@@ -11,6 +11,7 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+import os
 import shutil
 import tempfile
 
@@ -36,6 +37,7 @@ if VAULT_COMMAND is None:
 
 def encrypt_string(password: str, plaintext: str, vault_id: str) -> str:
     tmp = tempfile.NamedTemporaryFile("w+t")
+    os.chmod(tmp.name, 0o600)
     tmp.write(password)
     tmp.flush()
     label = f"{vault_id}@{tmp.name}"
--- a/src/aap_eda/api/vault.py
+++ b/src/aap_eda/api/vault.py
@@ -1,0 +1,72 @@
+#  Copyright 2024 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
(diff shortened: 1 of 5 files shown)


hsong-rh:
@Alex-Izquierdo `ansible-vault` is found, even though I didn't change CI workflow. Is this because it's added in Dockerfile image?

Alex-Izquierdo:
@hsong-rh ansible-core stills be in the project dependencies. It is going to be removed in https://github.com/ansible/eda-server/pull/679

Alex-Izquierdo:
Anyway, even we remove i, it could work on CI, because the worker could have installed ansible (as well as it has a lot of common applications for CI) But even in that case, It should be present in the workflow, just to make it explicit.

src/aap_eda/api/vault.py
```diff
@@ -0,0 +1,82 @@
...
+class AnsibleVaultEncryptionFailed(Exception):
+    pass
+
+
+class AnsibleVaultDecryptionFailed(Exception):
+    pass
+
+
+def encrypt_string(password: str, plaintext: str, vault_id: str) -> str:
+    if not shutil.which("ansible-vault"):
```
Alex-Izquierdo, line 33: I think we want to raise the error at startup as we do with git: 
https://github.com/ansible/eda-server/blob/main/src/aap_eda/services/project/git.py#L47

The concern of @benthomasson is to not use ansible as library, but even through the CLI, the presence of ansible-vault is still a requirement of the application.

src/aap_eda/api/vault.py
```diff
@@ -0,0 +1,82 @@
...
+    child.sendcontrol("D")
+    i = child.expect(["Decryption successful", "ERROR"])
+    if i == 0:
+        child.readline()
+        data = child.readline()
+        decrypted_text = ""
+        while data:
+            decrypted_text += data.decode()
+            data = child.readline()
+        return decrypted_text.strip()
```
Alex-Izquierdo, line 79:
```suggestion
        decrypted_text = "".join(line.decode() for line in child)
        return decrypted_text.strip()
```

src/aap_eda/api/vault.py
```diff
@@ -0,0 +1,82 @@
...
+    i = child.expect(["Encryption successful", "ERROR"])
+    if i == 0:
+        child.readline()
+        data = child.readline()
+        encrypted_text = ""
+        while data:
+            enc_str = data.decode()
+            if not enc_str.startswith("!vault"):
+                encrypted_text += enc_str.lstrip()
+            data = child.readline()
```
Alex-Izquierdo, line 54:
```suggestion
        encrypted_text = "".join(line.decode() for line in child if not line.startswith("!vault"))
```

src/aap_eda/api/vault.py
```diff
@@ -0,0 +1,72 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+import shutil
+import tempfile
+
+import pexpect
+
+
+class AnsibleVaultNotFound(Exception):
+    pass
```
Alex-Izquierdo, line 21: Should the exception have a clear message like with do with git:
https://github.com/ansible/eda-server/blob/main/src/aap_eda/services/project/git.py#L49

Alex-Izquierdo (dismissed):

bzwei (approved):

mkanoor (approved):

PR #683, opened 2024-02-16
feat: optionally save encrypted data to extra vars using system_vault_credentials

hsong-rh:
https://issues.redhat.com/browse/AAP-20874

Depends on https://github.com/ansible/ansible-rulebook/pull/651

### DIFF
--- a/src/aap_eda/api/vault.py
+++ b/src/aap_eda/api/vault.py
@@ -46,17 +46,32 @@ def encrypt_string(password: str, plaintext: str, vault_id: str) -> str:
     child.expect("Reading plaintext input from stdin*")
     child.sendline(plaintext)
     child.sendcontrol("D")
-    i = child.expect(["Encryption successful", "ERROR"])
-    if i == 0:
-        child.readline()
-        return "".join(
-            line.decode().lstrip()
-            for line in child
-            if not line.decode().startswith("!vault")
-        )
+
+    encrypted_successful_found = False
+    encrypted_start = False
+    encrypted_str = ""
+
+    # Different versions of ansible-core output the "Encryption Successful"
+    # either at the beginning or at the end, right after the data.
+    # This logic accounts for all the variants we have seen so far in the
+    # output of ansible-vault.
+
+    for line in child.readlines():
+        out = line.decode().lstrip()
+        if out.startswith("!vault"):
+            encrypted_start = True
+            continue
+
+        if "Encryption successful" in out:
+            encrypted_successful_found = True
+
+        if encrypted_start:
+            encrypted_str += out.split("Encryption successful")[0]
+
+    if encrypted_successful_found and encrypted_start:
+        return encrypted_str
     else:
-        error_msg = child.readline()
-        raise AnsibleVaultEncryptionFailed(error_msg)
+        raise AnsibleVaultEncryptionFailed("Failed to encrypt string")
 
 
 def decrypt(password: str, vault_string: str) -> str:
--- a/src/aap_eda/api/serializers/activation.py
+++ b/src/aap_eda/api/serializers/activation.py
@@ -12,12 +12,17 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 import logging
+import secrets
+import uuid
 from typing import Union
 
 import yaml
 from rest_framework import serializers
 
-from aap_eda.api.constants import PG_NOTIFY_TEMPLATE_RULEBOOK_DATA
+from aap_eda.api.constants import (
+    EDA_SERVER_VAULT_LABEL,
+    PG_NOTIFY_TEMPLATE_RULEBOOK_DATA,
+)
 from aap_eda.api.exceptions import InvalidEventStreamRulebook
 from aap_eda.api.serializers.credential import CredentialSerializer
 from aap_eda.api.serializers.decision_environment import (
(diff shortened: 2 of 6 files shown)


mkanoor:
This PR cannot be merged till ansible-rulebook is merged.

src/aap_eda/api/vault.py
```diff
@@ -46,17 +46,27 @@ def encrypt_string(password: str, plaintext: str, vault_id: str) -> str:
...
-    i = child.expect(["Encryption successful", "ERROR"])
-    if i == 0:
-        child.readline()
-        return "".join(
-            line.decode().lstrip()
-            for line in child
-            if not line.decode().startswith("!vault")
-        )
+
+    encrypted_successful_found = False
```
mkanoor, line 50: @hsong-rh can you give some examples of why this was needed with the different versions of ansible-core.

src/aap_eda/api/vault.py
```diff
@@ -46,17 +46,27 @@ def encrypt_string(password: str, plaintext: str, vault_id: str) -> str:
...
+    encrypted_start = False
+    encrypted_str = ""
+
+    for line in child.readlines():
```
Alex-Izquierdo, line 54: Can you explain the logic of this block? I don't understanding it.  According to my experience this would be the output of ansible-vault:
```
➜  ansible-vault encrypt_string --vault-id eda@test
Reading plaintext input from stdin. (ctrl-d to end input, twice if your content does not already have a newline)
hola
Encryption successful
!vault |
          $ANSIBLE_VAULT;1.2;AES256;eda
          32626239383264306562303631613137376633366434623035343238643462323036613765626162
          3932626337356539663933653036363162626665373039320a626139643331306564653231636563
          66643735663734386637323235313331323962336464366661373730333566373166316562326639
          3438366636633733330a303461303366313131393635363561623233353837356266636464666333
          3530
```

So it would be just as checking the return code and two lines before the encrypted text: "Encryption successful", and "!vault".

src/aap_eda/api/vault.py
Alex-Izquierdo: If it depends on versions of ansible-core we may want to enforce a version. We don't expect a version lower than 2.16

hsong-rh:
@Alex-Izquierdo @mkanoor 

The `ansible-core` was added in Dockerfile and installed by dnf. The version of installed `ansible-vault` in eda server image is: 2.14.14:
```
[eda@d4d661982e22 src]$ ansible-vault --version
ansible-vault [core 2.14.14]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/app/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  ansible collection location = /app/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible-vault
  python version = 3.9.18 (main, Jan  4 2024, 00:00:00) [GCC 11.4.1 20231218 (Red Hat 11.4.1-3)] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
```
Use this version of ansible-vault to encrypt string, the output format is as follow, which is different from the newer version:
```
[eda@d4d661982e22 src]$ ansible-vault encrypt_string --vault-id EDA@/tmp/password
Reading plaintext input from stdin. (ctrl-d to end input, twice if your content does not already have a newline)
this is a test
!vault |
          $ANSIBLE_VAULT;1.2;AES256;EDA
          64363033363031666531343937306663303331313563663531376663666233393136616531396661
          6535633939613061663232613938356165313432643434390a636263323639363533613663393630
          35373838633139316263663739363533333638353939323961653764643331313066363538343534
          6361633938386366360a373830373666636263633962303233363732613633646462336361663138
          6265Encryption successful
```
The status message `Encryption successful` is moved to the end and mixed with the encrypted output data. The modification in this PR is to support both older and newer output format from `ansible-vault`.

Alex-Izquierdo:
In that case I think we should ensure that we install ansible-core 2.16 (with pip instead of dnf) regardless if we are going to enforce the version constraint.

src/aap_eda/api/vault.py
Alex-Izquierdo: What if we apply different logic for 2.14 or 2.16 instead? Not even I'm sure if we should support 2.14, since 2.16 will be the version of ansible for aap 2.5

mkanoor: The new logic works for the 2 different versions of ansible-core which I feel is better than the old logic which was expecting the "Encryption successful" before the encrypted data.

Alex-Izquierdo: I meant encapsulate them to see easily which is the logic for 2.14 and which for 2.16. Or at least add a comment indicating it.

mkanoor:
@Alex-Izquierdo @bzwei We need this reviewed and merged for the Credential Spike can you please review it. The current logic is agnostic to the version of ansible-core and it will handle the "Encryption Successful" coming at the end or in the beginning.

bzwei (approved):

src/aap_eda/api/vault.py
hsong-rh: @Alex-Izquierdo Comments added

bzwei (approved):

PR #686, opened 2024-02-20
fix: pass jwt lifetime seconds as seconds

Alex-Izquierdo:
JWT_ACCESS_TOKEN_LIFETIME_SECONDS is passed as minutes not seconds.


bzwei (approved):

src/aap_eda/settings/default.py
```diff
@@ -133,14 +133,16 @@ def _get_secret_key() -> str:
 
 # JWT token lifetime
 JWT_ACCESS_TOKEN_LIFETIME_SECONDS = settings.get(
-    "JWT_ACCESS_TOKEN_LIFETIME_SECONDS", 5
```
bzwei, line 136: I think the task:format will do this one line format instead of 2 lines.

src/aap_eda/settings/default.py
```diff
@@ -133,14 +133,16 @@ def _get_secret_key() -> str:
...
 )
 JWT_REFRESH_TOKEN_LIFETIME_DAYS = settings.get(
-    "JWT_REFRESH_TOKEN_LIFETIME_DAYS", 2
+    "JWT_REFRESH_TOKEN_LIFETIME_DAYS",
+    2,
 )
 SIMPLE_JWT = {
     "ACCESS_TOKEN_LIFETIME": timedelta(
-        minutes=JWT_ACCESS_TOKEN_LIFETIME_SECONDS
+        seconds=JWT_ACCESS_TOKEN_LIFETIME_SECONDS,
```
bzwei, line 145: The original thought is actually `minutes=JWT_ACCESS_TOKEN_LIFETIME_MINUTES`

src/aap_eda/settings/default.py
Alex-Izquierdo: it depends on the final comma. My env with rust -all, always suggests the trailing comma but we are not enforcing it in CI.

src/aap_eda/settings/default.py
Alex-Izquierdo: changed.

bzwei (approved):

hsong-rh (approved):

PR #691, opened 2024-02-22
Automated release

benthomasson:
Adds an automated release process using ansible and github actions.

Run locally with:
```
ansible-playbook tools/ansible/release.yml -i localhost -e github_token=${GH_TOKEN}
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,39 @@
+---
+name: Release eda-server
+
+env:
+  LC_ALL: "C.UTF-8" # prevent ERROR: Ansible could not initialize the preferred locale: unsupported locale setting
+
+on:
+  workflow_dispatch:
+
+jobs:
+  stage:
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    permissions:
+      packages: write
+      contents: write
+    steps:
+      - name: Install go-task
+        run: |
+          curl -sL https://taskfile.dev/install.sh | sh
+          mv ./bin/task /usr/local/bin/task
+          task --version
+
+      - name: Checkout eda-server
+        uses: actions/checkout@v3
+
+      - name: Get python version from Taskfile
+        run: echo py_version=`task PYTHON_VERSION` >> $GITHUB_ENV
+
+      - name: Install python ${{ env.py_version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ env.py_version }}
+
+      - name: Install python deeps
+        run: pip install -r requirements_dev.txt
+
+      - name: Create release
+        run: ansible-playbook tools/ansible/release.yml -i localhost -e github_token=${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  stage:
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    permissions:
+      packages: write
+      contents: write
+    steps:
+      - name: Checkout dab
```
Dostonbek1, line 18: same comment on "dab" as in ansible-rulebook

Dostonbek1 (approved):

Dostonbek1 (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,33 @@
...
+    timeout-minutes: 90
+    permissions:
+      packages: write
+      contents: write
+    steps:
+      - name: Checkout eda-server
+        uses: actions/checkout@v3
+
+      - name: Get python version from Makefile
+        run: echo py_version=`make PYTHON_VERSION` >> $GITHUB_ENV
```
Alex-Izquierdo, line 22: This line would not going to fail? I don't see any PYTHON_VERSION in the makefile.

benthomasson: I am still trying to find a way to test this.

Alex-Izquierdo:
In this project we are using `task` I think there is no need to duplicate tools adding a makefile. 
https://github.com/ansible/eda-server/blob/main/Taskfile.dist.yaml

.github/workflows/release.yml
benthomasson: I added this make target

benthomasson:
> In this project we are using `task` I think there is no need to duplicate tools adding a makefile. https://github.com/ansible/eda-server/blob/main/Taskfile.dist.yaml

Task is not available everywhere and is actually hard to install on some older versions of Fedora.     We should consider using makefile since it is the standard on the rest of the platform.

benthomasson:
I removed the Makefile and used task instead.

Taskfile.dist.yaml
```diff
@@ -355,3 +355,16 @@ tasks:
       - task: macpodman
         vars:
           CLI_ARGS: tunnel
+
+  build:release:
+    desc: "builds all the files needed for a release"
+    cmds:
```
Alex-Izquierdo, line 361: Is missing the build command?

pyproject.toml
```diff
@@ -120,3 +120,25 @@ ban-relative-imports = "parents"
 
 [tool.ruff.pydocstyle]
 convention = "pep257"
+
+[tool.setuptools_scm]
```
Alex-Izquierdo, line 124: should we use poetry for building as it is configured? 
https://github.com/ansible/eda-server/blob/main/pyproject.toml#L3

benthomasson:
Consumers of eda-server should use the latest build from quay.io

PR #692, opened 2024-02-22
chore: Change AuditEvent's payload from json into yaml

hsong-rh:
https://issues.redhat.com/browse/AAP-21045


Alex-Izquierdo (approved):
LGTM, but I would place YAMLSerializerField in serializers.fields not in utils.

hsong-rh:
> LGTM, but I would place YAMLSerializerField in serializers.fields not in utils.

Yeah, I have struggled with where to put the definition of YAMLSerializerField, both EventStream and Rulebook use it.

PR #695, opened 2024-02-27
Add permission inheritance for managed roles

Dostonbek1:
This adds changes to managed roles in RBAC created by `create_initial_data` command to grant parent models to have permissions for their associated child models.
 
The models permission inheritance structure that are changed:

1. Activation -> Activation Instance -> Audit Rule
2. Project -> Rulebook

### DIFF
--- a/src/aap_eda/core/management/commands/create_initial_data.py
+++ b/src/aap_eda/core/management/commands/create_initial_data.py
@@ -151,11 +151,6 @@ ORG_ROLES = [
 
 class Command(BaseCommand):
     help = "Seed database with initial roles."
-    parent_child_models = {
-        "project": "rulebook",
-        "activation": "rulebookprocess",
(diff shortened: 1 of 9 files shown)


src/aap_eda/core/management/commands/create_initial_data.py
```diff
@@ -151,6 +151,11 @@
 
 class Command(BaseCommand):
     help = "Seed database with initial roles."
+    parent_child_models = {
+        "project": "rulebook",
+        "activation": "rulebookprocess",
+        "activationinstance": "auditrule",
```
AlanCoding, line 157: AuditRule is registered as a normal model with organization as its parent. Doesn't jive with this line.

AlanCoding (approved):
:+1: this will be easier to debug and work out the issues than having editable organization fields on these models

PR #696, opened 2024-02-27
Define YAMLSerializerField in its own file

hsong-rh:
Move the definition of YAMLSerializerField to a separate field file.


mkanoor (approved):

Dostonbek1 (approved):

Alex-Izquierdo:
The new "fields" folder must have a __init__.py file otherwise it would not be considered a package, so it might not be included at build time.

PR #697, opened 2024-02-28
fix: remove tracebacks from backend logs when pulling image is timed out

hsong-rh:
https://issues.redhat.com/browse/AAP-20968

### DIFF
--- a/src/aap_eda/services/activation/engine/podman.py
+++ b/src/aap_eda/services/activation/engine/podman.py
@@ -21,6 +21,7 @@ from podman import PodmanClient
 from podman.domain.images import Image
 from podman.errors import ContainerError, ImageNotFound
 from podman.errors.exceptions import APIError, NotFound
+from rq.timeouts import JobTimeoutException
 
 from aap_eda.core.enums import ActivationStatus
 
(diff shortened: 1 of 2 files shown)


mkanoor:
@hsong-rh Do you have a screen shot of how this would look like in the UI?

Alex-Izquierdo (commented):
Have you really tested it in a real environment? I would say the exception is raised by the RQ worker, not by the podman client, so catching the exception inside the code that is executed would not have any effect. I would say the logic to handle JobTimeoutException must be outside the job, at the orchestrator level.

hsong-rh:
@mkanoor With the fix, the UI displays as the screenshot when the job is timed out (Changed timeout to 5 seconds for testing purposes). 
image

hsong-rh:
@Alex-Izquierdo The backend logs are like:
```
docker-eda-activation-worker-2  | 2024-03-05 15:04:08,225 aap_eda.services.activation.engine.podman INFO     Image URL is quay.io/ansible/ansible-rulebook:main
docker-eda-activation-worker-2  | 2024-03-05 15:04:08,240 aap_eda.services.activation.engine.podman INFO     Pulling image : quay.io/ansible/ansible-rulebook:main
docker-eda-activation-worker-2  | 2024-03-05 15:04:12,866 aap_eda.services.activation.engine.podman ERROR    Timeout: Task exceeded maximum timeout value (5 seconds)
docker-eda-activation-worker-2  | 2024-03-05 15:04:12,879 aap_eda.services.activation.manager INFO     Failed policy called for activation id: 1
docker-eda-activation-worker-2  | 2024-03-05 15:04:12,879 aap_eda.services.activation.manager INFO     Activation id: 1 Restart policy is set to never.
docker-eda-activation-worker-2  | 2024-03-05 15:04:12,886 aap_eda.services.activation.manager INFO     Cleanup operation requested for activation id: 1
docker-eda-activation-worker-2  | 2024-03-05 15:04:12,886 aap_eda.services.activation.manager INFO     Cleanup operation activation id: 1 No instance or pod id found.
```
The timeout exception is caught in the real scenario. I think it's better to handle it ASAP.

Alex-Izquierdo (approved):
I don't understand how it works if the exception is raised by the worker but if it works, LGTM.

mkanoor (approved):

PR #698, opened 2024-03-01
wip: generic scm credential api changes

jshimkus-rh:
I was already down the path of adding the four scm-related fields to the api so I finished up the basics.  It gives a feel of what things could be like.  We can certainly change to the more "blob" like approach that was discussed.

Unsurprisingly there are a number of tests that fail, but only 11 out of 398. 😄


src/aap_eda/api/serializers/credential.py
```diff
@@ -78,5 +86,9 @@ class Meta:
             "credential_type",
             "username",
             "vault_identifier",
+            "scm_username",
```
mkanoor, line 89: @jshimkus-rh can we use username and secret instead of scm_username and scm_password since they serve the same purpose.

mkanoor:
@jshimkus-rh We need to use validators for the private key akin to [this](https://github.com/ansible/awx/blob/b7b7bfa520780c481046581f2ce9c38ccc6bc1d8/awx/main/validators.py#L178 ) we have our validators for EDA defined [here](https://github.com/ansible/eda-server/blob/4f69f6a801bbc9ebd0505d4d208345f6613f271b/src/aap_eda/core/validators.py#L52).

src/aap_eda/api/serializers/credential.py
jshimkus-rh: Can do.

jshimkus-rh:
> We need to use validators for the private key 
That's an interesting question.  
We have other credentials (and other types) where validation is a side effect of attempted use.

tests/integration/api/test_project.py
```diff
@@ -481,6 +481,8 @@ def assert_project_data_details(data: Dict[str, Any], project: models.Project):
             "credential_type": credential.credential_type,
             "username": credential.username,
             "vault_identifier": None,
+            "scm_ssh_key": None,
+            "scm_ssh_key_password": None,
```
mkanoor, line 485: @jshimkus-rh Should we use passphrase instead of password since that is the well known term for private keys

jshimkus-rh:
Changed to no longer a wip.  Any additional changes (e.g., checking ssh key validity) are internal and don't impact the api usage.

src/aap_eda/api/serializers/credential.py
```diff
@@ -37,9 +37,20 @@ class Meta:
         ]
 
 
-class CredentialCreateSerializer(serializers.ModelSerializer):
-    secret = serializers.CharField(required=True, allow_null=False)
+def _validate_scm_ssh_key_params(ssh_key, ssh_key_passphrase):
+    """Both, or neither, of ssh key-related values must be specified."""
+    if ((ssh_key is not None) and (ssh_key != "")) and (
```
Alex-Izquierdo, line 42: passphrase is optional. You should be able to use a private key without passphrase.

src/aap_eda/services/project/git.py
```diff
@@ -150,6 +150,7 @@ def clone(
             cmd.extend(["--depth", str(depth)])
         cmd.extend([final_url, os.fspath(path)])
         logger.info("Cloning repository: %s", url)
+        logger.warning(f"JBS: Cloning command: {cmd}")
```
Alex-Izquierdo, line 153: What means JBS? We should try to avoid acronyms in logs, IMO.

Alex-Izquierdo, line 153: Have you check if this line will not leak potential secrets?

Alex-Izquierdo, line 153: If the use of the ssh key at import is out of the scope of this PR and it will be implemented later, I think this line is not needed at all.

PR #700, opened 2024-03-01
chore: add missing __init__.py

hsong-rh:
Follow up https://github.com/ansible/eda-server/pull/696, add missing __init__.py


hsong-rh:
The __init__.py is not required for packages.

PR #701, opened 2024-03-01
Enable podman multinode

Alex-Izquierdo:
Add new config to define a list of queues (one queue per activation node)
Add a new model to relate the queue with the process id
Add a dispatcher to balance activations across queues or route the job in the right queue. 
Checks for the state of the queue to not use it for new activations or set as "unknown" the existing ones. 

Jira: https://issues.redhat.com/browse/AAP-21311


src/aap_eda/settings/default.py
```diff
@@ -293,32 +312,74 @@ def _get_secret_key() -> str:
...
-if RQ_UNIX_SOCKET_PATH:
-    RQ_QUEUES = {
-        "default": {
-            "UNIX_SOCKET_PATH": RQ_UNIX_SOCKET_PATH,
-            "DEFAULT_TIMEOUT": 300,
-        },
-        "activation": {
+# A list of queues to be used in multinode mode
+# If the list is empty, the single node mode is used
+WORKERS_RULEBOOK_QUEUES = settings.get("WORKERS_RULEBOOK_QUEUES", [])
```
mkanoor, line 317: @Alex-Izquierdo RULEBOOK_WORKERS_QUEUES might be a better name

Alex-Izquierdo: changed.

src/aap_eda/tasks/orchestrator.py
```diff
@@ -121,43 +125,139 @@ def _run_request(
...
-    id: int,
-    request_type: ActivationRequest,
-) -> None:
-    """Enqueue a task to manage the activation with the given id."""
-    requests_queue.push(process_parent_type, id, request_type)
-    job_id = _manage_process_job_id(process_parent_type, id)
-    unique_enqueue("activation", job_id, _manage, process_parent_type, id)
+class QueueNotFoundError(Exception):
+    """Raised when a queue is not found."""
+
```
mkanoor, line 130: @Alex-Izquierdo Should we capture the queue name it might be helpful in debugging.

src/aap_eda/tasks/orchestrator.py
```diff
@@ -121,43 +134,195 @@ def _run_request(
...
+            except KeyError:
+                LOGGER.exception(f"Queue {queue_name} not found")
+                raise QueueNotFoundError(
+                    f"Queue {queue_name} not found"
+                ) from None
 
+        return queues
 
+    @staticmethod
+    def get_most_free_queue_name() -> str:
```
benthomasson, line 211: This function fails to load balance work across multiple workers if there are pending jobs.  It should count running and queued jobs for that worker.

benthomasson (changes requested):

src/aap_eda/tasks/orchestrator.py
benthomasson: Replace this function with random.choice

Alex-Izquierdo: I don't think I should count jobs, the goal is to be aware of the running processes, not the jobs, that can be any of type and usually are very quick. The goal is to distribute rulebook processes, not the jobs to manage them. 

If I use random.choice I could schedule a new rulebook process in a node that has no capacity to run more rulebooks, that would be inefficient and bad UX if there is another node with capacity.

benthomasson: Currently this heuristic is exactly backwards in many cases so anything else is better.

benthomasson: We should use a single queue for distributing jobs.    So we should have one queue for the rulebook process starts and one queue per worker for the monitoring tasks.

src/aap_eda/core/migrations/0027_rulebookprocessqueue.py
```diff
@@ -0,0 +1,52 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("core", "0026_activation_log_level_eventstream_log_level"),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name="RulebookProcessQueue",
+            fields=[
+                ("queue_name", models.CharField(max_length=255)),
```
mkanoor, line 28: @benthomasson This is the queue name that is trying to map the node where the process is running. In your PR #759 you are calling this worker, but the value is the hostname. When creating queues we could use hostnames as the queue name, which in this PR is a configurable parameter via settings. The installer could set the queue name as HOSTNAME. We could have multiple workers running on a node listening on the same queue_name.

benthomasson: #759 uses a single queue.   This PR implements a work distribution system instead of using a queue to distribute work.  It still has the bugs that I outlined above.  We should use a single queue to distribute activations to workers.

benthomasson (changes requested):
This PR fails to handle the case of a down node.   When a node is down then all the work will be assigned to the down node because of the incorrect counting of queue length.

src/aap_eda/core/migrations/0027_rulebookprocessqueue.py
Alex-Izquierdo: @mkanoor In fact, for the installer, I specified to use  `/etc/machine-id` which is a unique id per system, generated by the kernel. This is to avoid unexpected behaviors if the customer reuse the same node for a new installation (that will happen for sure) Ensuring in the installer that a new installation will be new queue, we can handle it easily.

src/aap_eda/tasks/orchestrator.py
```diff
@@ -121,43 +135,193 @@ def _run_request(
...
+    process_parent_id: int,
 ) -> None:
     """Create a request to start the activation with the given id."""
-    _make_user_request(process_parent_type, id, ActivationRequest.START)
+    requests_queue.push(
+        process_parent_type,
+        process_parent_id,
+        ActivationRequest.START,
+    )
+    RequestDispatcher.dispatch(
```
benthomasson, line 269: Why is the dispatcher needed?  The task is already pushed to the requests_queue?   Can the workers monitor the requests queue and filter it for their own tasks?

benthomasson (changes requested):

src/aap_eda/tasks/orchestrator.py
benthomasson: In general a node would know if a task is for them.   The same information is already used in the dispatcher, so why not put it in the requests_queue?

benthomasson: The only case where we don't know which node to run a task on is a new activation and that can go into an all node queue.

Alex-Izquierdo: The requests queue is only for user requests, we still have to run a management task for monitor running activation which is not an user request. The dispatcher is in charge to know what is the right queue as well as checks if the queue is healthy. 

This implementation attempts to extend the current design and make as few changes as possible. There is a pool of queues and the dispatcher is in charge of selecting the queue in all cases: when a new activation is created, when it is updated and when a user request is executed. 

The queue itself cannot use a balancing system that takes into account the load of the node (the activations that are currently running) as well as being able to discard queues that are considered unhealthy. 

I think it is important to understand the difference between a task to manage an activation and the management of the activation itself. 

eda-server does not manage jobs, jobs are managed by RQ redis. Eda-server manages activations and does it through RQ jobs. RQ queue distribute jobs across their workers. Eda-server distributes activations across queues.

src/aap_eda/settings/default.py
```diff
@@ -310,34 +329,91 @@ def _get_databases_settings() -> dict:
     "JOB_CLASS": "aap_eda.core.tasking.Job",
 }
 
+# This only takes effect if RULEBOOK_WORKER_QUEUES is not defined or
```
ttuffin, line 332: Why are these variables named **RULEBOOK_** and not **ACTIVATION_**?

Alex-Izquierdo: To follow the new name pattern introduced by the fan out, where we are not really dealing with activations but with rulebook processes.

Alex-Izquierdo: we still are using "activation" in multiple places and that's fine for now because the event streams are on hold, but it is partially implemented and for the rulebook_process orchestration are processes, no longer activations. So, until perform a complete rename of everything, for now I think the new addtions should follow the new naming, not the old one.

ttuffin (approved):

Alex-Izquierdo:
The fail in e2e test_max_running_activations is expected, it needs updates in the test suite as well in the workflow to support the new docker-compose with multinode

ttuffin (dismissed):

Alex-Izquierdo:
sucessful E2E tests: https://github.com/ansible/eda-server/actions/runs/8647611348/job/23709454923

src/aap_eda/core/tasking/__init__.py
```diff
@@ -158,7 +159,7 @@ def __init__(
         connection: Optional[Connection] = None,
         exc_handler: Any = None,
         exception_handlers: _ErrorHandlersArgType = None,
-        default_worker_ttl: int = DEFAULT_WORKER_TTL,
+        default_worker_ttl: int = 10,
```
mkanoor, line 162: @Alex-Izquierdo Should this be using settings.SOMETHING?

Alex-Izquierdo: Good catch. rolling back to the default one. It is overrided later in line 185 anyway. 
https://github.com/ansible/eda-server/pull/701/files#diff-8919c4e8183e21876fcf7b5ad8884b1efbb37a0e08544fba092d83c6f733adfbR185

src/aap_eda/core/migrations/0029_alter_activation_status_alter_eventstream_status_and_more.py
```diff
@@ -0,0 +1,114 @@
...
+
+def populate_rulebook_process_queue(apps, schema_editor):
+    RulebookProcessQueue = apps.get_model(  # noqa: N806
+        "core", "RulebookProcessQueue"
+    )
+    RulebookProcess = apps.get_model("core", "RulebookProcess")  # noqa: N806
+    for running_process in RulebookProcess.objects.filter(
+        status="RUNNING",
+    ).all():
+        RulebookProcessQueue.objects.create(
```
mkanoor, line 15: @Alex-Izquierdo what would happen post upgrade when there won't be a queue called activation since we now have per node queue name. Will these rulebooks get stuck because after the upgrade there won't be a queue called activation. As part of the upgrade do all activations have to be stopped and they will get restarted automatically when the restart policy is applied on it.

src/aap_eda/services/activation/manager.py
```diff
@@ -960,7 +964,16 @@ def monitor(self):
...
 
+        # last case is the success case
         if container_status.status == ActivationStatus.RUNNING:
+            if self.db_instance.status == ActivationStatus.WORKERS_OFFLINE:
+                LOGGER.info(
+                    f"Monitor operation: activation id: {self.db_instance.id} "
+                    "Activation is in WORKERS_OFFLINE state. "
+                    "Setting activation and instance status to running.",
+                )
+                self._set_activation_status(ActivationStatus.RUNNING)
```
mkanoor, line 975: @Alex-Izquierdo Is this true even if the node is powered off? There won't be anything running at that point, is this setting it to running so that the Heartbeat check can mark it as Unresponsive.

src/aap_eda/core/models/rulebook_process.py
```diff
@@ -196,3 +196,29 @@ class Meta:
...
     log = models.TextField()
     log_timestamp = models.BigIntegerField(null=False, default=0)
+
+
+class RulebookProcessQueue(models.Model):
+    """Rulebook Process Queue model.
+
+    Rulebook Process Queue is an instance of ansible-rulebook process
+    that is created when a rulebook process is started.
+    It keeps track of the queue name where the process is running.
```
jshimkus-rh, line 206: I find this description unclear.  Perhaps something along the line of... An association between a Queue and a RulebookProcess indicating the Queue on which all instances of RulebookProcess will execute.

src/aap_eda/core/tasking/__init__.py
```diff
@@ -172,14 +173,16 @@ def __init__(
         if queue_class is None:
             queue_class = Queue
 
+        queue_name = settings.RULEBOOK_QUEUE_NAME
```
jshimkus-rh, line 176: The queue name is not unique?  Perhaps only until future changes, but then a comment to that affect would be good.

src/aap_eda/core/tasking/__init__.py
```diff
@@ -204,14 +207,15 @@ def unique_enqueue(queue_name: str, job_id: str, *args, **kwargs) -> Job:
...
-        return job
+    for name in settings.RQ_QUEUES:
+        job = job_from_queue(name, job_id)
+        if job:
+            logger.info(
+                f"Skip enqueing job: {job_id} because it is already enqueued"
+            )
+            return job
     else:
+        queue = get_queue(name=queue_name)
```
jshimkus-rh, line 218: Doesn't this mean that a job could be sent to a different queue than the one on which it was previously run?  Perhaps only used when there is an issue with the job's original queue?  Could be that the queuing manipulation isn't fully implemented yet?  If that's the case a comment would be good.

src/aap_eda/services/activation/manager.py
```diff
@@ -1052,14 +1058,23 @@ def _create_activation_instance(self):
...
-            models.RulebookProcess.objects.create(**args)
+            rulebook_process = models.RulebookProcess.objects.create(**args)
         except IntegrityError as exc:
             msg = (
                 f"Activation {self.db_instance.id} failed to create "
                 f"activation instance. Reason: {exc}"
             )
             self._error_activation(msg)
             raise exceptions.ActivationStartError(msg) from exc
+        queue_name = self._get_queue_name()
```
jshimkus-rh, line 1069: Do we expect the queue name could change since line 1050?

src/aap_eda/tasks/orchestrator.py
```diff
@@ -32,23 +37,36 @@
...
 
 
+def get_process_parent(
+    process_parent_type: str,
+    parent_id: int,
+) -> Union[Activation, EventStream]:
+    if process_parent_type == ProcessParentType.ACTIVATION:
+        klass = Activation
+    else:
+        klass = EventStream
```
jshimkus-rh, line 58: Possible use for the python match statement with a default of "unknown type" exception. Yes, it was pre-existing but a check that a known type was passed would be better than making a default assumption.

src/aap_eda/core/models/rulebook_process.py
```diff
@@ -196,3 +196,29 @@ class Meta:
...
+    """
+
+    queue_name = models.CharField(max_length=255)
```
mkanoor, line 209: @Alex-Izquierdo Wouldn't this be better as a separate Node object that can have a relationship to one or more Rulebook Processes. The Node object can have a Queue Name/Host Name/Uptime/..... and can be updated separately and when a failure id detected on the Node flag all the running rulebook processes.

src/aap_eda/core/migrations/0029_alter_activation_status_alter_eventstream_status_and_more.py
Alex-Izquierdo: The installer will kill the activations if it is converting a node. When the pod is not running it is restarted (missing container policy)

src/aap_eda/core/models/rulebook_process.py
Alex-Izquierdo: Right now we are not managing nodes, is out of the scope, queues (nodes) are provided through static settings and handled at runtime. Node management can be done in the next release, it would require explicit tools for managing them (register/deregister comands) as well as dynamic queues.

src/aap_eda/core/tasking/__init__.py
Alex-Izquierdo: I'm not sure if I'm following you. RULEBOOK_QUEUE_NAME is set in the worker node. Each node should have a different name. Uniqueness is handled at the deployment time (we will use /etc/machine-id)

src/aap_eda/core/tasking/__init__.py
Alex-Izquierdo: The monitor logic will schedule a job in default queue before get enqueued in the right activation queue. The uniqueness must be guaranteed for all the queues. Also there are cases where an activation might change the queue if it is restarted manually and the node is unhealthy.

src/aap_eda/services/activation/manager.py
Alex-Izquierdo: At this point the node can not be powered off, this code is executed in the node and we are setting the activation as running because its container is running.

src/aap_eda/services/activation/manager.py
```diff
@@ -1034,16 +1047,9 @@ def _create_activation_instance(self):
...
-        ):
-            msg = (
-                "Failed to create rulebook process. "
-                "Reason: Max running processes reached. "
-                "Waiting for a free slot."
-            )
-            self._set_activation_status(ActivationStatus.PENDING, msg)
+        queue_name = self._get_queue_name()
+
+        if not self.check_new_process_allowed():
```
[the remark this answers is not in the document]
Alex-Izquierdo: Indeed, it is executed in two places, including in the orchestrator, to avoid race conditions. In the long term we can explore what you say.

src/aap_eda/services/activation/manager.py
Alex-Izquierdo: Nope. It is a mistake. Good catch. I removed the 1050.

src/aap_eda/tasks/orchestrator.py
Alex-Izquierdo: Well, this is a existing code, just now I encapsulate it in a function to be reused. IRCC when we started to implement event streams, we discussed about this, the point is that it is not an assumption, but I agree that adding such a exception does't make any difference. To me that would be out of the scope of this PR, let's do it in a different one, do you agree?

Alex-Izquierdo: Anyway, I added the exception. ;)

src/aap_eda/tasks/orchestrator.py
```diff
@@ -121,43 +147,240 @@ def _run_request(
...
+        # If the queue is old or doesn't exist, we use a valid one
+        # to make sure the request is processed and the restart
+        # policy is respected. Same if the request is restart and
+        # the queue is not healthy.
+        if not queue_name or queue_name not in settings.RULEBOOK_WORKER_QUEUES:
+            LOGGER.warning(
+                f"Queue {queue_name} doesn't exist or is not valid. "
+                f"Rescheduling {process_parent_type} {process_parent_id}"
+                f" to the most free queue.",
+            )
```
jshimkus-rh, line 187: If a deployment changes the queue names we shouldn't be logging a warning every time we encounter an old name.   In fact we shouldn't be logging anything about it during runtime as we'll still process the job.

A single warning at start up to indicate the situation should be sufficient.  We should also issue a warning for a smaller number of queues as that could have a performance impact.

Now...how do we handle that intelligently?  If I start with queues A and B, restart with queues C and D then restart again with C and D that second restart shouldn't generate a warning as nothing's changed.  Similarly for queue count changes.  Maybe we ...

1. clean out all RulebookProcessQueue entries for non-existent queues
2. update the queue and job relation when the job is re-run

The first gets rid of the problem in one shot.
The second cleans it up over time assuming all jobs get re-run, but spits out the warning at startup until then; always if even a single job is never re-run

Alex-Izquierdo: I think the warning message is relevant because that change might be expected or not, so to me is relevant to keep a record. Any other idea or improvement can be discussed in later iterations. 
BTW, there is no relation between a queue and a job, but between a queue and some rulebook_process. Every new process will have a new record with the queue.

jshimkus-rh: By 'job' I mean the higher conceptual aspect of every instance run for a particular activation/event_stream.
I have a  mental bias to read "X process" as a singular instance rather than the aggregate.  Anyhow...
for clarity when you say "new process" you mean the aggregate, correct?  Having one per instance would be redundant it would seem.

jshimkus-rh: > I think the warning message is relevant because that change might be expected or not, 

How could that occur at runtime?  Yes, the queue could develop a problem and such should be logged but the queue name is not going to disappear from `settings.RULEBOOK_WORKER_QUEUES`; it just won't be there from the beginning.  What's the value in continually logging that?

The lack of a queue name would be odd and would imply some previous error in tracking content.  So that occurrence would be reasonable to log, though mostly for our usage I would think.

Alex-Izquierdo: I don't understand this discussion, honestly. It is not continually logging anything, it is logged just one time. I expect a queue, but the queue is not configured in the system and I consider that it must be logged, because it can be on purpose or not and to me is enough exceptional to be warning. Once the activation is moved to a valid queue all is fine. 

It would be fine if I move it to INFO?

Alex-Izquierdo: I've changed it to INFO level

src/aap_eda/core/models/rulebook_process.py
jshimkus-rh: Better.  A couple of tweaks...

    Rulebook Process Queue keeps track of the queue name for a
    Rulebook Process. Every rulebook process is associated with the
    queue, by name, where it ran at creation.

Alex-Izquierdo: updated

jshimkus-rh (dismissed):

mkanoor (approved):

src/aap_eda/core/tasking/__init__.py
```diff
@@ -204,14 +207,15 @@ def unique_enqueue(queue_name: str, job_id: str, *args, **kwargs) -> Job:
...
-        )
-        return job
+    for name in settings.RQ_QUEUES:
+        job = job_from_queue(name, job_id)
+        if job:
+            logger.info(
+                f"Skip enqueing job: {job_id} because it is already enqueued"
+            )
+            return job
     else:
```
hsong-rh, line 217: Should this `else` be indented to match?

Alex-Izquierdo: @hsong-rh that else belongs to the for loop. It is true [it's not necessary](https://docs.astral.sh/ruff/rules/useless-else-on-loop/), but doesn't have an impact either.

hsong-rh (dismissed):

hsong-rh (approved):

mkanoor (approved):

PR #702, opened 2024-03-01
fix: enhance rulebook validation during project import

bzwei:
Validate db model required fields exist in the rulebook. Skip such rulebook and log a warning.

Also fix some selected exceptions prevent the import_state from changing to failed.
AAP-20868: Project import hangs when importing malformed rulebook

### DIFF
--- a/src/aap_eda/services/project/imports.py
+++ b/src/aap_eda/services/project/imports.py
@@ -22,7 +22,6 @@ from typing import Any, Callable, Final, Iterator, Optional, Type
 
 import yaml
 from django.core import exceptions
-from django.db import utils
 
 from aap_eda.core import models
 from aap_eda.core.types import StrPath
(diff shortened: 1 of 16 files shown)


src/aap_eda/services/project/imports.py
```diff
@@ -46,27 +45,39 @@ class ProjectImportError(Exception):
...
-            raise
+                raise ProjectImportError(
+                    "Project may have been deleted"
+                ) from error
+            else:
+                if error and isinstance(error, ProjectImportError):
+                    raise error
+                elif error:
+                    raise ProjectImportError(
+                        "Failed to import the project"
```
Alex-Izquierdo, line 79: This message might be captured by the upper try block and exposed to the user in the import_error field, so it would be a message without much meaning. I think it should include the original error message as well. What do you think?

src/aap_eda/services/project/imports.py
```diff
@@ -234,7 +248,22 @@ def _try_load_rulebook(
             content=content,
         )
 
-    def _is_rulebook_file(self, data: Any) -> bool:
+    def _validate_rulebook_file(self, data: Any) -> None:
```
Alex-Izquierdo, line 251: should we use the schema for validation? https://github.com/ansible/ansible-rulebook/blob/main/ansible_rulebook/schema/ruleset_schema.json

tests/integration/services/test_project_import.py
```diff
@@ -295,6 +296,50 @@ def clone_project(_url, path, *_args, **_kwargs):
...
+
+    git_mock = mock.Mock(name="GitRepository", spec=GitRepository)
+    git_mock.clone.side_effect = clone_project
+
+    project = models.Project.objects.create(
+        name="test-project-04", url="https://git.example.com/repo.git"
+    )
+
+    logger = logging.getLogger("aap_eda")
+    propogate = logger.propagate
```
Alex-Izquierdo, line 321: propogate is a typo?

hsong-rh (approved):

Issue #381, opened 2023-08-31
Question \ FR - Ability to assign static Pod\Service\Job names to RuleBook Activations (K8s)


Daniel-Vaz:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/aap-eda/issues) for duplicates.
- [X] I understand that EDA-Controller is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

Currently when a RuleBook Activation is created and enabled a default name is generated based on its internal ID and given to the created Pod, Service and Job Objects inside a K8s clusters. 

When there is a use-case where a user wants to have a central, long-lived and unique RuleBook Activation, it would be nice if it could have some control over the generated Service Object name. This would make it much easier to create default Ingress\Mappings to expose the RuleBook endpoint. 

Currently the RuleBook Activation needs to always be first manually created, and then can we see the generated name for the service so they can be later exposed outside of the cluster. 

### Steps to reproduce

Create a RuleBook Activation. 
See the generated Pod, Service and Job names. 

### Current results

Pod, Service and Job Names follow the following naming convention:

- Pod: `activation-job-<ID>-XXX`
- Service: `activation-job-<ID>-<PORT>`
- Job: `activation-job-<ID>`

### Sugested feature result

Add the ability the statically define the generated resources names (at least Service), in the RuleBook Activation page, and have that name be used for the generated Service Object instead.

Alex-Izquierdo:
RFE: https://issues.redhat.com/browse/AAPRFE-821

kurokobo:
> Pod: `activation-job-<ID>-XXX`
> Service: `activation-job-<ID>-<PORT>`
> Job: `activation-job-<ID>`

In the recent implementation, these names are updated to use 2 different IDs:

- Pod: `activation-job-<Activation ID>-<Instance ID>-XXX`
- Service: `activation-job-<Activation ID>-<Instance ID>-<PORT>`
- Job: `activation-job-<Activation ID>-<Instance ID>`

These names change simply by reactivating the Rulebook, which increases the difficulty of maintaining a stable connection when users want to create an Ingress as the frontend of the Service. I would like to see this RFE implemented.

rooftopcellist:
[...]

Issue #616, opened 2024-01-19
Rulebook restart within kubernetes causes rulebooks to become unavailable


jon-nfc:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/aap-eda/issues) for duplicates.
- [X] I understand that EDA-Controller is open source software provided for free and that I might not receive a timely response.

### Bug Summary

Restarting a rulebook activation has two issues:

- if using kubernetes ingress, you must re-deploy the ingress as the service name has changed

- cluster Internal traffic that uses the service name for rulebook access will also loose access to the rulebook as the address has changed, specifically in this case the service name.

### Environment

Kubernetes

### Steps to reproduce

1. start a rulebook activation
2. check service name `kubectl get svc -n <namespace>`
3. note name
4. restart rulebook activation
5. check service name `kubectl get svc -n <namespace>`
6. bam!!! service name different

### Actual results

Service details before restart

``` bash
root@7dd5f0bf32b32b93:/home/deploy# kubectl get svc -n eda
NAME                                                     TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
activation-job-6-16-5003                                 ClusterIP   172.16.41.108   <none>        5003/TCP   6m40s
eda-api                                                  ClusterIP   172.16.41.45    <none>        8000/TCP   3d
eda-daphne                                               ClusterIP   172.16.41.171   <none>        8001/TCP   3d
eda-postgres-13                                          ClusterIP   None            <none>        5432/TCP   3d
eda-redis-svc                                            ClusterIP   172.16.40.199   <none>        6379/TCP   3d
eda-server-operator-controller-manager-metrics-service   ClusterIP   172.16.40.153   <none>        8443/TCP   11d
eda-ui                                                   ClusterIP   172.16.41.34    <none>        80/TCP     3d
root@7dd5f0bf32b32b93:/home/deploy# 
root@7dd5f0bf32b32b93:/home/deploy# 
root@7dd5f0bf32b32b93:/home/deploy# kubectl get svc -n eda activation-job-6-16-5003 -o yaml
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2024-01-19T18:09:20Z"
  labels:
    app: eda
    job-name: activation-job-6-16
  name: activation-job-6-16-5003
  namespace: eda
[...]

PR #704, opened 2024-03-04
feat: enable to use static service name for activation in k8s

hsong-rh:
https://issues.redhat.com/browse/AAP-21464

This PR enables users to provide a service name for activations when the deployment type is k8s. It gets ignored if the deployment type is podman.

1. The provided service name has to follow the rules: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#rfc-1035-label-names;
2. If users don't specify a service name, we will attempt to auto generate a service name from the activation name and use it. If the name cannot be generated we would prompt the user to provide a service name. 

if the service name is not RFC1036 compliant, UI will display the errors like:
1. auto-generated service name using the activation name:
image
2. user inputs an illegal service name:
image

Follow https://github.com/ansible/eda-server/issues/381,


Alex-Izquierdo:
I think we should have a internal ticket ref to track this feature.

Alex-Izquierdo:
Closes https://github.com/ansible/eda-server/issues/616
Closes https://github.com/ansible/eda-server/issues/381

src/aap_eda/core/models/activation.py
```diff
@@ -112,3 +115,8 @@ class Meta:
         choices=RulebookProcessLogLevel.choices(),
         default=get_default_log_level,
     )
+    k8s_service_name = models.TextField(
+        null=False,
+        default=get_default_k8s_service_name,
```
Alex-Izquierdo, line 120: I think this field can not have a default value from system settings because it must be different per activation. What if we use the name of the activation which is already unique?  In the migration we can update the existing records and ensure the new ones are created with that value.

Alex-Izquierdo (commented):
We need a internal ticket to track this feature. That ticket would need more actions alongside this change. Documentation user-guide must be updated. 

Also we will need an endpoint to let know to the UI that the deployment type is K8s. The UI should not show this field with podman. We already have a ticket to implement a "config" endpoint as AWX does, so we will need to prioritize it and tied it to this feature.

hsong-rh:
> We need a internal ticket to track this feature. 

https://issues.redhat.com/browse/AAP-21464

src/aap_eda/core/models/activation.py
hsong-rh: Why must it be different per activation? the purpose here is to allow users to choose their services. We don't need migration for the old records.

Alex-Izquierdo: We are creating a service per activation selecting a job_name (which is the current rulebook process) if you use the same name you would overwrite the existing one, so you can have only one service for just one activation.

hsong-rh: @Alex-Izquierdo I didn't get your point. Here we want to allow the user to select their pre-defined services.  Every activation can use it. And the `job_name` is changed to separate from the service name.

src/aap_eda/core/utils/k8s_service_name.py
```diff
@@ -0,0 +1,94 @@
...
+    "horizon",
+    "marianas",
+    "sirena",
+    "milwaukee",
+    "molloy",
+    "atacama",
+]
+
+
+def create_k8s_service_name(activation: str) -> str:
```
mkanoor, line 76: @hsong-rh We don't need this since we are using service name from activation name.

src/aap_eda/api/serializers/activation.py
```diff
@@ -562,6 +585,11 @@ class PostActivationSerializer(serializers.ModelSerializer):
         allow_null=True,
         child=serializers.IntegerField(),
     )
+    k8s_service_name = serializers.CharField(
+        required=False,
+        allow_null=True,
+        validators=[validators.check_if_rcf_1035_compatible],
```
mkanoor, line 591: @hsong-rh this function might be better called check_if_rfc_1035_compliant. In the current name rfc is mistyped as rcf.

src/aap_eda/core/utils/k8s_service_name.py
```diff
@@ -0,0 +1,43 @@
...
+class InvalidRFC1035Label(Exception):
+    pass
+
+
+def create_k8s_service_name(activation_name: str) -> str:
+    """Random create a service name based on the activation name."""
+    name = re.sub(r"[._\s+]", "-", activation_name.lower())
+    if not rfc_1035_compatible(name):
+        raise InvalidRFC1035Label(
+            f"{activation_name} must be a valid RFC 1035 label name"
```
mkanoor, line 27: @hsong-rh I think this message might be better as
```
"Please provide a service name, that's a valid RFC 1035 name, couldn't auto generate service name from activation name"
```

src/aap_eda/core/utils/k8s_service_name.py
```diff
@@ -0,0 +1,43 @@
...
+        )
+
+    return name
+
+
+def rfc_1035_compatible(name: str) -> bool:
```
mkanoor, line 33: @hsong-rh I think a better name would be **rfc_1035_compliant**

src/aap_eda/core/utils/k8s_service_name.py
```diff
@@ -0,0 +1,43 @@
...
+
+import re
+
+
+class InvalidRFC1035Label(Exception):
+    pass
+
+
+def create_k8s_service_name(activation_name: str) -> str:
+    """Random create a service name based on the activation name."""
```
mkanoor, line 23: @hsong-rh this comment needs to be fixed its not a random name

tests/unit/test_k8s_service_name.py
```diff
@@ -0,0 +1,70 @@
+#  Copyright 2023 Red Hat, Inc.
```
mkanoor, line 1: @hsong-rh 2024

tests/unit/test_k8s_service_name.py
```diff
@@ -0,0 +1,70 @@
...
+import pytest
+
+from aap_eda.core.utils.k8s_service_name import (
+    InvalidRFC1035Label,
+    create_k8s_service_name,
+    rfc_1035_complicant,
+)
+
+
+def test_rfc_1035_complicant():
```
mkanoor, line 24: @hsong-rh Should we use a parameterized test for all successful and failed tests, if we had an array of strings it will be much easier to see all the ones that pass and ones that fail and add new ones down the road.

mkanoor (dismissed):

src/aap_eda/core/utils/k8s_service_name.py
```diff
@@ -0,0 +1,47 @@
...
+class InvalidRFC1035Label(Exception):
+    pass
+
+
+def create_k8s_service_name(activation_name: str) -> str:
+    """Attempt to convert the activation name to a 1035 name.
+
+    Convert dot(.) underscore(_) space( ) to - and convert it to lowercase
+    """
+    name = re.sub(r"[._\s+]", "-", activation_name.lower())
```
Alex-Izquierdo, line 27: Can I suggest an improvement in the regex to negate the allowed characters instead?
`r"[^a-z0-9-]", "-"`

src/aap_eda/core/utils/k8s_service_name.py
```diff
@@ -0,0 +1,47 @@
...
+
+def rfc_1035_complicant(name: str) -> bool:
+    """Check if name is compatible with RFC 1035.
+
+    - contain at most 63 characters
+    - contain only lowercase alphanumeric characters or '-'
+    - start with an alphabetic character
+    - end with an alphanumeric character
+    """
+    qualified_pattern = re.compile(r"^[a-z]([a-z-0-9-]{0,61}[a-z])?$")
```
Alex-Izquierdo, line 45: 1. The hyphen - is specified twice, is redundant but not harmful.
2. RFC 1035 allows labels to end with digits. 

I suggest: 
`qualified_pattern = re.compile(r"^[a-z]([a-z0-9-]{0,61}[a-z0-9])?$")`

src/aap_eda/core/utils/k8s_service_name.py
```diff
@@ -0,0 +1,47 @@
...
+    if not rfc_1035_complicant(name):
+        raise InvalidRFC1035Label(
+            "Please provide a service name, that's a valid RFC 1035 name, "
+            "couldn't auto generate service name from activation name"
+        )
+
+    return name
+
+
+def rfc_1035_complicant(name: str) -> bool:
```
Alex-Izquierdo, line 37: Can It be named `is_rfc_1035_compliant` to follow the convention about function names that returns a boolean? 
It has a typo: "complicant"

src/aap_eda/core/utils/k8s_service_name.py
```diff
@@ -0,0 +1,47 @@
...
+
+def create_k8s_service_name(activation_name: str) -> str:
+    """Attempt to convert the activation name to a 1035 name.
+
+    Convert dot(.) underscore(_) space( ) to - and convert it to lowercase
+    """
+    name = re.sub(r"[._\s+]", "-", activation_name.lower())
+    if not rfc_1035_complicant(name):
+        raise InvalidRFC1035Label(
+            "Please provide a service name, that's a valid RFC 1035 name, "
```
Alex-Izquierdo, line 30: This executed against the activation name, it should not include the word "service". The function would not called if the user already provides a service name.

src/aap_eda/core/utils/k8s_service_name.py
```diff
@@ -0,0 +1,47 @@
...
+            "couldn't auto generate service name from activation name"
```
Alex-Izquierdo, line 31: Right now we only support activations, but since we are supporting implicitily the event stream, I think it would be better if we just omit the "activation" reference. `couldn't auto generate a service name`

src/aap_eda/services/activation/engine/kubernetes.py
```diff
@@ -379,6 +393,15 @@ def _create_service(self, port: int) -> None:
...
             else:
                 LOGGER.info(f"Service already exists: {service_name}")
+                opened_ports = [
+                    port_item.port for port_item in service.items[0].spec.ports
+                ]
+                if bool(set(request_ports) - set(opened_ports)):
+                    raise ContainerStartError(
+                        f"Request ports {request_ports} is not opened in the "
+                        f"service {service_name} with ports: {opened_ports}"
+                    )
```
Alex-Izquierdo, line 403: If ports are plural, the verb should be "are", am I wrong? 
This is because the service already exists but the ports are different. Should we not recreate the service in that case?

src/aap_eda/core/utils/k8s_service_name.py
hsong-rh: I feel that replacement is too much. The auto-generated service name will be unreadable if the activation name has many replaced characters.

src/aap_eda/services/activation/engine/kubernetes.py
hsong-rh: > If ports are plural, the verb should be "are", am I wrong? 
yes, changed.

> This is because the service already exists but the ports are different. Should we not recreate the service in that case?
We are not clear about the situation of the user's input service. It's not a good idea to recreate it . We can open the needed ports in the service, but it's hard to do the cleanup after activation is completed. The simple way is let user know and make the change.

Alex-Izquierdo (approved):

mkanoor (approved):

PR #706, opened 2024-03-05
chore: remove redundant partial update mixin

jshimkus-rh:


Alex-Izquierdo (approved):

Dostonbek1 (approved):
lgtm. One minor comment, we might want to remove other files in this directory but it is a bit confusing to me why we have both `mixins` folder and file that basically have the same serializer classes.

jshimkus-rh:
> lgtm. One minor comment, we might want to remove other files in this directory but it is a bit confusing to me why we have both `mixins` folder and file that basically have the same serializer classes.

@Dostonbek1 A valid point.

Alex-Izquierdo (approved):

Dostonbek1 (approved):

PR #707, opened 2024-03-05
fix: allowlist for source plugins with port forward

mkanoor:
On the EDA Server side we try to open ports for port forwarding using the container engine (podman/k8s). We only allow a specific set of source plugins listed in the settings
**SAFE_PLUGINS_FOR_PORT_FORWARD** to be eligible for port forwarding. The default settings only supports the following source plugins

* ansible.eda.webhook
* ansible.eda.alertmanager

The user can augment this settings to include other source plugins

https://issues.redhat.com/browse/AAP-17416

### DIFF
--- a/src/aap_eda/services/activation/engine/ports.py
+++ b/src/aap_eda/services/activation/engine/ports.py
@@ -3,6 +3,7 @@ import logging
 
 import jinja2
 import yaml
+from django.conf import settings
 from jinja2.exceptions import UndefinedError
 from jinja2.nativetypes import NativeTemplate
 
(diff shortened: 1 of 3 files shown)


bzwei:
@mkanoor For users using other webhook like plugins but not ansible.eda.webhook, their activations will fail after upgrading eda-server unless they modify the settings for SAFE_PLUGINS_FOR_PORT_FORWARD. Is this a breaking change that cannot be included in a minor release?

The default should include ansible.eda.alertmanager which is provided in our collection.

benthomasson (approved):

PR #710, opened 2024-03-06
feat: implement RBAC 

Dostonbek1:
# Description

This PR introduces 

1. A new RBAC functionality for EDA based on [RBAC library](https://github.com/ansible/django-ansible-base/tree/devel/ansible_base/rbac) from django-ansible-base.
2. Organization and teams models/endpoints.
3. Optimization for integration tests using general fixtures.

JIRA Epic: [AAP-19859](https://issues.redhat.com/browse/AAP-19859)

UI PR: https://github.com/ansible/ansible-ui/pull/2036 & https://github.com/ansible/ansible-ui/pull/2071
QE PR: https://github.com/ansible/eda-qa/pull/313

NOTE: we'll cut a new release in DAB once all fixes for RBAC are in and pin to that release version before merging this PR.

# Background

This RBAC implementation introduces a new way of managing roles and permissions that mainly consists of organization level and object level access. The existing 6 managed roles are migrated to org-level roles and apply to all resources in that organization. In addition to these roles, we are also pre-creating object-level managed roles that will grant object-level access to a user or a team. Creation of custom roles is also introduced that will allow admin users to create new custom roles with a list of permissions of their choice as per their needs.

Since some resources in EDA have parent-child relation, the managed object-level roles will also reflect this relation. For example, a project is a parent model for rulebooks thus assigning a user `Project Admin` role will also grant them full access to the project's rulebook objects. `Project Admin` role looks something like this:
```
{
      "id": 13,
      "name": "Project Admin",
      "description": "Has all permissions to a single project and its child resources - rulebook",
      "content_type": "eda.project",
      "permissions": [
        "eda.change_project",
        "eda.delete_project",
        "eda.view_project",
        "eda.view_rulebook"
      ],
}
```
These parent-child relations are defined [here](https://github.com/ansible/eda-server/blob/26f9ab792496325e7a53985b70b8837c42c8d6e3/src/aap_eda/core/models/__init__.py#L77).

Full documentation for RBAC library from DAB can be found [here](https://github.com/ansible/django-ansible-base/blob/devel/docs/apps/rbac.md). 

## URLs

The following urls are imported from DAB RBAC library:

1. `/api/eda/v1/role_definitions/` - used to create/retrieve/list roles that has a list of permissions on different resources, which will replace the current `/roles` endpoints. The POST payload looks as follows:
```
{
    "permissions": ["view_project"],
    "content_type": "eda.project",
    "name": "Project Viewer",
    "description": "custom role"
}
```
2. `/api/eda/v1/role_user_assignments/` - used to assign an existing role to a user. The POST endpoint payload looks as follows
```
{
    "role_definition": 1,
    "object_id": 1,
    "user": 1
}
```
`role_definition` is the role that is to be assigned to the user. `object_id` is the ID of the object the role applies to, and `user` is the ID of the user that the role is being assigned to. 
4. `/api/eda/v1/role_team_assignments/` - used to assign an existing role to a team. This is very similar to `role_user_assignment` but instead of a user the team gets the role assigned and all the members of the team inherit the role as well as its permissions. Example POST payload
```
{
    "role_definition": 3,
    "object_id": 3,
    "team": 2
}
```

In addition, we also implemented custom `api/eda/v1/organizations/` and `api/eda/v1/teams/` endpoints that are used for organization and team operations accordingly.

NOTE: Organization and Team memberships are also managed by `role_user_assignment` using managed roles called `Organization Member` and `Team Member` that will have the following permissions, `["eda.member_organization", "eda.view_organization"]` and `["eda.member_team", "eda.view_team"]`, respectively.

## Data Migration

1. The currently existing 6 system roles in EDA will be migrated to organization level roles. In addition to this, we are also creating object-level roles per resource in `create_initial_data.py`, e.g. Activation Admin, Project Admin, etc., that can be used for object level permissions.
2. All the existing resources data will be moved to a Default Organization. 

# Testing

Follow the steps below to test an object-level access for resources, such as Project, Decision Environment and Activation.
1. Log in as admin user. The user with `is_superuser=True` flag still works and has system level access.
2. Create a new Project and a new Decision Environment, and start up a new Activation.
3. Create a new user for testing through the API (the UI might expect roles value, thus we need to use the API for now). The payload can be as simple as this:
```
{
  "username": "doston",
  "password": "doston"
}
```
4. Go to `/api/eda/v1/role_definitions/` and get the ID for `Activation Admin`, which we will use in next steps.
5. Make a POST request to `/api/eda/v1/role_user_assignments/` with the following payload; make sure to replace `role_definition` value with the ID from step 4. `object_id` is the ID of the Activation we created in step 2 and user is the test user `doston` created in step 3. (we can ignore `object_ansible_id` and `user_ansible_id` for now)
```
{
  "object_id": "1",
  "role_definition": 14,
  "user": 3
}
```
This should give the user `doston` full access to the Activation we created earlier as well as its child objects such as Activation Instances, Audit Rules, Audit Events and Audit Actions.
6. Log out and log in as the user `doston` and check to make sure the user has access to the Activation in question. 

More roles can be assigned to the user to have access to other objects like Project and Decision Environments by creating new role user assignments.


src/aap_eda/api/filters/organization.py
```diff
@@ -0,0 +1,34 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import django_filters
+
+from aap_eda.core.models import Organization
+
+
+class OrganizationFilter(django_filters.FilterSet):
+    name = django_filters.CharFilter(
```
mkanoor, line 21: @Dostonbek1 Should these be coming from DAB?

src/aap_eda/core/models/organization.py
```diff
@@ -0,0 +1,26 @@
...
+from django.conf import settings
+from django.db import models
+
+
+class OrganizationManager(models.Manager):
+    def get_default(self):
+        return self.get(name=settings.DEFAULT_ORGANIZATION_NAME)
+
+
+class Organization(AbstractOrganization):
```
mkanoor, line 11: @Dostonbek1 Shouldn't these models be coming from DAB. We install the app ansible_base.rbac shouldn't that app be installing the models needed by the DAB app.

AlanCoding: `dab_rbac` (ansible_base.rbac as import path) is an app, but it's agnostic to the organization model. It allows defining parent models of other models so that roles can delegate permission to child objects, and making organizations a common parent is just what's usually done.

The `Team` model (analogous to `auth.Group`) is special, in that it has RBAC implications, but it is not a concrete model inside of the rbac app. The project using DAB points DAB to its own team model in settings, and rbac app migrations point to it as a swappable model. This is similar to (and modeled after) the `User` model. The models in DAB RBAC have relational connections to both user & team models, but don't define these models.

mkanoor: Is every component AAP, EDA going to have their own Organization, Team concepts? I thought that was a shared concept across the platform, and if a change needs to be made it would be made in the shared model instead of each app creating their own models, leading to customer confusion about these concepts. In EDA we don't have a requirement for a custom Organization and Team, we can use the same concrete model that AAP uses,  which would be consistent across the platform. If we look at the django.contrib.auth it does provide a concrete User model that apps can use but if they need to customize it you can inherit from the AbstractUser and build your own models. I was under the impression that DAB would be providing a similar capability with RBAC support.

AlanCoding: I'll volunteer the background that AWX already uses the `auth.User` model, and galaxy_ng already uses the `auth.Group` model. Earlier iterations of this RBAC library used `auth.Permission` but then we reversed course and added a concrete permission model in the dab_rbac app, due to similar arguments to the ones you are making here.

src/aap_eda/api/serializers/auth.py
```diff
@@ -114,19 +131,18 @@ class RoleDetailSerializer(serializers.Serializer):
...
 
-    modified_at = serializers.DateTimeField(
-        required=True,
-        help_text="The modified_at timestamp of the role",
-    )
+    def get_created_at(self, obj):
+        return obj.created
+
+    def get_modified_at(self, obj):
+        return obj.modified
```
Alex-Izquierdo, line 141: IMO It's crucial to document the purpose of these methods and its role in satisfying the DAB interface requirements. A detailed docstring would help future developers understand why this method exists and prevent accidental removal if its usage isn't immediately obvious.

src/aap_eda/api/serializers/auth.py
```diff
@@ -72,18 +84,24 @@ class RoleSerializer(serializers.ModelSerializer):
...
         default=False, help_text="Indicates if the role is default"
     )
+    created_at = serializers.SerializerMethodField()
+    modified_at = serializers.SerializerMethodField()
+
+    def get_created_at(self, obj):
+        return obj.created
+
+    def get_modified_at(self, obj):
+        return obj.modified
```
Alex-Izquierdo, line 94: IMO It's crucial to document the purpose of these methods and its role in satisfying the DAB interface requirements. A detailed docstring would help future developers understand why this method exists and prevent accidental removal if its usage isn't immediately obvious.

src/aap_eda/core/models/organization.py
benthomasson: We can have a long term goal of a shared implementation, but we need a plan to get there over time.  Moving everything to DAB is not feasible in the short term.

benthomasson: First we need to list the requirements, fields, and behaviors for each model in each component and come up with a union of those for the general case.   We might need to split out certain fields and behaviors into other models if they don't really fit into the general one that works for the other components.    Then we need to plan when those changes can happen over several releases and the migration path for each one.

benthomasson: The cost of that effort is much greater than having a concrete implementation in each component.   The concrete implementation helps us define the requirements for that component anyway.  Defining the concrete requirements is part of the abstraction/generalization process.

pyproject.toml
```diff
@@ -42,8 +42,10 @@ cryptography = ">=38,<42"
 kubernetes = "26.1.*"
 podman = "4.9.*"
 rq-scheduler = "^0.10"
-django-ansible-base = { git = "https://github.com/ansible/django-ansible-base.git", tag = "2024.2.12", extras = [
+# temporarily pinning to devel branch to use RBAC libary
+django-ansible-base = { git = "https://github.com/ansible/django-ansible-base.git", branch = "devel", extras = [
```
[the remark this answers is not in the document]
AlanCoding: https://github.com/ansible/django-ansible-base/actions/workflows/release.yml

We can request and release when needed, and then update this. That is a good idea. But even at this present moment I have several outstanding DAB patches I want to get in before doing that.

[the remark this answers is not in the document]
Dostonbek1: Yes, we will have an updated pinned version before merging this PR. Thus, I had added this in the PR descriptions:
```
NOTE: we'll cut a new release in DAB once all fixes for RBAC are in and pin to that release version before merging this PR.
```

Alex-Izquierdo:
@Dostonbek1 what about the errors in the e2e test suite? Seems related with changes of the type of the ID.

pyproject.toml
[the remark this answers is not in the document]
Alex-Izquierdo: Sorry by skip that comment

src/aap_eda/api/filters/organization.py
Dostonbek1: @mkanoor I think that would be ideal. I know we've discussed using the filters from DAB for other views as well. So I wonder if this should be done in a separate epic for all views later on or use it for this view then implement for others later. Thoughts?

src/aap_eda/api/filters/eda_credential.py
```diff
@@ -24,7 +24,7 @@ class EdaCredentialFilter(django_filters.FilterSet):
         label="Filter by EDA credential name.",
     )
     credential_type_id = django_filters.NumberFilter(
-        field_name="credential_type_id",
+        field_name="credential_type__id",
```
mkanoor, line 27: @Dostonbek1 @hsong-rh Why does this have to be double underscore? This was added in PR https://github.com/ansible/eda-server/pull/819

hsong-rh: @mkanoor I feel both ways should be ok. One is on the field directly, the other is based on the relation.

Dostonbek1: This might be overlook on my end from resolving merge conflicts. I'll test it and push a fix in a bit.

Dostonbek1: It seems to work fine in both scenarios but put the `credential_type_id` back so as not to divert from main branch changes.

src/aap_eda/core/models/credential_type.py
```diff
@@ -34,5 +36,14 @@ class Meta:
     managed = models.BooleanField(default=False)
     kind = models.TextField(default="cloud", blank=True, null=False)
     namespace = models.TextField(default=None, null=True)
+    organization = models.ForeignKey(
+        "Organization", on_delete=models.CASCADE, null=True
```
mkanoor, line 40: @Dostonbek1 Can we use a default field here akin to this https://github.com/ansible/eda-server/blob/d99a07a23852bae28f896c4ccb2f17292bd37d84/src/aap_eda/core/models/event_stream.py#L89 and we can skip all the save calls.

src/aap_eda/core/models/project.py
```diff
@@ -68,6 +70,9 @@ class ScmType(models.TextChoices):
         default=None,
         on_delete=models.SET_NULL,
     )
+    organization = models.ForeignKey(
```
mkanoor, line 73: @Dostonbek1 @benthomasson Can we use simple inheritance from a base model that has the ForeignKey for organization, if we add a new model down the road and it needs to have Organization in it we can inherit from the base model. Then we don't need this in every class. We had done this in catalog with Tenants which is needed in majority of our objects. https://github.com/ansible/pinakes/blob/122fc3184b0c6b6745de220bbd42e632531ec93b/pinakes/main/models.py#L49

benthomasson: Sure.  This is a good use of inheritance in Django, but we should try to keep the inheritance hierarchy to two levels:   
base models and models.

jshimkus-rh:
Not as part of this effort (don't want to delay) but seeing all the api code written to handle externals being adopted and front-ended by eda perhaps we should consider bundling those and making them generally available for re-use? 

In an ideal world such would be taken and maintained by the source of the externals, but even if not it could be a benefit to a large community.

src/aap_eda/core/models/activation.py
```diff
@@ -117,3 +127,9 @@ class Meta:
         default=None,
         help_text="Name of the kubernetes service",
     )
+
+    def save(self, *args, **kwargs):
```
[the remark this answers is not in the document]
benthomasson: https://docs.djangoproject.com/en/5.0/topics/db/queries/#updating-multiple-objects-at-once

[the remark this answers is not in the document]
benthomasson: Using filter/update instead of get/save is how django scales to large numbers of rows and bulk operations.   Putting logic in save() eliminates the possibility of doing bulk operations.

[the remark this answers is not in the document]
Dostonbek1: Put up this PR to change this following @mkanoor 's suggestion in earlier comments:
https://github.com/ansible/eda-server/pull/831

Dostonbek1:
Last e2e run: https://github.com/ansible/eda-server/actions/runs/8753315930

mkanoor (approved):

PR #714, opened 2024-03-06
Remove ExtraVar and EventStream from permission registry

Dostonbek1:


AlanCoding (approved):
Yes, great to get this in! Having fewer root-CRUD models kind of makes the whole app easier for me to understand.

AlanCoding:
In addition to these - I found 1 more exception I can't easily make sense of - eventstream.

This model lists an "add" permission, but I can't find any URL that would correspond to adding an event stream object.

In other cases, a model might not have an obvious view to add a new object of that type, but won't list the "add" permission so that makes sense. Presumably some other action creates the object as a side effect. But as far as I can tell, the event stream add permission does no good.

Dostonbek1:
> In addition to these - I found 1 more exception I can't easily make sense of - eventstream.
> 
> This model lists an "add" permission, but I can't find any URL that would correspond to adding an event stream object.
> 
> In other cases, a model might not have an obvious view to add a new object of that type, but won't list the "add" permission so that makes sense. Presumably some other action creates the object as a side effect. But as far as I can tell, the event stream add permission does no good.

EventStream viewset exists on main branch and needs to be brought into rbac branch. It is very similar to how Activations work. https://github.com/ansible/eda-server/blob/main/src/aap_eda/api/views/event_stream.py

AlanCoding (approved):

PR #719, opened 2024-03-07
fix: rulebook validation accepts actions

bzwei:
AAP-21573: Rulebook validation fails on rulebooks with the 'actions' keyword


Alex-Izquierdo (approved):

Dostonbek1 (approved):

ttuffin (approved):
Tests are passing locally. CI failure is due to another problem.
`122 passed, 5 skipped, 1484 warnings in 1488.79s (0:24:48)`

PR #722, opened 2024-03-07
Validate credential schema and inputs

bzwei:


src/aap_eda/api/serializers/utils.py
```diff
@@ -140,3 +140,85 @@ def inputs_to_store(inputs: dict, old_inputs_str: str = None) -> str:
 
 def inputs_from_store(inputs: str) -> dict:
     return yaml.safe_load(inputs)
+
+
+def validate_inputs(schema: dict, inputs: dict) -> dict:
```
mkanoor, line 145: @bzwei can you add please some test data around this?

mkanoor: @bzwei Also lets move this to a different file maybe credentials/utils.py or credentials_utils.py

bzwei: moved to aap_eda.core.utils.credential_utils.py

src/aap_eda/core/utils/credential_utils.py
```diff
@@ -0,0 +1,154 @@
+#  Copyright 20244 Red Hat, Inc.
```
mkanoor, line 1: @bzwei 20244 or 2024? also I think credentials.py would be a better name since its already in the utils folder

src/aap_eda/core/utils/credential_utils.py
```diff
@@ -0,0 +1,154 @@
...
+    else:
+        for field in fields:
+            for option in ["id", "label"]:
+                value = field.get(option)
+                if not value or not isinstance(value, str):
+                    errors.append(f"{option} must exist and be a string")
+                elif option == "id":
+                    field_ids.append(value)
+
+            field_type = field.get("type")
```
mkanoor, line 124: @bzwei do we need to check for duplicate field ids.

bzwei: I'll leave it a separate PR.

mkanoor (approved):

PR #729, opened 2024-03-08
Tests for DAB RBAC, organization_id fix, related obj permission checking

AlanCoding:
Premise: Use `user_api_client` to test that eda-server views are giving the correct answer, given a certain permission setup in DAB RBAC.

Because this is mostly testing object or action permissions (querysets are not in scope right here), the permission evaluation is a True/False kind of result. To fully test a single case, we always need 2 requests - one made without the permission and one made after getting the permission.

Breaking the work down:
 - [x] Test CRUD of all models registered with the permission registry
 - [x] Test custom actions for all models where that applies
 - [x] Test permission checks for related objects
 - ~[ ] Test views of any models _not_ registered with the registry, assuming they defer to some related object~

Kicked the 4th point out of scope.


AlanCoding:
Test result update after rebase:

```
FAILED tests/integration/dab_rbac/test_crud_permissions.py::test_view_permissions[RulebookProcess] - django.urls.exceptions.NoReverseMatch: Reverse for 'rulebookprocess-detail' not found. 'rulebookprocess-detail' ...
FAILED tests/integration/dab_rbac/test_crud_permissions.py::test_change_permissions[Activation] - AssertionError: {'detail': ErrorDetail(string='Method "PATCH" not allowed.', code='method_not_allowed')}
FAILED tests/integration/dab_rbac/test_related_permissions.py::test_organization_field_not_editable[Activation] - AssertionError: {'detail': ErrorDetail(string='Method "PATCH" not allowed.', code='method_not_allowed')}
========================== 3 failed, 56 passed, 25 skipped, 2 warnings in 94.08s (0:01:34) ==========================
```

This still suggests removing "change" permission from the Activation model. And RulebookProcess is registered, but can't find the URL.

src/aap_eda/api/serializers/activation.py
```diff
@@ -143,6 +143,8 @@ def to_representation(self, activation):
 class ActivationCreateSerializer(serializers.ModelSerializer):
     """Serializer for creating the Activation."""
 
+    organization_id = serializers.IntegerField(required=False, allow_null=True)
```
Dostonbek1, line 146: Why do we need this?

AlanCoding: You are looking at the `ActivationCreateSerializer`. This is used in the view create code to create `serializer`

https://github.com/ansible/eda-server/blob/41e181dee75467e0d75810f998871d0723a19508/src/aap_eda/api/views/activation.py#L83-L85

And then from `serializer`, a `response` is created, which is really the new activation object, in `response = serializer.create(serializer.validated_data)`.

If `organization_id` is not added as a serializer field, then it cannot be in `validated_data`. I know there is some confusion about this because of the use of `ModelSerializer`, which automatically builds fields from the Django model definition. But that would be `organization` that gets automatically built, which is not listed in fields of this serializer. So for the created object to have the organization specified in the request data, this is needed. Similar patterns already exist in other serializers.

Dostonbek1: Hmm.. we have `organization_id` listed in Meta.fields. I thought that would be enough to have the organization id surfaced since we are using ModelSerializer.

Dostonbek1 (approved):

PR #731, opened 2024-03-11
fix: ensure default supported plugins for port forwarding

Alex-Izquierdo:
Follow up https://github.com/ansible/eda-server/pull/707, 
We should support always the default list of plugins, users should not be aware of that in advance to override


Alex-Izquierdo:
On a second thought, I think users might want to not support the default plugins, if they override them, they must provide the whole list, adding implicit config behind the scenes is not a good idea.

PR #735, opened 2024-03-12
fix: wrong serializer for worker nodes

benthomasson:


Alex-Izquierdo (commented):
I think this is already defined here: https://github.com/ansible/eda-server/blob/main/src/aap_eda/core/tasking/__init__.py#L127

benthomasson:
This isn't needed yet,  I accidentally update rq to a newer version.

PR #741, opened 2024-03-14
fix: signature_validation_credential_id not persisted in create api

bzwei:


mkanoor:
@bzwei can you redo this PR with the other changes from credential-spike @hsong-rh has fixed all the test failures

mkanoor (approved):

PR #755, opened 2024-03-18
refactor: parameterize use of credential type names

jshimkus-rh:


Alex-Izquierdo (approved):

src/aap_eda/api/serializers/activation.py
```diff
@@ -157,7 +157,7 @@ def _create_system_eda_credential(
 
 
 def _get_vault_credential_type():
-    return models.CredentialType.objects.get(name="Vault")
+    return models.CredentialType.objects.get(name=CredentialType.VAULT)
```
mkanoor, line 160: @jshimkus-rh This CredentialType is going to be deleted https://github.com/ansible/eda-server/blob/06584d3d153db869a536710349b3d68eeeec4f57/src/aap_eda/core/enums.py#L84 those types don't conform to the new CredentialType which can have user defined names. We have ones that we ship from the Factory which will be
1. GPG Public Key
2. Vault
3. Red Hat Ansible Automation Platform
4. Source Control
5. Container Registry
The users are going to add their own types, so it can be just an enum anymore. The default 5 can be added as a string constants to be shared but this doesn't match the old Credential Type definition.

PR #766, opened 2024-03-20
Refactor & fix integration tests on RBAC branch

Dostonbek1:
This PR aims to refactor existing integration tests as well as updating them to adjust to the new RBAC implementation, although the bulk of the changes are refactors for tests to use fixtures. 

EDIT:
All tests are now passing 🟢

### DIFF
--- a/src/aap_eda/api/views/user.py
+++ b/src/aap_eda/api/views/user.py
@@ -11,7 +11,6 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-import django.db.utils
 from django_filters.rest_framework import DjangoFilterBackend
 from drf_spectacular.utils import (
     OpenApiParameter,
@@ -25,7 +24,6 @@ from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet
 
 from aap_eda.api import filters, serializers
-from aap_eda.api.exceptions import Conflict
 from aap_eda.core import models
 
 from .mixins import (
@@ -156,15 +154,7 @@ class CurrentUserAwxTokenViewSet(
         )
 
     def perform_create(self, serializer):
-        try:
-            serializer.save(user=self.request.user)
-        except django.db.utils.IntegrityError:
-            name_exists = models.AwxToken.objects.filter(
-                user=self.request.user, name=serializer.validated_data["name"]
-            ).exists()
-            if name_exists:
-                raise Conflict("Token with this name already exists.")
-            raise Conflict
+        serializer.save(user=self.request.user)
 
 
 @extend_schema_view(
--- a/tests/integration/api/conftest.py
+++ b/tests/integration/api/conftest.py
@@ -17,7 +17,6 @@ from unittest import mock
 
 import pytest
 from ansible_base.rbac.models import DABPermission, RoleDefinition
-from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from rest_framework.test import APIClient
 
@@ -32,15 +31,6 @@ from aap_eda.core.enums import (
 ADMIN_USERNAME = "test.admin"
 ADMIN_PASSWORD = "test.admin.123"
 DUMMY_UUID = "8472ff2c-6045-4418-8d4e-46f6cffc8557"
-
-
-@pytest.fixture
-def default_organization():
-    "Corresponds to migration add_default_organization"
-    return models.Organization.objects.get_or_create(
-        name=settings.DEFAULT_ORGANIZATION_NAME,
-        description="The default organization",
-    )[0]
 
 
 @pytest.fixture
--- a/tests/integration/api/conftest.py
+++ b/tests/integration/api/conftest.py
@@ -413,17 +413,36 @@ def audit_event_3(
 
 
 @pytest.fixture
-def default_extra_var(
-    default_organization: models.Organization,
-) -> models.ExtraVar:
-    """Return a default ExtraVar"""
-    return models.ExtraVar.objects.create(
-        extra_var="""
+def extra_var_data() -> str:
+    return """
         ---
         collections:
         - community.general
         - benthomasson.eda
-        """,
+        """.strip()
+
+
+@pytest.fixture
+def vault_extra_var_data() -> str:
+    return """
+        limit: !vault |
+                $ANSIBLE_VAULT;1.1;AES256
+                32323466393537363831636134336565656265336564633366396632616431376363353231396562
+                6334646433623764383863656365386363616136633138390a656331383939323930383061363262
+                62376665646431376464653831633634356432323531613661346339643032356366613564386333
+                6433633539353862620a343931393734343437613666343039643764333162303436306434663737
+                3633
+        """.strip()
+
+
+@pytest.fixture
+def default_extra_var(
+    default_organization: models.Organization,
+    extra_var_data: str,
+) -> models.ExtraVar:
+    """Return a default ExtraVar"""
+    return models.ExtraVar.objects.create(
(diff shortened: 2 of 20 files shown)


tests/integration/api/test_activation.py
```diff
@@ -362,51 +362,32 @@ def test_disable_activation(
...
         f"{api_url_v1}/activations/{default_activation.id}/instances/"
     )
     data = response.data["results"]
     assert response.status_code == status.HTTP_200_OK
     assert len(data) == len(instances)
-    assert data[0]["name"] == instances[0].name
-    assert data[1]["name"] == instances[1].name
+    # invert the indices because response data is descending by default
+    assert data[0]["name"] == instances[1].name
+    assert data[1]["name"] == instances[0].name
```
jshimkus-rh, line 378: Performing a suitable sort on both lists will insulate this test from any future changes that affect response content ordering.

Dostonbek1: Good point. I will add sorting for both lists in the test.

Dostonbek1:
Update:
```
104 failed, 345 passed, 28 skipped, 3 warnings
```

tests/integration/api/conftest.py
```diff
@@ -270,6 +272,56 @@ def new_activation(
     )
 
 
+@pytest.fixture
+def default_activation_instances(
+    default_activation: models.Activation, default_project: models.Project
+) -> models.RulebookProcess:
+    """Return a list of Activation Instances"""
+    return models.RulebookProcess.objects.bulk_create(
```
Alex-Izquierdo, line 280: I think one of our problems is that we rely on models instead of objects created by the api. Through the model, we skip the logic in the serializer and with bulk we skip logic in the save method. Have you considered to create the objects through api requests?

Alex-Izquierdo (commented):
Is the plan to unify fixture across all the tests or only for the api? because there are duplicated or similar fixtures in other places that can be also reused. 
For example:
https://github.com/ansible/eda-server/blob/main/tests/integration/services/activation/test_activation.py
https://github.com/ansible/eda-server/blob/main/tests/integration/services/activation/test_activation.py


Anycase, @Dostonbek1 thank you very much for this effort. It is a good improvement. 

There is a internal ticket that I created to approach some of this actions. You may want to reuse it: https://issues.redhat.com/browse/AAP-15713

Dostonbek1:
> Is the plan to unify fixture across all the tests or only for the api? because there are duplicated or similar fixtures in other places that can be also reused. For example: https://github.com/ansible/eda-server/blob/main/tests/integration/services/activation/test_activation.py https://github.com/ansible/eda-server/blob/main/tests/integration/services/activation/test_activation.py
> 
> Anycase, @Dostonbek1 thank you very much for this effort. It is a good improvement.
> 
> There is a internal ticket that I created to approach some of this actions. You may want to reuse it: https://issues.redhat.com/browse/AAP-15713

@Alex-Izquierdo I have been going through tests file by file in unit tests and have only done test_activation and test_activation_instance so far. That's why you might see the rest of the tests are still waiting to be refactored. But the goal is to make fixtures to be reusable across all tests and avoid duplication as much as possible.

tests/integration/api/conftest.py
Dostonbek1: I agree. But I actually don't have much info on what the standard conventions are regarding using fixtures for object creations in tests. I know we already do what you are suggesting in e2e tests because that's the only way in e2e tests. Anyway, I'm open to going that way if others also deem it more reasonable.

jshimkus-rh: As with many things, it depends. 

1. If a test is exercising a class behavior it's reasonable to use the class directly avoiding potential issues from other source.
2. If a test is exercising a higher-level behavior that requires an instance of a particular class creating the instance at the level of that behavior (assuming that is where it is normally created) is reasonable by the same rationale as in #1.  
3. Any other creation of an instance would be done as in production, as it would be assumed to work, providing the benefit of early detection of issues.

Of course, the above isn't meant to be a straitjacket.  If there are issues with code that is not under test getting in the way of progress relaxing the above may be the correct thing to do.  Though it should be noted so that it can later be "corrected."

jshimkus-rh: Making changes to adhere to what I described would not be a short-term effort. 🤕

Dostonbek1:
Update after refactoring test_credential tests:
```
97 failed, 352 passed, 28 skipped, 3 warnings
```

tests/integration/api/conftest.py
Alex-Izquierdo: Agree with all. @Dostonbek1 It should not be a effort to address everything. We can split it and iterate. But I think it can be a good opportunity to define the foundations. I think we can make https://issues.redhat.com/browse/AAP-15713 a epic so we can create multiple tickets there. 

Regarding models I would say that by default, we always want the object created through the request. There may be exceptions, but I think there will be less expections in that way rather than models.

Dostonbek1: @jshimkus-rh @Alex-Izquierdo Thank you both for feedback/thoughts. I have reached a point where I refactored most of the test modules to use fixtures and all tests are now passing. In some of the modules, I made changes just enough for tests to pass and left comments as needed for future refactoring; I wanted to avoid making this PR bigger than it already is and we can work on those test modules in future iterations. Please, give this PR another look. :)

Dostonbek1:
All tests are passing now 🥳

src/aap_eda/api/views/user.py
```diff
@@ -156,15 +154,7 @@ def get_queryset(self):
...
-        try:
-            serializer.save(user=self.request.user)
-        except django.db.utils.IntegrityError:
-            name_exists = models.AwxToken.objects.filter(
-                user=self.request.user, name=serializer.validated_data["name"]
-            ).exists()
-            if name_exists:
-                raise Conflict("Token with this name already exists.")
-            raise Conflict
+        serializer.save(user=self.request.user)
```
AlanCoding, line 157: Sanity check of diff - `perform_create` override here is still necessary because the intent is to get the user from the _request_, not from the _request data_.

tests/integration/dab_rbac/conftest.py
```diff
@@ -64,6 +64,10 @@ def make_name(self):
 
     def fk_fields(self, cls):
         for field in cls._meta.concrete_fields:
+            # corner case for when either an activation or a event_stream
+            # must be set, but not both
+            if cls == RulebookProcess and field.name == "event_stream":
+                continue
```
AlanCoding, line 70: Nice to see this worked :D

tests/integration/api/conftest.py
```diff
@@ -41,13 +40,12 @@ def admin_user(default_organization):
...
     )
-    organization = models.Organization.objects.get_default()
     admin_role = RoleDefinition.objects.create(
         name="Test Admin",
-        content_type=ContentType.objects.get_for_model(organization),
+        content_type=ContentType.objects.get_for_model(default_organization),
     )
     admin_role.permissions.add(*DABPermission.objects.all())
-    admin_role.give_permission(user, organization)
+    admin_role.give_permission(user, default_organization)
```
AlanCoding, line 48: :+1: all looks good here. I know this is convoluted, but this did help clear up some issues.

So the testing system is roughly "same as the old system" but in the default organization. Except user permissions and some other corner cases...

src/aap_eda/api/views/user.py
Dostonbek1: Correct. This POST endpoint doesn't accept `user` data but rather gets the user making the request to assign in the AxwToken object. This makes sense because users can only create tokens for themselves, not for other users.

AlanCoding (approved):

jshimkus-rh (approved):

PR #776, opened 2024-03-21
feat: add per-entity restart/retention parameters [AAP-20053]

jshimkus-rh:


jshimkus-rh:
Closing.  If we ever revisit we can re-open.

PR #777, opened 2024-03-21
feat: use a playbook to clone projects

bzwei:
Support checkout from branch/tag/commit
Support git ssh connection with private key and passphrase

AAP-20751: Implement Source Control credential authentication in import sync service
AAP-20752: Support ssh url for project SCM url


src/aap_eda/playbooks/project_clone.yml
```diff
@@ -0,0 +1,39 @@
...
+      tags:
+        - update_git
+      block:
+        - name: Update project using git
+          ansible.builtin.git:
+            dest: "{{ project_path | quote }}"
+            repo: "{{ scm_url }}"
+            version: "{{ scm_branch | default(omit) | quote }}"
+            refspec: "{{ scm_refspec | default(omit) }}"
+            key_file: "{{ key_file | default(omit) }}"
```
mkanoor, line 26: @bzwei We should incorporate the GPG Key since that is just data flowing thru

bzwei: I will need to create another PR to include GPG key and test it out.

src/aap_eda/services/project/scm.py
```diff
@@ -0,0 +1,248 @@
...
+    @classmethod
+    def decrypt_key_file(cls, key_file: str, password: str) -> str:
+        cmd = f'{KEYGEN_COMMAND} -p -P {password} -N "" -f {key_file}'
+        child = pexpect.spawn(cmd)
+        index = child.expect(["Failed to load key", pexpect.EOF])
+        if index == 0:
+            raise ScmError("Incorrect passhprase for the private key")
+
+
+class PlaybookExecutor:
```
Alex-Izquierdo, line 211: This class doesn't seem to be reusable for playbooks like its name suggests because we are expecting interactive outputs strongly related with a specific operation/playbook. Should be renamed to avoid misunderstandings?

tools/docker/Dockerfile
```diff
@@ -47,6 +46,7 @@ RUN poetry install -E all --no-root --no-cache \
     && ${PYTHON_BIN} -m pip install gunicorn
 COPY . $SOURCES_DIR/
 RUN poetry install -E all --only-root
+RUN pip install ansible-core
```
Alex-Izquierdo, line 49: I think we should force the expected supported version of ansible-core (2.16)

src/aap_eda/services/project/scm.py
```diff
@@ -0,0 +1,248 @@
...
+class ExecutableNotFoundError(Exception):
+    """Executable Not Found error."""
+
+    pass
+
+
+PLAYBOOK_COMMAND = shutil.which("ansible-playbook")
+if PLAYBOOK_COMMAND is None:
+    raise ExecutableNotFoundError("Cannot find ansible-playbook executable")
+PLAYBOOK = f"{os.path.dirname(os.path.abspath(__file__))}/../../playbooks/project_clone.yml"  # noqa E501
```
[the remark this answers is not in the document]
Alex-Izquierdo: @jamesmarshall24 I'm not sure how is built in downstream, do you need to make changes in your side to include data files? Should we create too a `MANIFEST.in` file?

src/aap_eda/services/project/scm.py
```diff
@@ -0,0 +1,283 @@
...
+    ) -> str:
+        result = urlparse(url)
+        domain = result.netloc.split("@")[-1]
+        scheme = result.scheme
+        path = result.path
+
+        if ssh_key and scheme == "" and path.startswith("git@"):
+            return url
+        elif user and password:
+            domain = f"{user}:{password}@{domain}"
```
Alex-Izquierdo, line 213: We need also to support only a token, without user. Ref: https://github.com/ansible/eda-server/blob/main/src/aap_eda/services/project/git.py#L143

There is a bug, that I can not find, where there are providers that have special characters in the user. It might be a good opportunity to set the url with url encoding. 

```
from urllib.parse import quote_plus
...
    encoded_user = quote_plus(user)
    encoded_password = quote_plus(password)
    domain = f"{encoded_user}:{encoded_password}@{domain}"
```
Ref: https://stackoverflow.com/questions/37841914/how-do-i-pass-username-and-password-while-using-ansible-git-module

src/aap_eda/services/project/scm.py
```diff
@@ -0,0 +1,280 @@
...
+        finally:
+            if key_file:
+                key_file.close()
+            if gpg_key_file:
+                gpg_key_file.close()
+        instance = cls(path, _executor=_executor)
+        instance.git_hash = git_hash
+        return instance
+
+    def _execute_cmd(self, cmd: Iterable[str]):
```
Alex-Izquierdo, line 198: This method doesn't seem to be called never.

src/aap_eda/services/project/scm.py
```diff
@@ -0,0 +1,280 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from __future__ import annotations
+
+import logging
+import os
+import shutil
+import tempfile
+from typing import Iterable, Optional
```
Alex-Izquierdo: deprecated import for iterable. Use `from collections.abc import Iterable` instead. 
Refs:
https://docs.astral.sh/ruff/rules/deprecated-import/
https://docs.python.org/3.11/library/typing.html

src/aap_eda/services/project/scm.py
```diff
@@ -0,0 +1,287 @@
...
+        cmd = f'{KEYGEN_COMMAND} -p -P {password} -N "" -f {key_file}'
+        child = pexpect.spawn(cmd)
+        index = child.expect(["Failed to load key", pexpect.EOF])
+        if index == 0:
+            raise ScmError("Incorrect passhprase for the private key")
+
+    @classmethod
+    def add_gpg_key(cls, key_file: str) -> None:
+        cmd = f"{GPG_COMMAND} --import {key_file}"
+        pexpect.run(cmd)
```
Alex-Izquierdo, line 246: If the content of the key is wrong this command will fail. Don't we want to handle that case?

src/aap_eda/services/project/scm.py
```diff
@@ -0,0 +1,287 @@
...
+
+
+class GitAnsibleRunnerExecutor:
+    ERROR_PREFIX = "Failed to clone the project:"
+
+    def __call__(
+        self,
+        extra_vars: dict,
+    ):
+        with tempfile.TemporaryDirectory(prefix="AAP_RUNNER") as data_dir:
```
Alex-Izquierdo, line 256: Would be better to use "EDA_RUNNER" as prefix?

src/aap_eda/services/project/scm.py
```diff
@@ -0,0 +1,287 @@
...
+        cls, url: str, user: str, password: str, ssh_key: str
+    ) -> str:
+        result = urlparse(url)
+        domain = result.netloc.split("@")[-1]
+        scheme = result.scheme
+        path = result.path
+
+        if ssh_key and scheme == "" and path.startswith("git@"):
+            return url
+        elif user and password:
```
Alex-Izquierdo, line 217: unnecessary elif after a return statement.

Alex-Izquierdo (dismissed):

mkanoor (approved):

Alex-Izquierdo (approved):

PR #778, opened 2024-03-25
chore: cleanup old credentials

hsong-rh:
https://issues.redhat.com/browse/AAP-21987


tests/integration/wsapi/test_consumer.py
```diff
@@ -155,69 +154,6 @@ async def test_handle_workers_with_system_vault_credential(
             assert data[0]["label"] == EDA_SERVER_VAULT_LABEL
 
 
-@pytest.mark.django_db(transaction=True)
```
mkanoor, line 158: @hsong-rh We might want to switch this to using the new eda-credential

src/aap_eda/core/migrations/0028_remove_activation_credentials_and_more.py
```diff
@@ -0,0 +1,35 @@
...
+    operations = [
+        migrations.RemoveField(
+            model_name="activation",
+            name="credentials",
+        ),
+        migrations.RemoveField(
+            model_name="activation",
+            name="system_vault_credential",
+        ),
+        migrations.RemoveField(
```
mkanoor, line 23: @hsong-rh If we did this the migration in create_initial_data will fail.

mkanoor: @hsong-rh https://github.com/ansible/eda-server/blob/909c4b40b53d506a0f9004aa617ff25ddcc445e1/src/aap_eda/core/management/commands/create_initial_data.py#L435 this will fail

mkanoor:
@hsong-rh We might need a TODO even for the removal of credential from project https://github.com/ansible/eda-server/blob/909c4b40b53d506a0f9004aa617ff25ddcc445e1/src/aap_eda/core/management/commands/create_initial_data.py#L468 for this code https://github.com/ansible/eda-server/blob/909c4b40b53d506a0f9004aa617ff25ddcc445e1/src/aap_eda/core/models/project.py#L60

PR #780, opened 2024-03-27
Add back queryset, use filter_queryset

AlanCoding:
This adopts the standard change from https://github.com/ansible/django-ansible-base/pull/243

It needs that merged to update the poetry.lock.

The diff here _adds back_ the `queryset =` lines. So that should be removed from the change set in https://github.com/ansible/eda-server/pull/710, and the only net change for implementing DAB RBAC will be the addition of `filter_queryset`.

This resolves the issue where OpenAPI can't figure out what the viewset model is (although I can, I don't see what their problem is) because we used `get_queryset`.

AAP-22130


AlanCoding:
Fixing remaining failure requires https://github.com/ansible/django-ansible-base/pull/244, which also needs approval and merge.

AlanCoding:
The DAB change is merged so I am hopeful for this to pass now.

dhaustein:
@AlanCoding Hi, this looks good on the E2E API tests side! :+1:

Dostonbek1:
@AlanCoding are we good to merge this, or is there more work to be done?

AlanCoding:
the lockfile and stuff will be out of date by now. I will try to do a quick rebase now.

AlanCoding:
Merge after https://github.com/ansible/eda-server/pull/799

AlanCoding:
That other merge should address check I see failing that migrations are up-to-date

PR #782, opened 2024-03-27
feat: multi node support with local queues

mkanoor:
### Key Features
* Create a local queue per node for a group of activation workers using the following
    setting **EDA_LOCAL_QUEUE_NAME**

* Activation workers listen on **global activation** queue and the **EDA_LOCAL_QUEUE_NAME**

* Nodes can be added post installation without running the main installer on the master node, new nodes would get discovered and work distributed to them.

* A dedicated podman/k8s event listener per node On each node run an activation worker that just listens to events emanating from the container engine (podman/k8s) Once the event comes in it dispatches the work to an activation worker to check the status of activation.
* Increases the monitor time interval to 10 minutes so we don't spend too many CPU cycles on monitoring
* At the end of processing a request kick off another monitor requests queue so we can run any pending tasks without having to wait for the timer to pop
* Added a new activation end point activation/{id}/monitor which can called by the UI when a user is watching the logs, this can be triggered every 10 seconds when the user is logged in and is interested in watching the logs.

### New Settings

**LOCAL_QUEUE_NAME** Name of Local queue can be the hostname, default is None
**REFRESH_RQ_QUEUES** Refresh the RQ_QUEUES from MQ default is True
**WORKER_HEARTBEAT_DELAY** If the worker doesn't respond in 360 seconds we consider it to be dead and move the activation to another node

### New Periodic Task:
Run a periodic task on the **default** queue to monitor all the nodes, if the node stops responding we can move the task to a newer node.
```
    "func": "aap_eda.tasks.orchestrator.monitor_nodes",
     "interval": 300,
```


src/aap_eda/services/activation/manager.py
```diff
@@ -1050,6 +1073,23 @@ def _create_activation_instance(self):
...
+            monitor_events_queue = settings.LOCAL_QUEUE_NAME
+        else:
+            monitor_events_job_id = "activation"
+            monitor_events_queue = "activation"
+
+        # Start the Monitor Engine events once
+        if not ActivationManager.started_monitor_events:
+            ActivationManager.started_monitor_events = True
+            start_monitor_engine_events_job(
+                monitor_events_queue, monitor_events_job_id
```
Alex-Izquierdo, line 1090: `monitor_events_job_id` should not be of the form "{queue}-{activation-id}"? otherwise the unique_enqueue will discard it

src/aap_eda/tasks/orchestrator.py
```diff
@@ -184,4 +223,113 @@ def monitor_rulebook_processes() -> None:
...
+
+
+def monitor_nodes() -> None:
+    """Monitor if nodes are alive.
+
+    Started by the scheduler at a regular cadence
+    """
+    LOGGER.info("Checking Worker Health")
+    activation_values = QUEUES["activation"]
+    redis = Redis(
```
Alex-Izquierdo, line 288: While working with django-rq we should use `django_rq.get_connection`

src/aap_eda/tasks/orchestrator.py
```diff
@@ -13,9 +13,14 @@
...
 import logging
-from typing import Union
+import time
+from typing import Optional, Union
 
+from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
+from django_rq.settings import QUEUES
+from redis import Redis
+from rq import Worker as _Worker
```
Alex-Izquierdo, line 23: This should be our custom worker class.

src/aap_eda/tasks/orchestrator.py
```diff
@@ -184,4 +223,113 @@ def monitor_rulebook_processes() -> None:
...
+
+    Started by the scheduler at a regular cadence
+    """
+    LOGGER.info("Checking Worker Health")
+    activation_values = QUEUES["activation"]
+    redis = Redis(
+        host=activation_values["HOST"], port=activation_values["PORT"]
+    )
+    now = int(time.time())
+    in_use_queues = {node.name for node in models.Node.objects.all()}
```
Alex-Izquierdo, line 292: How we would handle obsolete nodes? this should take into account the related activations that are in "running" state?

src/aap_eda/tasks/orchestrator.py
```diff
@@ -184,4 +223,113 @@ def monitor_rulebook_processes() -> None:
...
+                if queue.name in ["activation", "default"]:
+                    continue
+                active_queues.add(queue.name)
+
+    LOGGER.info(f"In Use Queues {in_use_queues}")
+    LOGGER.info(f"Active Queues {active_queues}")
+    for queue in in_use_queues:
+        dead_queues.add(queue)
+
+    dead_queues = in_use_queues - active_queues
```
Alex-Izquierdo, line 318: Here we are considering a queue dead if it has a worker dead, but it could have other workers running ok.

Alex-Izquierdo (commented):
The problem that I see are the implicit nodes without control from the user. Adding a node is just a new worker connected. What if that worker/node is obsolete? How the user can remove/ban a node? 

I think the user must have explicit control to let know to the system what are the expected nodes and the topology. 

I tried a [similar approach](https://github.com/ansible/eda-server/compare/main...Alex-Izquierdo:eda-server:worker-monitor) storing the queues in the DB and managing them through CLI commands. But I putted it on-hold because it looked it to me too complex and I needed to hack the initialization, that approach requires definitely the full removal of django-rq. The local queue seems a good workaround, but I wonder if we should avoid django-rq and just use dynamic queues based on the input of the user. 

A second problem that I see is the lack of a balancing system, a new activation could fall in a node without capacity when there could be nodes with capacity, eventually this loop might happen indefinitely. I think that in every multinode scenario we need to make sure the activation is started if it's possible or be sure that can not be started.

src/aap_eda/tasks/orchestrator.py
```diff
@@ -184,4 +223,113 @@ def monitor_rulebook_processes() -> None:
...
+
+    Started by the activation worker once, before it kicks on activation
+    It enqueues a monitor task for each activation when it ends
+    """
+    LOGGER.info("Monitor engine events has started")
+    container_engine = new_container_engine(1, ProcessParentType.ACTIVATION)
+    for event in container_engine.monitor_events():
+        # monitor running instances
+        LOGGER.info(f"Container engine event {event}")
+        if event.reason in ["died", "Completed", "BackoffLimitExceeded"]:
```
Alex-Izquierdo, line 241: I think this should be a enum defined from the Event dataclass which acts as a interface for the engine, otherwise you have to deal with free strings that comes from different engines.

Alex-Izquierdo (commented):
What do you think about adding a similar docker-compose with multinode as I have in https://github.com/ansible/eda-server/pull/701/files
In that way this PR can be tested in place.

src/aap_eda/settings/default.py
```diff
@@ -319,7 +319,11 @@ def _get_secret_key() -> str:
 RQ_PERIODIC_JOBS = [
     {
         "func": "aap_eda.tasks.orchestrator.monitor_rulebook_processes",
-        "interval": 5,
+        "interval": 600,
```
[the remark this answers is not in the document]
mkanoor: @Alex-Izquierdo What we need to know is when the activation process ends. For that the event monitor will trigger a refresh as soon as the container ends which is better than what we have today we have to wait for the monitor to wake up and detect that the container has ended..The even monitor relegates the periodic monitor to run with longer delays since it doesn't play an important role anymore.

src/aap_eda/tasks/orchestrator.py
mkanoor: Not sure about that each worker has a list of queues that they are connected to it, so if we have 2 workers
* worker1 (local_node1, activation)
* worker2(local_node1, activation)

Even if one of the dies above we will still have local_node1 and activation as the queues still available

src/aap_eda/settings/default.py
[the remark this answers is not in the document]
Alex-Izquierdo: Should I guess there is a missing periodic job for `monitor_engine_events` ?
Even in that case we still have the unresponsive case and regular logs updates. You want to update the logs each 10 minutes or force it to be a manual sync? that's a bad UX IMO.

src/aap_eda/services/activation/manager.py
mkanoor: @Alex-Izquierdo its not a periodic job, its a dedicated activation worker listening to events from the container engine. It's started by the Manager once with a unique id and it will block a single activation worker to listening the events. Doing busy loop anywhere is not a good design if we can handle events coming back from the engine we can completely eliminate the busy loop approach. If there are 3 activation workers on a node, 1 worker is dedcated to listening for events from the engine and the 2 workers do the start/stop/monitoring

src/aap_eda/settings/default.py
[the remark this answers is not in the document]
mkanoor: @Alex-Izquierdo The design allows for the UI to mimic the periodic 10 second monitoring that it can call on the activation that will fetch the logs. There is a new end point /activations/<nn>/monitor which the UI can call when a user is involved. When there is no user involved and we are running with lights out or the log level set to Error there is no need to be wasting CPU cycles.

PR #787, opened 2024-03-29
fix: add missing validation in is_activation_valid

hsong-rh:


src/aap_eda/api/serializers/activation.py
```diff
@@ -37,7 +37,11 @@
 from aap_eda.api.serializers.rulebook import RulebookRefSerializer
 from aap_eda.api.vault import encrypt_string
 from aap_eda.core import models, validators
-from aap_eda.core.enums import CredentialType, ProcessParentType
+from aap_eda.core.enums import (
+    CredentialType,
```
mkanoor, line 41: @hsong-rh This should be removed, there are references to this CredentialType but they are old and should be swapped with DefaultCredentialType

tests/integration/api/test_activation_with_credential.py
```diff
@@ -346,3 +445,113 @@ def test_create_activation_with_conflict_credentials(
...
+    test_activation = TEST_ACTIVATION.copy()
+    test_activation.pop("extra_var_id")
+    test_activation["decision_environment_id"] = fks["decision_environment_id"]
+    test_activation["rulebook_id"] = fks["rulebook_id"]
+
+    credential_types = models.CredentialType.objects.bulk_create(
+        [
+            models.CredentialType(
+                name="user_type",
+                inputs={"fields": [{"a": "b"}]},
```
mkanoor, line 466: @hsong-rh should these fields be called sasl_username and sasl_password instead of "a", since the inputs is being set to different keys. if someone looks at it might be confusing.

tests/integration/api/test_activation_with_credential.py
```diff
@@ -346,3 +445,113 @@ def test_create_activation_with_conflict_credentials(
...
+                injectors={"extra_vars": {"username": "adam"}},
+            ),
+            models.CredentialType(
+                name="password_type",
+                inputs={"fields": [{"1": "2"}]},
```
mkanoor, line 471: @hsong-rh we should fix the inputs here too to be sasl_username and sasl_password.

tests/integration/api/test_activation_with_credential.py
```diff
@@ -346,3 +445,113 @@ def test_create_activation_with_conflict_credentials(
...
+                credential_type_id=credential_types[1].id,
+            ),
+        ]
+    )
+
+    eda_credential_ids = [credential.id for credential in eda_credentials]
+    test_activation["eda_credentials"] = eda_credential_ids
+
+    response = client.post(f"{api_url_v1}/activations/", data=test_activation)
+    assert response.status_code == status.HTTP_201_CREATED
```
mkanoor, line 496: @hsong-rh since we are using 2 credentials with the same values but different injectors a cleaner test would be to convert this to a failure because of extra var name conflicts.
And create another one with just 1 credential with no conflict

hsong-rh: @mkanoor Such test cases have already covered in `test_create_activation_with_key_conflict` and `test_create_activation_with_compatible_credentials`

tests/integration/api/test_activation_with_credential.py
```diff
@@ -346,3 +447,172 @@ def test_create_activation_with_conflict_credentials(
...
+                inputs={"sasl_username": "adam", "sasl_password": "secret"},
+                credential_type_id=credential_types[0].id,
+            )
+        ]
+    )
+
+    eda_credential_ids = [credential.id for credential in eda_credentials]
+    test_activation["eda_credentials"] = eda_credential_ids
+
+    response = client.post(f"{api_url_v1}/activations/", data=test_activation)
```
mkanoor, line 497: @hsong-rh should we check that the extra_var got created here.

tests/integration/api/test_activation_with_credential.py
```diff
@@ -346,3 +447,172 @@ def test_create_activation_with_conflict_credentials(
...
+    test_activation = TEST_ACTIVATION.copy()
+    test_activation["decision_environment_id"] = fks["decision_environment_id"]
+    test_activation["rulebook_id"] = fks["rulebook_id"]
+    test_activation["extra_var_id"] = fks["extra_var_id"]
+
+    credential_types = models.CredentialType.objects.bulk_create(
+        [
+            models.CredentialType(
+                name="user_type",
+                inputs={"fields": [{"a": "b"}]},
```
mkanoor, line 576: @hsong-rh This seems to be off

tests/integration/api/test_activation_with_credential.py
```diff
@@ -346,3 +447,172 @@ def test_create_activation_with_conflict_credentials(
...
+                injectors={"extra_vars": {"username": "adam"}},
+            ),
+            models.CredentialType(
+                name="password_type",
+                inputs={"fields": [{"1": "2"}]},
```
mkanoor, line 581: @hsong-rh This seems to be off

mkanoor (approved):

bzwei (approved):

PR #796, opened 2024-04-04
fix: prevent managed credential types from updating

hsong-rh:
https://issues.redhat.com/browse/AAP-22443


tests/integration/api/test_credential_type.py
```diff
@@ -199,3 +199,21 @@ def test_partial_update_inputs_credential_type(
         assert obj.inputs == new_inputs
     else:
         assert message in response.data["inputs"]
+
+
+@pytest.mark.django_db
+def test_update_managed_credential_type(client: APIClient):
+    obj = models.CredentialType.objects.create(
```
mkanoor, line 206: @hsong-rh can we directly use our SCM credential type here?

mkanoor (approved):

PR #801, opened 2024-04-05
General integration of DAB viewsets into eda-server schema system

AlanCoding:
Replaces https://github.com/ansible/eda-server/pull/775

Requires upstream patch https://github.com/ansible/django-ansible-base/pull/276

Accomplishes the basics of the ask there.

```json
                "responses": {
                    "201": {
                        "content": {
                            "application/json": {
                                "schema": {
                                    "$ref": "#/components/schemas/RoleDefinition"
                                }
                            }
                        },
                        "description": "Return a created RoleDefinition."
                    }
                }
            }
```

Under "post" of the role_definitions list.

Also has the special create serializer

```json
            "post": {
                "operationId": "role_definitions_create",
                "description": "Create a RoleDefinition.",
                "tags": [
                    "role_definitions"
                ],
                "requestBody": {
                    "content": {
                        "application/json": {
                            "schema": {
                                "$ref": "#/components/schemas/RoleDefinitionCreate"
                            }
                        },
                        "application/x-www-form-urlencoded": {
                            "schema": {
                                "$ref": "#/components/schemas/RoleDefinitionCreate"
                            }
                        },
                        "multipart/form-data": {
                            "schema": {
                                "$ref": "#/components/schemas/RoleDefinitionCreate"
                            }
                        }
                    },
                    "required": true
                },
```

Also need to repeat this for the assignments views, but this provides the overall structure of my solution so that we may all live in harmony.

With this, https://github.com/ansible/django-ansible-base/pull/249 will be taken out of draft.


AlanCoding:
Repeated the same general structure for the user & team assignment endpoints as well.

src/aap_eda/api/serializers/auth.py
```diff
@@ -134,10 +136,10 @@ class RoleDetailSerializer(serializers.Serializer):
     created_at = serializers.SerializerMethodField()
     modified_at = serializers.SerializerMethodField()
 
-    def get_created_at(self, obj):
+    def get_created_at(self, obj) -> datetime.datetime:
         return obj.created
 
-    def get_modified_at(self, obj):
+    def get_modified_at(self, obj) -> datetime.datetime:
```
Dostonbek1, line 142: Could you add docstrings for these methods that would knock out one of the [review comments](https://github.com/ansible/eda-server/pull/710#discussion_r1553425548) in feature/rbac PR, since you are already touching this line of code?

src/aap_eda/api/views/dab_base.py
```diff
@@ -0,0 +1,57 @@
+from rest_framework.views import APIView
```
Dostonbek1, line 1: We might wanna add the copyright notices in these new files.

src/aap_eda/api/views/dab_decorate.py
```diff
@@ -0,0 +1,25 @@
+from ansible_base.rbac.api import views
```
Dostonbek1, line 1: Same thing here for copyright notice.

Dostonbek1 (changes requested):

src/aap_eda/api/serializers/auth.py
AlanCoding: This might sound pedantic, but I'd rather keep it out of the scope here since this patch doesn't change anything about it. Also, it's a bit hard for me to write that. I suspect whatever I write would be subject to changes and later revision. There are problem more problems with these endpoints that we haven't digested.

src/aap_eda/api/views/dab_base.py
```diff
@@ -0,0 +1,77 @@
...
+    registry[create_serializer_name] = create_cls
+
+    return create_cls
+
+
+class BaseAPIView(APIView):
+    def get_serializer_class(self):
+        if not hasattr(super(), "get_serializer_class"):
+            # Use base serializer for DAB views that do not give any serializer
+            return Serializer
```
AlanCoding, line 57: See https://github.com/ansible/django-ansible-base/issues/283 for the effect of this change in particular.

PR #806, opened 2024-04-08
[AAP-22421] - public ping endpoint for http healthchecks

msmagnanijr:
Please see:  https://issues.redhat.com/browse/AAP-22421 and https://issues.redhat.com/browse/AAPRFE-1202

### DIFF
--- a/src/aap_eda/core/views.py
+++ b/src/aap_eda/core/views.py
@@ -1,10 +1,33 @@
+from django.db import connection
+from django.db.utils import OperationalError
+from rest_framework.permissions import AllowAny
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
 
 class HealthView(APIView):
-    authentication_classes = []
-    permission_classes = []
+    permission_classes = [AllowAny]
 
-    def get(self, _request):
-        return Response("OK")
+    def get(self, request):
+        try:
+            connection.ensure_connection()
+            return Response({"status": "OK"}, status=200)
+        except OperationalError:
+            return Response(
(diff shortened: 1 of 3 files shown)


Alex-Izquierdo (commented):
we need to expose this view in a public path, /ping/, (right now is only available in /_healthz) 
Also I think it is a good opportunity to improve it adding a basic check of the database connection.
```
    try:
        connection.ensure_connection()
    except OperationalError:
        return HttpResponse("Database connection failed", status=500)
```

msmagnanijr:
> we need to expose this view in a public path, /ping/, (right now is only available in /_healthz) Also I think it is a good opportunity to improve it adding a basic check of the database connection.
> 
> ```
>     try:
>         connection.ensure_connection()
>     except OperationalError:
>         return HttpResponse("Database connection failed", status=500)
> ```

okay, got it!

src/aap_eda/core/views.py
```diff
@@ -1,10 +1,31 @@
...
 
-    def get(self, _request):
-        return Response("OK")
+    def get(self, request):
+        try:
+            connection.ensure_connection()
+            return Response({"status": "OK"}, status=200)
+        except OperationalError:
+            return Response(
+                {"error": "Database connection failed"}, status=500
```
Alex-Izquierdo, line 17: Should we include here also the "status" key for consistency?

msmagnanijr: yep, makes sense!

ttuffin (approved):

PR #808, opened 2024-04-08
chore: check related resources when try to delete eda credentials

hsong-rh:
https://issues.redhat.com/browse/AAP-22026

### DIFF
--- a/src/aap_eda/api/views/eda_credential.py
+++ b/src/aap_eda/api/views/eda_credential.py
@@ -220,17 +220,13 @@ class EdaCredentialViewSet(
                 {"errors": error}, status=status.HTTP_400_BAD_REQUEST
             )
 
-        # If the credential is in use and the 'force' flag
-        # is not True, raise a PermissionDenied exception
-        is_used = models.Activation.objects.filter(
-            decision_environment__eda_credential=eda_credential
-        ).exists()
+        references = get_references(eda_credential)
 
-        if is_used and not force:
+        if bool(references) and not force:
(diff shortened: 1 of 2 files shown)


src/aap_eda/api/views/eda_credential.py
```diff
@@ -220,15 +220,11 @@ def destroy(self, request, *args, **kwargs):
...
-        is_used = models.Activation.objects.filter(
-            decision_environment__eda_credential=eda_credential
-        ).exists()
+        references = get_references(eda_credential)
 
-        if is_used and not force:
+        if bool(references) and not force:
             raise exceptions.Conflict(
-                "Credential is being used by Activations "
+                "Credential is being referenced by resources "
```
mkanoor, line 227: @hsong-rh I think this would be a better message.
```
Credential {name} is being referenced by other resources and cannot be deleted. If you want to force delete, please add the /?force=true query param.
```

mkanoor:
@hsong-rh We would need a test with for the force=true case and when that happens we have to delete the credential type and the credentials. Since the credential cannot be used without the credential type

hsong-rh:
@mkanoor 

> @hsong-rh We would need a test with for the force=true case and when that happens we have to delete the credential type and the credentials. Since the credential cannot be used without the credential type

@mkanoor What we try to delete here are credentials, not credential types. we should keep credential types, even no credentials are associated with it.

mkanoor (approved):

Dostonbek1 (approved):

PR #810, opened 2024-04-08
feat: check if has associated credetials before updating the credential type

hsong-rh:
https://issues.redhat.com/browse/AAP-22563

UI will show the following error if credentials exist.
image

### DIFF
--- a/src/aap_eda/api/views/credential_type.py
+++ b/src/aap_eda/api/views/credential_type.py
@@ -122,6 +122,22 @@ class CredentialTypeViewSet(
                 {"errors": error}, status=status.HTTP_400_BAD_REQUEST
             )
 
+        if models.EdaCredential.objects.filter(
+            credential_type=credential_type
+        ).exists():
+            inputs = request.data.get("inputs")
+            injectors = request.data.get("injectors")
+            if inputs or injectors:
+                field = "inputs" if inputs else "injectors"
+                error = (
+                    f"Modifications to {field} are not allowed for credential "
+                    "types that are in use"
+                )
+                return Response(
+                    data={field: [error]},
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
+
         serializer = serializers.CredentialTypeCreateSerializer(
             credential_type, data=request.data, partial=True
         )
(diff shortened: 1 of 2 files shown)


hsong-rh:
> @hsong-rh The message remains the same for inputs/injectors shouldn't it change based on the field name. Also I think the errors should be reported with the field name so the UI can display it properly. 

Updated

src/aap_eda/api/views/credential_type.py
```diff
@@ -122,6 +122,22 @@ def partial_update(self, request, pk):
...
 
+        if models.EdaCredential.objects.filter(
+            credential_type=credential_type
+        ).exists():
+            inputs = request.data.get("inputs")
+            injectors = request.data.get("injectors")
+            if inputs or injectors:
+                field = "inputs" if inputs else "injectors"
+                error = (
+                    "Modifications to inputs are not allowed for credential "
```
mkanoor, line 133: @hsong-rh should this be
```
f"Modifications to {field} are not allowed for credential "
```

mkanoor (approved):

eclarizio:
Very small nitpick: shouldn't this be "feat" and not "chore"? "chore" (if I understand it correctly) is for something that is internal to the project like adjusting a variable name, refactoring a method, or more broadly upgrading versions of plugins, things like that. This, however, adds new functionality that can be seen directly in the UI and seems more in line with a feature

hsong-rh:
> Very small nitpick: shouldn't this be "feat" and not "chore"? "chore" (if I understand it correctly) is for something that is internal to the project like adjusting a variable name, refactoring a method, or more broadly upgrading versions of plugins, things like that. This, however, adds new functionality that can be seen directly in the UI and seems more in line with a feature

@eclarizio Got you, I thought it meant minor changes.

lgalis (approved):

src/aap_eda/api/views/credential_type.py
```diff
@@ -122,6 +122,22 @@ def partial_update(self, request, pk):
                 {"errors": error}, status=status.HTTP_400_BAD_REQUEST
             )
 
+        if models.EdaCredential.objects.filter(
+            credential_type=credential_type
+        ).exists():
+            inputs = request.data.get("inputs")
+            injectors = request.data.get("injectors")
+            if inputs or injectors:
+                field = "inputs" if inputs else "injectors"
```
bzwei, line 131: What if both fields are changed? The message only says inputs.
Maybe we don't pick field here, simply says Modifications to inputs or injectors are not allowed

bzwei (approved):

PR #813, opened 2024-04-08
[RBAC] Remove Roles & Permission models/endpoints in favor of DAB RBAC

Dostonbek1:
* Remove Role & Permission models/endpoints
* Remove `role` field from User model

JIRA: [AAP-21064](https://issues.redhat.com/browse/AAP-21064)


AlanCoding (approved):
As long as the stakeholders are on-board, this will help a ton with finalizing the RBAC work as a whole.

benthomasson (approved):
I approve of this approach to migrate the roles to RBAC.

PR #814, opened 2024-04-08
chore: fix the image build for e2e tests

mkanoor:
The image was being built from the tools/docker directory which doesn't have the .dockerignore and the .git
directory was getting included in the image.
Moved the image build to be from the same directory that contains the .dockerignore

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,3 +5,4 @@ dist
 .idea
 .venv
 venv
+.github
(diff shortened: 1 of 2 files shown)


Alex-Izquierdo (approved):

bzwei (approved):

mkanoor:
The e2e test failure is because of https://github.com/ansible/eda-qa/pull/326

PR #817, opened 2024-04-09
fix: disable drf browsable api by default

Alex-Izquierdo:
openapi / swagger is the supported documentation for the api. Disable DRF views by default. 
Jira: https://issues.redhat.com/browse/AAP-12014


Alex-Izquierdo:
Seems that we are going to do the contrary.

PR #822, opened 2024-04-11
test: fix TLS-enabled failing tests

jshimkus-rh:


Alex-Izquierdo (commented):
Can you wait to merge first https://github.com/ansible/eda-server/pull/701/files#diff-b1dc31d817d52a3e8b384026c1442223022f882ba1898b45998239041d8c3611

jshimkus-rh:
No problem.  I doubt anyone but me has been affected by TLS. 😄 Yet.

src/aap_eda/settings/default.py
```diff
@@ -346,6 +346,45 @@ def _get_databases_settings() -> dict:
...
+    if (not REDIS_UNIX_SOCKET_PATH) and REDIS_CLIENT_CERT_PATH:
+        params |= {
+            "ssl_certfile": REDIS_CLIENT_CERT_PATH,
+            "ssl_keyfile": REDIS_CLIENT_KEY_PATH,
+            "ssl_ca_certs": REDIS_CLIENT_CACERT_PATH,
+        }
+    return params
+
+
+def rq_test_redis_client_parameters():
```
Alex-Izquierdo, line 379: This function is for tests, should not be here. Is it a mistake or intended?

jshimkus-rh: It is intended.  It eliminates duplication of logic to establish the correct group of parameters to connect to redis for both non-TLS and TLS.

If we want to locate it in some utility source that's fine and will also mean moving the other functions on which it depends to avoid any code "reaching" into settings (something I didn't want to do with tests, but figured it might be ok for that specific use).

jshimkus-rh: OK.  That makes sense.

Alex-Izquierdo (approved):

mkanoor (approved):

Issue #716, opened 2024-03-06
Option to use a Proxy when connecting to Github - Looking for more ways to source rulebooks


mk33650:
### Please confirm the following

- [X] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [X] I have checked the [current issues](https://github.com/ansible/aap-eda/issues) for duplicates.
- [X] I understand that EDA-Controller is open source software provided for free and that I might not receive a timely response.

### Feature type

Enhancement to Existing Feature

### Feature Summary

A field/option (either in the UI, or a parameter offered via the inventory file) to use a proxy to fetch/source Github projects. We use similar functionality with AAP already, however after setting up an instance of EDA we are unable to source projects outside of our internal infrastructure due to firewall rulings.

### Steps to reproduce

Trying to source rulebooks from Github results in a timeout due to firewall rules that we're unable to work around without the use of a proxy.

### Current results

We're unable to use EDA due to limited options to source rulebooks.

### Sugested feature result

The ability for us to use rulebooks :)

PR #823, opened 2024-04-11
feat: support proxy for project cloning

bzwei:
Add proxy field in project model. Set http_proxy and https_proxy envs when running git clone command through ansible-runner.

AAP-22611: Allow user to use proxy for project cloning

https://issues.redhat.com/browse/AAPRFE-835
https://github.com/ansible/eda-server/issues/716


src/aap_eda/services/project/scm.py
```diff
@@ -257,8 +258,18 @@ class GitAnsibleRunnerExecutor:
     def __call__(
         self,
         extra_vars: dict,
+        proxy: str,
     ):
         with tempfile.TemporaryDirectory(prefix="EDA_RUNNER") as data_dir:
+            if proxy:
```
mkanoor, line 264: @bzwei I think according to this doc https://ansible.readthedocs.io/projects/runner/en/latest/ansible_runner/#ansible_runner.interface.run we don't have to create a directory we can directly pass in via enviers
```
envvars ([dict](https://docs.python.org/3/library/stdtypes.html#dict)) – Environment variables to be used when running Ansible. Environment variables will also be read from env/envvars in private_data_dir
```

mkanoor (approved):

hsong-rh (approved):

PR #827, opened 2024-04-12
feat: protect proxy url that may contain a password

bzwei:
Change the proxy field to EncryptedTextField.
Hide password ($encrypted$) part of the proxy when retrieve a project. Require plain password when update a proxy.

AAP-22825: hide password in proxy url


src/aap_eda/api/serializers/project.py
```diff
@@ -48,6 +55,14 @@ class Meta:
             *read_only_fields,
         ]
 
+    def get_proxy(self, obj: models.Project) -> str:
```
Alex-Izquierdo, line 58: The same function appears twice, should be try to avoid duplicated code? maybe through a mixin?

src/aap_eda/api/views/project.py
```diff
@@ -211,6 +221,27 @@ def partial_update(self, request, pk):
...
+                        == value
+                    ):
+                        continue  # value is unchanged. skip
+
+                    error = (
+                        "The password in the proxy field should be unencrypted"
+                    )
+                    return Response(
+                        {"errors": error}, status=status.HTTP_400_BAD_REQUEST
+                    )
```
Alex-Izquierdo, line 244: can you give more context about this logic? Why a partial update would receive a proxy with the content encrypted by the application?

src/aap_eda/api/serializers/project.py
```diff
@@ -247,3 +271,27 @@ class Meta:
...
+def get_proxy_for_display(proxy: str) -> str:
+    if not (proxy.startswith("http://") or proxy.startswith("https://")):
+        proxy = f"http://{proxy}"
+    result = urlparse(proxy)
+    if "@" not in proxy:
+        return proxy
+    cred, domain = result.netloc.split("@")
+    if ":" in cred:
+        user, _ = cred.split(":")
+        domain = f"user:{ENCRYPTED_STRING}@{domain}"
```
Alex-Izquierdo, line 285: User is a plain string, I believe you want here to use the variable. 
`domain = f"{user}:{ENCRYPTED_STRING}@{domain}"`

src/aap_eda/api/serializers/project.py
```diff
@@ -247,3 +271,27 @@ class Meta:
...
+    else:
+        domain = f"{ENCRYPTED_STRING}@{domain}"
```
Alex-Izquierdo, line 287: If you are going to hide whichever is before the "@" you don't need to search for an user in line 284

src/aap_eda/api/serializers/project.py
```diff
@@ -176,6 +192,14 @@ class Meta:
             *read_only_fields,
         ]
 
+    def get_proxy(self, obj: models.Project) -> str:
+        if not obj.proxy:
+            return ""
+        url = obj.proxy
+        if isinstance(url, SecretValue):
```
Alex-Izquierdo, line 199: In the migration you are not encrypting existing values. I think it is because we assume there will not be unencrypted values because the field was implemented a few days ago. So, we always expect an encrypted value and this conditional should not be necessary.  You follow the same logic at the import logic. 

Otherwise I think dealing with a field that might encrypted or not would be inconsistent, so it would be better to encrypt the existing ones in the migration.

src/aap_eda/api/serializers/project.py
bzwei: it is a bug

src/aap_eda/api/serializers/project.py
bzwei: We still want to show {user} if it exists.

src/aap_eda/api/serializers/project.py
```diff
@@ -247,3 +271,27 @@ class Meta:
         model = models.ExtraVar
         fields = ["id"]
         read_only_fields = ["id"]
+
+
+def get_proxy_for_display(proxy: str) -> str:
+    if not (proxy.startswith("http://") or proxy.startswith("https://")):
```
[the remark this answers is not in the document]
mkanoor: @Alex-Izquierdo Yes we don't want to policing what gets set by the user for proxies, they come in all kinds of formats. By inserting ourselves in the middle we will be limiting its use

https://everything.curl.dev/usingcurl/proxies/http.html
The tools that use it apply the appropriate defaults, we are not establishing connection to the Proxy the other tools like git clone, lib curl etc are doing the connection

src/aap_eda/api/serializers/project.py
bzwei: nobody yet started to use this setting. Now directly amend to the original proxy field. No need an extra migration.

src/aap_eda/api/views/project.py
Alex-Izquierdo: I see your point. Would it not be better to run the transformation against the request.data so you can know in advance if the field is updated or not?

mkanoor:
@bzwei @Alex-Izquierdo 
https://linuxtutorials.org/socks5-proxy-environment-variable-linux/

bzwei:
> @bzwei @Alex-Izquierdo https://linuxtutorials.org/socks5-proxy-environment-variable-linux/

I removed the part to force a default http://

src/aap_eda/api/serializers/project.py
Alex-Izquierdo: Agree, what I'm saying is that you don't need to check if the field is a SecretValue, because it should be a SecretValue always.

bzwei: We have to check the type during the create process because it is still the plain text at runtime, haven't been updated to the encrypted value.

Alex-Izquierdo: ok

Alex-Izquierdo (approved):

Alex-Izquierdo (approved):

mkanoor:
Tested with passwords on squid. 👍

mkanoor (approved):

PR #828, opened 2024-04-15
[RBAC] Specify type for AnsibleResourceField serializer

Dostonbek1:


dhaustein:
This fixes the tests failing on:
```
eda_api.exceptions.ApiTypeError: Invalid type for variable 'resource'. Required value type is str and passed type was dict at ['received_data']['resource']
```

dhaustein (approved):

PR #829, opened 2024-04-16
refactor: move out status management into StatusManager

Alex-Izquierdo:
Refactor the activation manager to move out the status management into a parent class "StatusManager". This is used when we want to change the status of activations in the orchestrator prior to the container livecycle. 

Jira: https://issues.redhat.com/browse/AAP-22452

### DIFF
--- a/src/aap_eda/services/activation/manager.py
+++ b/src/aap_eda/services/activation/manager.py
@@ -76,7 +76,58 @@ def run_with_lock(func: tp.Callable) -> tp.Callable:
     return _run_with_lock
 
 
-class ActivationManager:
+class StatusManager:
+    """Status Manager manages the status of a process parent.
+
+    The Status Manager is responsible for updating the status
+    of the process parent and its latest instance.
+    """
+
+    def __init__(
+        self,
+        db_instance: tp.Union[models.Activation, models.EventStream],
+    ):
+        """Initialize the Process Parent Status Manager.
+
+        Args:
+            db_instance: The database instance of the process parent.
+        """
+        self.db_instance = db_instance
+        if isinstance(db_instance, models.Activation):
+            self.db_instance_type = ProcessParentType.ACTIVATION
+        else:
+            self.db_instance_type = ProcessParentType.EVENT_STREAM
+
+    @property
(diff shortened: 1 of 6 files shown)


src/aap_eda/services/activation/manager.py
```diff
@@ -76,7 +76,58 @@ def _run_with_lock(self: HasDbInstance, *args, **kwargs):
     return _run_with_lock
 
 
-class ActivationManager:
+class StatusManager:
```
mkanoor, line 79: @Alex-Izquierdo Can we move this to a separate file? This file is already too big. It also allows us to carve out other pieces into smaller classes like the Restart Policies

Alex-Izquierdo: Hi @mkanoor I've created status_manager file and renamed manager to activation_manager. Let me know if that's fine to you.

hsong-rh (approved):

src/aap_eda/services/activation/status_manager.py
```diff
@@ -0,0 +1,104 @@
...
+from aap_eda.core.enums import ActivationStatus, ProcessParentType
+
+LOGGER = logging.getLogger(__name__)
+
+
+class HasDbInstance(tp.Protocol):
+    db_instance: tp.Any
+
+
+def run_with_lock(func: tp.Callable) -> tp.Callable:
```
mkanoor, line 32: @Alex-Izquierdo Is this generic enough to be moved into a util like db_util and shared from there for everyone to use.

Alex-Izquierdo: I don't think so, at least for now. It's bound to a "has_db_instance" type which currently is only the activation_manager. In the future we might want to change, but for now I would say it is strictly related with the activation manager.

Alex-Izquierdo: Also, This refactor is intended to be small, only to split the status manager which I need it to be able to update the activation from the orchestrator. I would prefer to consider any other changes in different PR's.

mkanoor (approved):

PR #831, opened 2024-04-17
[RBAC] Add default arg for organization foreign key fields

Dostonbek1:


benthomasson:
@Dostonbek1 where is the default organization created?

benthomasson (approved):

mkanoor (approved):

src/aap_eda/services/activation/manager.py
```diff
@@ -1053,6 +1053,7 @@ def _create_activation_instance(self):
             "name": self.db_instance.name,
             "status": ActivationStatus.STARTING,
             "git_hash": git_hash,
+            "organization": self.db_instance.organization,
```
AlanCoding, line 1056: Calling it out in a comment, this is a case where the child inherits the organization from its parent

src/aap_eda/wsapi/consumers.py
```diff
@@ -234,6 +240,7 @@ def insert_audit_rule_data(self, message: ActionMessage) -> None:
                 fired_at=message.rule_run_at,
                 job_instance_id=job_instance_id,
                 status=message.status,
+                organization=activation_org,
```
AlanCoding, line 243: second case `AuditRule` (2)

AlanCoding (changes requested):
Looking for Rulebook to inherit org from project (3)

AlanCoding (approved):

benthomasson (approved):

PR #832, opened 2024-04-17
fix: update credentials when activations is restarted or enabled

hsong-rh:
https://issues.redhat.com/browse/AAP-22972


tests/integration/api/test_activation_with_credential.py
```diff
@@ -347,6 +347,75 @@ def test_is_activation_valid_with_run_job_template_and_no_token_no_credential(
     ) in message
 
 
+@pytest.mark.django_db
+def test_is_activation_valid_with_updated_credential(
+    preseed_credential_types,
+):
+    fks = create_activation_related_data(TEST_EXTRA_VAR)
```
Dostonbek1, line 354: We should replace this data creation with fixtures from conftest.py

hsong-rh: yeah, updated.

Dostonbek1: @hsong-rh re-opening this since it is not updated.

Dostonbek1 (changes requested):

src/aap_eda/api/serializers/activation.py
```diff
@@ -383,31 +402,28 @@ def create(self, validated_data):
...
+                validated_data, vault_data
             )
 
         vault = _get_vault_credential_type()
 
         if validated_data.get("eda_credentials"):
-            system_vault_credential_needed = _handle_eda_credentials(
-                validated_data, extra_var_obj, password
+            _update_extra_vars_from_eda_credentials(
+                validated_data, vault_data, True
```
Alex-Izquierdo, line 416: Can you use named arguments as you do in other places? passing booleans as positional arguments is misleading and considered a bad practice.

tests/integration/api/test_activation_with_credential.py
hsong-rh: Updated to use fixture

tests/integration/api/test_activation_with_credential.py
```diff
@@ -347,17 +199,49 @@ def test_is_activation_valid_with_run_job_template_and_no_token_no_credential(
...
+    default_rulebook: models.Rulebook,
+    default_extra_var: models.ExtraVar,
     preseed_credential_types,
 ):
-    fks = create_activation_related_data(TEST_EXTRA_VAR)
-    test_activation = TEST_ACTIVATION.copy()
-    test_activation["decision_environment_id"] = fks["decision_environment_id"]
-    test_activation["rulebook_id"] = fks["rulebook_id"]
-    test_activation["extra_var_id"] = fks["extra_var_id"]
+    test_activation = {
```
Dostonbek1, line 239: We also have `activation_payload` fixture for POST activation, could that be used here?

Dostonbek1 (approved):

mkanoor (approved):

PR #833, opened 2024-04-18
chore: log errors when gpg import fails

bzwei:


Alex-Izquierdo (dismissed):

src/aap_eda/services/project/scm.py
```diff
@@ -254,6 +254,7 @@ def add_gpg_key(cls, key_file: str) -> None:
         child = pexpect.spawn(cmd)
         child.expect([pexpect.EOF])
         if child.exitstatus != 0:
+            logger.error(f"gpg import failed with error: {child.before}")
```
mkanoor, line 257: @bzwei do you also want to log the exit code that will also give us clues.

Alex-Izquierdo (approved):

jshimkus-rh (approved):

PR #835, opened 2024-04-18
feat: display $vaulted$ if data is vaulted in extra vars

hsong-rh:
https://issues.redhat.com/browse/AAP-22976

when UI displays extra vars with vaulted data, 

before it looks like:
image
after it looks like:
image


mkanoor (approved):

src/aap_eda/api/serializers/project.py
```diff
@@ -246,6 +249,23 @@ class Meta:
...
         read_only_fields = ["id"]
 
+    def to_representation(self, extra_var):
+        return {
+            "id": extra_var.id,
+            "extra_var": self.replace_vault_data(extra_var),
+        }
+
+    def replace_vault_data(self, extra_var):
+        data = {}
```
Alex-Izquierdo, line 259:
```suggestion
        data = {
            key: (VAULTED_STRING if isinstance(value, str) and ANSIBLE_VAULT_STRING in value else value)
            for key, value in yaml.safe_load(extra_var.extra_var).items()
        }
```

Alex-Izquierdo (approved):

src/aap_eda/api/serializers/project.py
```diff
@@ -22,6 +23,8 @@
 from aap_eda.core.utils.crypto.base import SecretValue
 
 ENCRYPTED_STRING = "$encrypted$"
+ANSIBLE_VAULT_STRING = "$ANSIBLE_VAULT"
```
bzwei, line 26: how about `"$ANSIBLE_VAULT;"` to match more characters?

src/aap_eda/api/serializers/project.py
```diff
@@ -246,6 +249,23 @@ class Meta:
...
+            "id": extra_var.id,
+            "extra_var": self.replace_vault_data(extra_var),
+        }
+
+    def replace_vault_data(self, extra_var):
+        data = {}
+        for key, value in yaml.safe_load(extra_var.extra_var).items():
+            data[key] = (
+                VAULTED_STRING
+                if isinstance(value, str) and ANSIBLE_VAULT_STRING in value
```
bzwei, line 263: it is safe to change the condition to `value.startwith(ANSIBLE_VAULT_STRING)`?

hsong-rh: No, the actual vaulted string doesn't start with `ANSIBLE_VAULT_STRING`

mkanoor:
Should we change the $vaulted$ to $encrypted$ otherwise we would have to explain what it and how different it is from encrypted strings

Alex-Izquierdo (approved):

mkanoor (approved):

bzwei (approved):

PR #838, opened 2024-04-19
chore: turn on DEBUG in docker-compse-mac

bzwei:


Alex-Izquierdo (approved):
I think you probably want DJANGO_SETTINGS_MODULE=aap_eda.settings.default

Dostonbek1 (approved):

PR #839, opened 2024-04-19
Changes to fully connect user permissions

AlanCoding:
Depends on https://github.com/ansible/django-ansible-base/pull/318 --> MERGED, released

I'm still trying to wrap up the final bits, but with this I mainly wanted to demo the integration point of `validate_is_superuser`, which can be a bit tricky to test, but this got want I wanted. There's a lot of test cleanup that could happen here, but I also don't want to re-do everything in DAB.

Might as well link AAP-21811


src/aap_eda/api/serializers/user.py
```diff
@@ -102,6 +103,12 @@ class Meta:
             "is_superuser",
         ]
 
+    def validate_is_superuser(self, value):
```
Dostonbek1, line 106: I don't see this getting called anywhere, and while testing I was able to update `superuser`'s username with a normal user, which I don't think is the expectation here. Maybe it should be renamed to `validate()` or get called in `override()` method?

Dostonbek1 (changes requested):

Dostonbek1:
@AlanCoding Running it locally and trying to get to `/api/eda/v1/docs` page somehowe errors out with the following:
```
eda-api-1                      |   File "/app/venv/lib64/python3.11/site-packages/ansible_base/rbac/api/serializers.py", line 193, in __init__
eda-api-1                      |     qs = self.Meta.model.get_field(self.actor_field).model.objects.all()
eda-api-1                      |          ^^^^^^^^^^^^^^^^^^^^^^^^^
eda-api-1                      | AttributeError: type object 'RoleTeamAssignment' has no attribute 'get_field'
```

AlanCoding:
@Dostonbek1 thanks https://github.com/ansible/django-ansible-base/pull/347

src/aap_eda/api/serializers/user.py
AlanCoding: Ah, that could be a bug in that I put this on `UserCreateUpdateSerializer` but not `CurrentUserUpdateSerializer` or the other serializers. I would prefer to put this on a base user serializer.

But the pattern of `validate_{field_name}` is a DRF pattern. I tested this for one case and it is effective. I'd like to keep this method, because I think the issue is multiple serializer.

AlanCoding: Good find, https://github.com/ansible/django-ansible-base/pull/351 tackles the root of the problem.

I'm pushing a commit here, but now that's just adding more tests for the most part and makes little functional difference.

tests/integration/api/test_user.py
```diff
@@ -11,21 +11,71 @@
...
 from aap_eda.core import models
-from aap_eda.core.enums import Action, ResourceType
 from tests.integration.constants import api_url_v1
 
 from .conftest import ADMIN_USERNAME
 
 DATETIME_FORMAT = "%Y-%m-%dT%H:%M:%S.%fZ"
 
 
+@pytest.fixture
```
Alex-Izquierdo, line 29: I think we already have fixtures that can be reused in tests/integration/api/conftest.py

AlanCoding: I went through the list and consolidated what fixtures I could.

AlanCoding:
I rebased and updated DAB so that this is hopefully fully ready now.

Alex-Izquierdo (dismissed):

Dostonbek1 (approved):

Alex-Izquierdo (approved):

PR #841, opened 2024-04-19
fix: convert extra vars to a TextField in activation and event stream

Dostonbek1:
* Remove `ExtraVar` views and model
* Remove ExtraVar object from RBAC
* Change `extra_var` field in `Activation` and `EventStream` models from FK to TextField
* Migrate existing extra var data from `ExtraVar` model
* Update tests

NOTE: This is a breaking change for the UI so they will have to make the necessary changes and merge the PRs in sync. cc @lgalis 

JIRA: [AAP-22982](https://issues.redhat.com/browse/AAP-22982) & [AAP-18485](https://issues.redhat.com/browse/AAP-18485)


Dostonbek1:
Remaining test failures are from `test_activation_with_credential.py`; I plan to resolve them once https://github.com/ansible/eda-server/pull/832 merges.

src/aap_eda/api/serializers/activation.py
```diff
@@ -126,18 +122,14 @@ def _handle_eda_credentials(
 
         data = substitute_variables(injectors, user_inputs)
 
-        if extra_var_obj:
-            existing_data = yaml.safe_load(extra_var_obj.extra_var)
+        if validated_data["extra_var"]:
+            existing_data = yaml.safe_load(validated_data["extra_var"])
```
dhaustein, line 126: Should be `extra_vars`  (plural) so it's aligned with what the rest of the code expects. I get a HTTP 500 and a KeyError when doing a POST to `/credential-types/` something like:
```json
{
    "name": "QE Secret with injector",
    "namespace": "namespace",
    "kind": "kind",
    "description": "foo",
    "inputs": {
        "fields": [
            {
                "id": "injected_secret",
                "type": "string",
                "label": "Injected Secret",
                "secret": true
            }
        ]
    },
    "injectors": {
        "extra_vars": {
            "my_msg": "bla bla"
        }
    }
}
```

dhaustein (changes requested):

dhaustein:
For the activation schema you added extra_var (singular), I changed the E2E tests accordingly but I keep wondering whether it wouldn't be better to align with the previous extra_vars (plural)

```json
"components": {
    "schemas": {
        "ActivationCreate": {
            "type": "object",
            "description": "Serializer for creating the Activation.",
            "properties": {
                "name": {
                    "type": "string"
                },
                ...
                "extra_var": { <-- HERE
                    "type": "string",
                    "nullable": true
                }
                ...
            },
            "required": [
                "decision_environment_id",
                "name",
                "rulebook_id"
            ]
        }
    }
}
```

src/aap_eda/api/serializers/event_stream.py
```diff
@@ -89,8 +89,7 @@ def _get_extra_var_and_credential_ids(validated_data: dict) -> tuple[int, int]:
         validated_data, extra_vars, encrypt_vars, password
     )
 
-    extra_var = models.ExtraVar.objects.create(extra_var=yaml.dump(extra_vars))
-    return extra_var.id, credential_id
+    return extra_vars, credential_id
```
hsong-rh, line 92: The `credential_id` is unused, can be removed

Dostonbek1:
e2e test run: https://github.com/ansible/eda-server/actions/runs/8816963799

src/aap_eda/api/serializers/activation.py
```diff
@@ -181,6 +187,19 @@ def _get_vault_credential_type():
...
+    data = {
+        key: (
+            ENCRYPTED_STRING
+            if isinstance(value, str) and ANSIBLE_VAULT_STRING in value
+            else value
+        )
+        for key, value in yaml.safe_load(extra_var).items()
+    }
+
+    return yaml.safe_dump(data).rstrip("\n")
```
hsong-rh, line 200: @Dostonbek1 Based on the extra var discussion, we need to remove `rstrip("\n")`, and fix the e2e failures in the eda-qa side if exist

src/aap_eda/core/models/activation.py
```diff
@@ -56,9 +56,7 @@ class Meta:
     rulebook = models.ForeignKey(
         "Rulebook", on_delete=models.SET_NULL, null=True
     )
-    extra_var = models.ForeignKey(
-        "ExtraVar", on_delete=models.CASCADE, null=True
-    )
+    extra_var = models.TextField(null=True)
```
hsong-rh, line 59: check if we should use JSONField here

src/aap_eda/core/models/event_stream.py
```diff
@@ -41,11 +41,7 @@ class EventStream(StatusHandlerModelMixin, ContainerableMixin, models.Model):
         on_delete=models.SET_NULL,
         null=True,
     )
-    extra_var = models.ForeignKey(
-        "ExtraVar",
-        on_delete=models.CASCADE,
-        null=True,
-    )
+    extra_var = models.TextField(null=True)
```
hsong-rh, line 44: Same thing here, should it be JSONField?

src/aap_eda/core/models/activation.py
benthomasson: Do not use JSONField here.   Keep it as text to support YAML.

src/aap_eda/api/serializers/event_stream.py
```diff
@@ -89,8 +88,7 @@ def _get_extra_var_and_credential_ids(validated_data: dict) -> tuple[int, int]:
         validated_data, extra_vars, encrypt_vars, password
     )
 
-    extra_var = models.ExtraVar.objects.create(extra_var=yaml.dump(extra_vars))
-    return extra_var.id, credential_id
+    return extra_vars
```
hsong-rh, line 91: Minor suggestion: it's better to rename the function name to remove `credential_ids`, something like `_get_extra_var`, otherwise looks good to me.

hsong-rh (approved):

src/aap_eda/api/serializers/event_stream.py
Dostonbek1: @hsong-rh Updated.

src/aap_eda/api/serializers/activation.py
```diff
@@ -104,17 +105,24 @@ def _update_k8s_service_name(validated_data: dict) -> str:
     return service_name or create_k8s_service_name(validated_data["name"])
 
 
+def _extend_extra_vars_from_credentials(
```
Alex-Izquierdo, line 108: I see this function very confusing. `credential_data` is defined as `Union[str, int, Dict, List]` but it is expected to be a dictionary. The return type is defined as `Union[str, int, Dict, List]` but it can only return a dict or a string.

hsong-rh (dismissed):

src/aap_eda/api/serializers/activation.py
Dostonbek1: updated.

hsong-rh (approved):

Alex-Izquierdo (approved):

dhaustein (approved):

PR #842, opened 2024-04-19
fix: validate the existence of association by id for creation or update

bzwei:
validations of credential_id, credential_type_id, organization_id were missing in various places.

AAP-23105: Need to validate association ids


mkanoor:
@bzwei @hsong-rh We already encountered this error due to RBAC permission check, @hsong-rh has fixed it in his PR

hsong-rh (approved):

Dostonbek1 (approved):

PR #844, opened 2024-04-22
fix: add missing default status msg for workers_offline

Alex-Izquierdo:
The implementation of the `workers offline` status always set a status_message, but a default message must be defined for consistency. It updates the test to ensure always we have a default status message per each status.

### DIFF
--- a/tests/integration/core/test_status_handler.py
+++ b/tests/integration/core/test_status_handler.py
@@ -95,19 +95,8 @@ def test_save(instance):
         == ACTIVATION_STATUS_MESSAGE_MAP[instance.status]
     )
 
-    for status in [
-        ActivationStatus.STARTING,
-        ActivationStatus.RUNNING,
-        ActivationStatus.STOPPING,
-        ActivationStatus.DELETING,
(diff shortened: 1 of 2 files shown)


jshimkus-rh (approved):

tests/integration/core/test_status_handler.py
```diff
@@ -95,19 +95,8 @@ def test_save(instance):
...
-        ActivationStatus.COMPLETED,
-        ActivationStatus.FAILED,
-        ActivationStatus.STOPPED,
-        ActivationStatus.UNRESPONSIVE,
-        ActivationStatus.ERROR,
-        ActivationStatus.PENDING,
-    ]:
-        instance.status = status
+    for status in ActivationStatus:
+        instance.status = status.value
```
hsong-rh, line 99: Do we need to call `value` here?

Alex-Izquierdo: probably don't, but it's the interface of every enum, so it's valid.

hsong-rh (approved):

PR #846, opened 2024-04-23
Test and fix related obj permission check

AlanCoding:
AAP-23288

### DIFF
--- a/tests/integration/dab_rbac/test_related_permissions.py
+++ b/tests/integration/dab_rbac/test_related_permissions.py
@@ -76,7 +76,8 @@ def test_project_credential_access(
     response = user_api_client.patch(
         url, data={"eda_credential_id": credential.pk}
     )
-    assert response.status_code == 403, response.data
+    # NOTE: the ideal would probably be 400, but still a matter of discussion
+    assert response.status_code in (403, 400), response.data
     assert "eda_credential" in str(response.data), response.data
     project.refresh_from_db()
     assert project.eda_credential_id != credential.pk
--- a/tests/integration/dab_rbac/test_related_permissions.py
+++ b/tests/integration/dab_rbac/test_related_permissions.py
@@ -63,13 +63,26 @@ def test_project_credential_access(
     credential = cls_factory.create(models.EdaCredential)
     assert project.eda_credential_id != credential.pk  # sanity
 
+    # User can not view provided credential
     response = user_api_client.patch(
         url, data={"eda_credential_id": credential.pk}
     )
     assert response.status_code == 400, response.data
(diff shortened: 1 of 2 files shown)


tests/integration/dab_rbac/test_related_permissions.py
```diff
@@ -63,13 +63,26 @@ def test_project_credential_access(
     credential = cls_factory.create(models.EdaCredential)
     assert project.eda_credential_id != credential.pk  # sanity
 
+    # User can not view provided credential
     response = user_api_client.patch(
         url, data={"eda_credential_id": credential.pk}
     )
     assert response.status_code == 400, response.data
```
bzwei, line 70: after #842 is merged, we need to remove line 54 to enable this test, and the status_code here is 403.

AlanCoding: A 403 would disclose that a credential with id `credential.pk` exists, and the requesting user doesn't have the right to know this. It's pedantic, but that's the argument.

Alex-Izquierdo: Would not be more appropriate 404?

AlanCoding: There are 2 objects involved, for this request like PATCH `/api/eda/v1/projects/3/` with `{"eda_credential_id": 4}`

1. If user can't see project pk=3 --> 404
2. If EDA credential pk=4 actually doesn't exist, then that's not up to us, behavior already exists --> 400 w/ message that the credential doesn't exist
3. If user can't see credential, then the technical and pedantic answer is that they get a response which is indistinguishable from (2), due to the principle of not disclosing existence of objects user doesn't have view permission for. This is the same philosophy at work as in (1).

tests/integration/dab_rbac/test_related_permissions.py
```diff
@@ -63,13 +63,26 @@ def test_project_credential_access(
     credential = cls_factory.create(models.EdaCredential)
     assert project.eda_credential_id != credential.pk  # sanity
 
+    # User can not view provided credential
```
Dostonbek1, line 66: we should re-enable this test, it is skipping it on line 54.

Dostonbek1 (changes requested):

AlanCoding:
Somewhat related https://github.com/ansible/django-ansible-base/pull/346

Dostonbek1 (approved):
verified this resolves the outstanding bug:

Screenshot 2024-04-25 at 12 52 24 PM

bzwei (approved):

PR #848, opened 2024-04-23
fix: replace pexpect in scm with subprocess for better exit code detection

bzwei:


Alex-Izquierdo (approved):
When I read the title I thought you would go to change in all the places where pexpect is not needed. Do you expect to do it in later iterations?

bzwei:
> When I read the title I thought you would go to change in all the places where pexpect is not needed. Do you expect to do it in later iterations?

I updated the title. It applies only to scm.py.

mkanoor (approved):

PR #849, opened 2024-04-23
[ON-HOLD]: check unresponsive activations in monitor

Alex-Izquierdo:
Since the check for unresponsive activations doesn't require necessarily connection with podman, it can be executed by any worker and can be used outside the regular monitor flow to handle activations that might be dead in a unhealthy node, so we can restart them automatically. The heartbeat of the rulebook is the only mechanism that we have to be sure that an activation is dead and it can be restarted automatically. 

Includes:
- Execution of the unresponsiveness check for activations in "workers offline" state. 
- A refactor of the monitor_process function, encapsulating the three stages. (two existing one + a new one for unresponsive activations)
- Tests
- Move some fixtures in upper scope for reusability

Jira: https://issues.redhat.com/browse/AAP-22422


Alex-Izquierdo:
The errors in the e2e test suite are not related with this PR.

mkanoor:
@Alex-Izquierdo The heartbeat depends on a healthy Daphne websocket worker, if that dies will it lead to non-responsive behavior.

tools/docker/docker-compose-dev.yaml
```diff
@@ -25,6 +25,8 @@ x-environment: &common-env
   EDA_RULEBOOK_WORKER_QUEUES: "activation-node1,activation-node2"
   EDA_ANSIBLE_BASE_JWT_VALIDATE_CERT: ${EDA_ANSIBLE_BASE_JWT_VALIDATE_CERT:-False}
   EDA_ANSIBLE_BASE_JWT_KEY: ${EDA_ANSIBLE_BASE_JWT_KEY:-'https://localhost'}
+  EDA_RULEBOOK_LIVENESS_CHECK_SECONDS: ${EDA_RULEBOOK_LIVENESS_CHECK_SECONDS:-10}
+  EDA_RULEBOOK_LIVENESS_TIMEOUT_SECONDS: ${EDA_RULEBOOK_LIVENESS_TIMEOUT_SECONDS:-5}
```
jshimkus-rh, line 29: This inverts the default relationship between EDA_RULEBOOK_LIVENESS_CHECK_SECONDS and EDA_RULEBOOK_LIVENESS_TIMEOUT_SECONDS where the latter is larger than the former if neither are specified.

src/aap_eda/tasks/orchestrator.py
```diff
@@ -387,31 +389,95 @@ def monitor_rulebook_processes() -> None:
...
+            f"Checking unresponsiveness for process id {process.id} of ",
+            f"{process.parent_type} {process_parent.id}",
+        )
+        # we expect connection errors here, we are monitoring
+        # processes that can not be monitored by activation workers
+        with contextlib.suppress(engine_exceptions.ContainerEngineInitError):
+            try:
+                _check_unresponsive_process(process_parent)
+            except engine_exceptions.ContainerEngineInitError:
+                raise
```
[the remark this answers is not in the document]
Alex-Izquierdo: Agree, thanks for catching. changed.

tools/docker/docker-compose-dev.yaml
Alex-Izquierdo: What it's your point, anyway? The default values are intended for production and and provide some protection in cases with high workload. That's not the case of the development environment, where we want to detect it quickly and we don't need high timeouts.

jshimkus-rh: It's not about the values; smaller ones for development is reasonable.
It's that the default relation between the two is opposite the default implemented for deployment and is thus "odd" without explanation.  A comment explaining would benefit those, particularly the unfamiliar, who recognize the oddity and are left wondering "Why?"

Alex-Izquierdo:
on-hold until further discussions. Doubts remain as to whether the unresponsiveness check is sufficient to automatically restart the activation.

PR #850, opened 2024-04-23
fix: move validated error message under the right field

hsong-rh:
https://issues.redhat.com/browse/AAP-23311

The missing passphrase key error message is moved from the `SCM Private Key` field to the `Private Key Passphrase` field:

image

### DIFF
--- a/src/aap_eda/core/utils/credentials.py
+++ b/src/aap_eda/core/utils/credentials.py
@@ -25,6 +25,9 @@ from aap_eda.core.utils.awx import validate_ssh_private_key
 ENCRYPTED_STRING = "$encrypted$"
 EDA_PREFIX = "EDA_"
 SUPPORTED_KEYS_IN_INJECTORS = ["extra_vars"]
+PROTECTED_PASSPHRASE_ERROR = (
+    "The key is passphrase protected, please provide passphrase."
+)
 
 
 class InjectorMissingKeyException(Exception):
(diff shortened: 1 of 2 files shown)


src/aap_eda/core/utils/credentials.py
```diff
@@ -108,7 +111,10 @@ def validate_inputs(schema: dict, inputs: dict) -> dict:
                 inputs=inputs,
             )
             if bool(result):
-                errors[display_field] = result
+                if PROTECTED_PASSPHRASE_ERROR in result:
+                    errors["inputs.ssh_key_unlock"] = result
```
mkanoor, line 115: @hsong-rh do we have to check if the customer provided inputs has the ssh_key_unlock? Or would that validation kick in only when both keys are present.

hsong-rh: @mkanoor Yes, we check them at https://github.com/ansible/eda-server/blob/main/src/aap_eda/core/utils/credentials.py#L332

bzwei (approved):

Alex-Izquierdo (approved):

src/aap_eda/core/utils/credentials.py
hsong-rh: @mkanoor We will handle your scenario later: create a credential type without `Private Key Passphrase` field. In that case, we may show a warning message.

mkanoor (approved):

PR #851, opened 2024-04-23
fix: clean up dispatch queue log entries

jshimkus-rh:


Alex-Izquierdo (approved):

src/aap_eda/tasks/orchestrator.py
```diff
@@ -178,50 +178,61 @@ def dispatch(
...
+            LOGGER.warning(
+                f"Restarting {process_parent_type} {process_parent_id} "
+                "on the least busy queue; The workers of its associated queue "
+                f"'{queue_name}' are failing liveness checks. "
+                "There may be an issue with the node; please contact "
+                "the administrator.",
             )
-            LOGGER.warning(msg)
-            return
+            queue_name = get_least_busy_queue_name()
```
hsong-rh, line 235: @jshimkus-rh This is the newly added line. Do we need it if the queue is checked to be unhealthy? or this is the fix: we continue for the `restart` type?

Alex-Izquierdo: This just a little refactor and it splits the message. There is around here another bug that will be addressed later. https://issues.redhat.com/browse/AAP-23378

jshimkus-rh: @hsong-rh Sorry, I didn't notice your question until Alejandro responded.

The functionality remains the same.  The flow got changed to have only one use of check_rulebook_queue_health which resulted in swapping the previous clauses with the restart case being the non-exceptional processing in the case of a non-healthy queue.

hsong-rh (approved):

PR #852, opened 2024-04-23
feat: remove credentials api

bzwei:
AAP-23219: remove old credentials endpoints


Alex-Izquierdo (dismissed):

Dostonbek1 (dismissed):

src/aap_eda/api/serializers/__init__.py
```diff
@@ -126,9 +121,6 @@
     "CredentialCreateSerializer",
     "CredentialRefSerializer",
     # credential type
-    "CredentialTypeSerializer",
```
mkanoor, line 129: @bzwei I don't think this can be removed it was added in this PR https://issues.redhat.com/browse/AAP-22997

hsong-rh: Yes, we need these serializers

mkanoor (changes requested):
CredentialType cannot be removed its needed. Please remember that the CredentialType and EdaCredentials are new and Credential is old. The old Credential didn't have a CredentialType

mkanoor (approved):

hsong-rh (approved):

PR #853, opened 2024-04-24
chore: add a test case for filter on eda credentials

hsong-rh:
https://issues.redhat.com/browse/AAP-23330

The issue comes from a typo on the UI side, here just add a test case.

### DIFF
--- a/tests/integration/api/test_eda_credential.py
+++ b/tests/integration/api/test_eda_credential.py
@@ -134,6 +134,13 @@ def test_list_eda_credentials_with_kind_filter(
     )
     assert len(response.data["results"]) == 2
 
+    name_prefix = default_eda_credential.name[0]
+    response = client.get(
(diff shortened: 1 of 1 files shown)


mkanoor (approved):

tests/integration/api/test_eda_credential.py
```diff
@@ -134,6 +134,12 @@ def test_list_eda_credentials_with_kind_filter(
     )
     assert len(response.data["results"]) == 2
 
+    response = client.get(
+        f"{api_url_v1}/eda-credentials/?credential_type__kind=scm"
+        "&credential_type__kind=registry&name=m",
```
mkanoor, line 139: @hsong-rh instead of m can we use a name_prefix and set it to the first character of the credential name

mkanoor (approved):

bzwei (approved):

PR #854, opened 2024-04-24
chore: add summary_fields for permission based models

Dostonbek1:
This PR adds `summary_fields` method to models that are part of RBAC as it is required for the new RBAC UI to display `name` of the resource in roles assignments page.

JIRA: [AAP-23305](https://issues.redhat.com/browse/AAP-23305)


src/aap_eda/core/models/base.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class BaseModel(models.Model):
+    class Meta:
+        abstract = True
+
+    def summary_fields(self):
+        return {
+            "id": self.id,
+            "name": self.name,
```
AlanCoding, line 27: It looks like it would be _viable_ to also consolidate

```python
name = models.TextField(null=False, unique=True)
```

Meaning, delete these fields from all subclasses of `BaseModel`. That is roughly approaching how AWX & DAB and other things did this. Instead of just a base model, it becomes a NamedModel, and that might start becoming familiar if you've seen the inheritance structure in other projects...

Doing this, you don't have to ask yourself the question "does this model have a name field".

I realize this won't work for some models now because they are exceptions. It would be good to actually dive into what drove those. Like, does `RulebookProcess` not have _any_ unique constraints? Is that really intended? For those models, maybe you should define `summary_fields` in the model instead of inheriting from the base.

AlanCoding (approved):
Nice patterns evolving. I'm also pretty confident these `summary_fields` methods will solve the issue. I could help add a test if you wanted to confirm that.

src/aap_eda/core/models/base.py
Dostonbek1: @AlanCoding Updated the PR. I left out the models that have `name` field defined differently. `Rulebook`, etc. have `unique_together` constraints for cases like two projects could have a rulebook with the same name, so we have to do `unique_together = ["project_id", "name"]`. As for `RulebookProcess`, it inherits name from its parent `Activation` so all RulebookProcess objects relating to one Activation all have the same name.

hsong-rh (approved):

Alex-Izquierdo (approved):

PR #856, opened 2024-04-24
fix: decrypt private key file with passphrase

bzwei:
Remove the unnecessary quotes for -N argument.

AAP-23334: Cloning of private project now fails when using private key after change in eda-server using pexpect


src/aap_eda/services/project/scm.py
```diff
@@ -243,7 +243,7 @@ def build_url(
     @classmethod
     def decrypt_key_file(cls, key_file: str, password: str) -> None:
         result = subprocess.run(
-            [KEYGEN_COMMAND, "-p", "-P", password, "-N", '""', "-f", key_file],
+            [KEYGEN_COMMAND, "-p", "-P", password, "-N", "", "-f", key_file],
```
hsong-rh, line 246: @bzwei `[-N new_passphrase] ` is an optional parameter, do you have to set it to an empty string?

hsong-rh (approved):

mkanoor (approved):

Alex-Izquierdo (approved):

PR #860, opened 2024-04-25
fix: validate credential's inputs when it is empty

hsong-rh:
https://issues.redhat.com/browse/AAP-23402


Alex-Izquierdo (approved):

mkanoor (approved):

Alex-Izquierdo (commented):
Seems that one e2e test is failing, but I'm not sure if its related

hsong-rh:
> Seems that one e2e test is failing, but I'm not sure if its related

it's strange, shouldn't related, let me check

PR #867, opened 2024-04-29
fix: credential deletion when it's attached project as GPG type

hsong-rh:
https://issues.redhat.com/browse/AAP-23498

### DIFF
--- a/tests/integration/api/test_eda_credential.py
+++ b/tests/integration/api/test_eda_credential.py
@@ -395,6 +395,41 @@ def test_delete_credential_used_by_activation_forced(
     assert default_activation.eda_credentials.count() == 0
 
 
+@pytest.mark.django_db
+def test_delete_credential_used_by_project_with_gpg_credential(
+    client: APIClient,
+    preseed_credential_types,
+):
+    gpg_credential_type = models.CredentialType.objects.get(
+        name=enums.DefaultCredentialType.GPG
+    )
+    eda_credential = models.EdaCredential.objects.create(
+        name="test_gpg_credential",
+        inputs={"gpg_public_key": "secret"},
+        credential_type=gpg_credential_type,
+    )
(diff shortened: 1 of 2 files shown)


tests/integration/api/test_eda_credential.py
```diff
@@ -395,6 +395,41 @@ def test_delete_credential_used_by_activation_forced(
...
+        scm_branch="main",
+        proxy="http://user:secret@myproxy.com",
+        import_state=models.Project.ImportState.COMPLETED,
+        import_task_id="c8a7a0e3-05e7-4376-831a-6b8af80107bd",
+    )
+    response = client.delete(
+        f"{api_url_v1}/eda-credentials/{eda_credential.id}/",
+    )
+
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
```
mkanoor, line 427: @hsong-rh Shouldn't this be returning a 409 like in test test_delete_credential_with_project_reference
Also the managed=True is not valid since this is a user created Credential, they should be able to delete it if there are no references to it.

tests/integration/api/test_eda_credential.py
```diff
@@ -395,6 +395,41 @@ def test_delete_credential_used_by_activation_forced(
...
+    preseed_credential_types,
+):
+    gpg_credential_type = models.CredentialType.objects.get(
+        name=enums.DefaultCredentialType.GPG
+    )
+    eda_credential = models.EdaCredential.objects.create(
+        name="test_gpg_credential",
+        inputs={"gpg_public_key": "secret"},
+        credential_type=gpg_credential_type,
+        managed=True,
```
mkanoor, line 410: @hsong-rh This should not have the managed=True this is a user defined EdaCredential.

bzwei (approved):

mkanoor (approved):

PR #870, opened 2024-04-29
fix: validate and raise errors if required fields are empty

hsong-rh:
https://issues.redhat.com/browse/AAP-23402


src/aap_eda/core/utils/credentials.py
```diff
@@ -102,6 +102,10 @@ def validate_inputs(schema: dict, inputs: dict) -> dict:
             if required and not default:
                 errors[display_field] = ["Cannot be blank"]
                 continue
+        else:
+            if required and len(user_input) == 0:
```
bzwei, line 106: @hsong-rh do we want to trim white spaces for user_input? We also need to consider the case when the data is a boolean rather than string.

bzwei (approved):

mkanoor (approved):

PR #871, opened 2024-04-30
fix: update organizations/<id>/teams/ endpoint method

Dostonbek1:
* update `organizations/<id>/teams/` method to properly check for organization existence and raising exception 

JIRA: [AAP-23508](https://issues.redhat.com/browse/AAP-23508)

### DIFF
--- a/src/aap_eda/api/views/organization.py
+++ b/src/aap_eda/api/views/organization.py
@@ -26,6 +26,7 @@ from rest_framework.response import Response
 
 from aap_eda.api import exceptions as api_exc, filters, serializers
 from aap_eda.core import models
+from aap_eda.core.enums import Action
 
 from .mixins import PartialUpdateOnlyModelMixin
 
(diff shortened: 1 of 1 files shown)


src/aap_eda/api/views/organization.py
```diff
@@ -123,15 +124,24 @@ def destroy(self, request, *args, **kwargs):
         ],
     )
     @action(
-        detail=True,
+        detail=False,
```
bzwei, line 127: why do we need to turn it to False?

Dostonbek1: `detail=False` is for when we are dealing with a collection of instances (teams) rather than a single instance; additionally, it also helps us use Team specific queryset and filterset on organization viewset. This follows the pattern that we have for existing endpoints, such as `activations/<id>/instances/`.

bzwei (approved):

dhaustein:
@Dostonbek1 I gave this a look locally and looks good, thanks!

hsong-rh (approved):

PR #876, opened 2024-05-01
fix: [AAP-23586] allow blank activation text fields

jshimkus-rh:
https://issues.redhat.com/browse/AAP-23586


Dostonbek1:
Should `extra_var` field also allow empty string or are we not allowing that? @lgalis

jshimkus-rh:
@doston I took the comments in slack yesterday to indicate you and @lgalis were already on it.  I can add it.

Dostonbek1:
@jshimkus-rh It is a valid point that extra var might be a bit complex to just allow empty strings, that's why it was more of a question about if the UI should disallow passing empty string or the API allows empty string as valid input, although I think this is very trivial. @mkanoor any thoughts on what should be the best approach?

Alex-Izquierdo (approved):

Alex-Izquierdo:
We should consider to not use "nullable" for char fields. https://docs.astral.sh/ruff/rules/django-nullable-model-string-field/

Dostonbek1:
> We should consider to not use "nullable" for char fields. https://docs.astral.sh/ruff/rules/django-nullable-model-string-field/

Does this mean the default in models should also be set to empty string, or is it enough to set blank=True and the request payload sends empty string?

mkanoor:
@jshimkus-rh Should we be swapping allow_null with allow_blank, do we need both of them in there?

Alex-Izquierdo:
> > We should consider to not use "nullable" for char fields. https://docs.astral.sh/ruff/rules/django-nullable-model-string-field/
> 
> Does this mean the default in models should also be set to empty string, or is it enough to set blank=True and the request payload sends empty string?

Yeah, ideally all our models should be without "null=true" and just use an empty string as default. I'm not saying should be done in this PR, totally out of scope. I guess as it is now is fine

At least, starting from now, we could not use null for string-based fields.

Dostonbek1 (approved):
Based on @Alex-Izquierdo 's last comment, this looks good to me.

PR #880, opened 2024-05-02
fix: handle HealthyQueueNotFoundError exception

jshimkus-rh:
This addresses the 500 error return described in AAP-23378 allowing the activation state to be, as expected, "workers offline".

### DIFF
--- a/src/aap_eda/tasks/orchestrator.py
+++ b/src/aap_eda/tasks/orchestrator.py
@@ -158,9 +158,10 @@ def dispatch(
     job_id = _manage_process_job_id(process_parent_type, process_parent_id)
     LOGGER.info(
         f"Dispatching request type {request_type} for {process_parent_type} "
-        "{process_parent_id}",
+        f"{process_parent_id}",
     )
 
+    queue_name = None
     # new processes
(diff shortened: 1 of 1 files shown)


Alex-Izquierdo:
Hi @jshimkus-rh I think this PR needs a rebase, it seems to include code unrelated.

src/aap_eda/tasks/orchestrator.py
```diff
@@ -156,26 +156,27 @@ def dispatch(
     process_parent_id: int,
     request_type: Optional[ActivationRequest],
 ):
+    request_type_text = f" type {request_type}" if request_type else ""
```
Alex-Izquierdo, line 159: I would say we don't need this after https://github.com/ansible/eda-server/pull/863

Alex-Izquierdo (approved):

src/aap_eda/tasks/orchestrator.py
jshimkus-rh: It's needed to avoid having `type None` appear in the message text.

eclarizio (approved):

PR #882, opened 2024-05-02
feat: add validation on credentials to make sure their types are valid during creating/updating of activations, projects or decision_environments

hsong-rh:
https://issues.redhat.com/browse/AAP-23583

### DIFF
--- a/src/aap_eda/api/serializers/activation.py
+++ b/src/aap_eda/api/serializers/activation.py
@@ -403,6 +403,7 @@ class ActivationCreateSerializer(serializers.ModelSerializer):
         required=False,
         allow_null=True,
         child=serializers.IntegerField(),
+        validators=[validators.check_multiple_credentials],
     )
     k8s_service_name = serializers.CharField(
         required=False,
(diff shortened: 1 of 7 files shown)


Dostonbek1:
> @Dostonbek1 Will your pr #879 fix these CI errors? They were because incorrect credentials were randomly chosen.

Yes, they should as long as the fixtures for these objects are correctly set up and have the right credentials assigned because my changes depend on existing fixtures to generate post data for all the objects in RBAC.

src/aap_eda/api/serializers/activation.py
```diff
@@ -403,6 +403,7 @@ class Meta:
         required=False,
         allow_null=True,
         child=serializers.IntegerField(),
+        validators=[validators.check_multiple_credentials],
```
bzwei, line 406: To align with how you name other validator methods, this one may be `validators.check_credential_types_for_activation`

hsong-rh: The name has been used already: https://github.com/ansible/eda-server/pull/882/files#diff-0fc6d501fb19fb05efa10f75a58fc762f5aec1ecf8ecd80fa488d619f346b626R70

bzwei (approved):

mkanoor (approved):

PR #890, opened 2024-05-06
fix: validate inputs/injectors when creating/updating credential types

hsong-rh:
https://issues.redhat.com/browse/AAP-23687

Raise 400 error if inputs/injectors payload are empty when creating/updating credential types.

### DIFF
--- a/src/aap_eda/api/serializers/credential_type.py
+++ b/src/aap_eda/api/serializers/credential_type.py
@@ -53,10 +53,36 @@ class CredentialTypeCreateSerializer(serializers.ModelSerializer):
     )
 
     def validate(self, data):
-        if data.get("injectors") and data.get("inputs"):
-            errors = validate_injectors(
-                data.get("inputs"), data.get("injectors")
-            )
+        injectors = data.get("injectors")
+        inputs = data.get("inputs")
+
+        if self.partial:
+            # updating allow injectors/inputs be None, but not empty dict
+            if injectors is not None and not bool(injectors):
+                raise serializers.ValidationError(
+                    "Injectors field cannot be empty"
+                )
+
+            if inputs is not None and not bool(inputs):
+                raise serializers.ValidationError(
+                    "inputs field cannot be empty"
+                )
+
+            injectors = injectors or self.instance.injectors
+            inputs = inputs or self.instance.inputs
+        else:
+            if not bool(injectors):
+                raise serializers.ValidationError(
+                    "Injectors field cannot be empty"
+                )
+
+            if not bool(inputs):
+                raise serializers.ValidationError(
+                    "Inputs field cannot be empty"
+                )
+
+        if injectors and inputs:
+            errors = validate_injectors(inputs, injectors)
             if bool(errors):
                 raise serializers.ValidationError(errors)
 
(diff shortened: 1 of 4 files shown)


src/aap_eda/api/serializers/credential_type.py
```diff
@@ -53,10 +53,36 @@ class CredentialTypeCreateSerializer(serializers.ModelSerializer):
     )
 
     def validate(self, data):
-        if data.get("injectors") and data.get("inputs"):
-            errors = validate_injectors(
-                data.get("inputs"), data.get("injectors")
-            )
+        injectors = data.get("injectors")
```
mkanoor, line 56: @hsong-rh In the CredentialTypeCreateSerializer inputs is marked as required=True should we have the same for injectors.
https://github.com/ansible/eda-server/blob/25deac499e99af1f892d9b7927e76dd7a6b84363/src/aap_eda/api/serializers/credential_type.py#L43

tests/integration/conftest.py
```diff
@@ -721,22 +721,9 @@ def default_activation_instance_logs(
...
-        },
-        {
-            "id": "ssh_key_unlock",
-            "label": "Private Key Passphrase",
-            "type": "string",
-            "secret": True,
-        },
     ]
 }
+INJECTORS = {"extra_vars": {"username": "Fred"}}
```
bzwei, line 726: `{"extra_vars": {"password": "{{ password }}"}}` to better simulate the real use

src/aap_eda/api/serializers/credential_type.py
```diff
@@ -53,10 +53,36 @@ class CredentialTypeCreateSerializer(serializers.ModelSerializer):
...
+        if self.partial:
+            # updating allow injectors/inputs be None, but not empty dict
+            if injectors is not None and not bool(injectors):
+                raise serializers.ValidationError(
+                    "Injectors field cannot be empty"
+                )
+
+            if inputs is not None and not bool(inputs):
+                raise serializers.ValidationError(
+                    "inputs field cannot be empty"
```
bzwei, line 68: Capitalize `Inputs` to match other error messages, or change to lower case for all messages.

src/aap_eda/api/serializers/credential_type.py
hsong-rh: @mkanoor for updating credential types, `inputs` or `injectors` should mark as required=False. we need to define a separate updating serializer.

tests/integration/conftest.py
hsong-rh: The RBAC autogenerated test data don't have a defined `password`, which will cause tests to fail.

mkanoor (approved):

tests/integration/conftest.py
hsong-rh: updated

src/aap_eda/core/utils/credentials.py
```diff
@@ -236,13 +236,9 @@ def validate_schema(schema: dict) -> list[str]:
...
 def validate_injectors(schema: dict, injectors: dict) -> dict:
     errors = []
-    context = _default_context(schema)
 
     if not isinstance(injectors, dict):
-        errors.append(
-            "Injectors must be a dict type and defines keys"
-            f" in {SUPPORTED_KEYS_IN_INJECTORS}"
-        )
+        errors.append("Injectors must be in dict format")
```
bzwei, line 241: This error will be shown to UI/API users, but dict is a python term.

hsong-rh: OK, let me change `dict` into `Key-Value pair`

bzwei (approved):

Dostonbek1 (approved):

PR #892, opened 2024-05-06
WIP: refactor: reduce code duplication in request processing

jshimkus-rh:
This provides the agreed upon usage of PENDING w/ status text containing the specific message:

    The workers available to process request type {request_type} for
    {process_parent_type} {process_parent_id} are failing liveness checks.
    There may be an issue with the node; please contact the administrator.

as distinguishing a request that cannot be run due to complete lack of workers.

The message text above is slightly modified from that documented in AAP-23378 for consistency with other log messages from the processing code.

Orchestrator log messages were updated due to previous change that specifically set the request type if none was specified.


src/aap_eda/tasks/orchestrator.py
```diff
@@ -230,31 +229,45 @@ def dispatch(
...
+                    ActivationRequest.START,
+                    ActivationRequest.AUTO_START,
+                ]
+            ) or (
+                status_manager.db_instance.status
+                in [
+                    ActivationStatus.PENDING,
+                    ActivationStatus.PENDING_WORKERS_OFFLINE,
+                ]
+            ):
```
eclarizio, line 264: This is kind of nitpicky and my python idioms aren't great so take my suggestion with a grain of salt, but would it make sense to wrap this conditional in a private method with a simple name? (Or is this ugly because of the need to pass in the parameters?)
```python
if __request_is_not_currently_running(request_type, status_manager.db_instance.status):
```

jshimkus-rh:
> To me that message is contradictory - it implies there are workers available, when in fact they are not responding to liveliness checks, which in essence means they are unavailable.

@ttuffin Updated per our slack convo.

src/aap_eda/tasks/orchestrator.py
jshimkus-rh: We've got two independent conditions and we're manipulating an external object.  I don't think it's significantly better/worse either way; we're still going to want to explain why those conditions (as per the comment) are significant.

However, moving the whole thing (such that status_manager doesn't even appear in this location) to a separate method would be reasonable.

src/aap_eda/core/enums.py
```diff
@@ -70,6 +70,7 @@ class ActivationStatus(DjangoStrEnum):
     STARTING = "starting"
     RUNNING = "running"
     PENDING = "pending"
+    PENDING_WORKERS_OFFLINE = "pending (workers offline)"
```
Alex-Izquierdo, line 73: I think we need to discuss this. The last agreement was to use a regular "pending" with a different status message. I think a new pending status is redundant and it would require changes in the UI as well.  
I may missing something but I think we don't need a new pending status.

Alex-Izquierdo (changes requested):

jshimkus-rh:
All outstanding issues and critiques have been addressed.

Alex-Izquierdo:
HI @jshimkus-rh I'm finally covered this with https://github.com/ansible/eda-server/pull/894

jshimkus-rh:
> HI @jshimkus-rh I'm finally covered this with #894

True.  However, I saw that as a stopgap to fixing the regression from #880.

jshimkus-rh:
Converted to draft.  To be revisited post next release.

PR #893, opened 2024-05-07
fix: add custom permission for project sync action

Dostonbek1:
* add a custom "sync_project" permission for project sync action

JIRA: [AAP-23739](https://issues.redhat.com/browse/AAP-23739)


src/aap_eda/api/views/project.py
```diff
@@ -196,14 +197,18 @@ def partial_update(self, request, pk):
...
-        except models.Project.DoesNotExist:
-            raise api_exc.NotFound
+        # we have to use filter_queryset here instead of get_object in order
+        # to use select_for_update to lock project instance to update
+        # TODO: check how important is using select_for_update
+        queryset = self.filter_queryset(
+            self.get_queryset().select_for_update()
+        )
+        project = get_object_or_404(queryset, id=pk)
+        self.check_object_permissions(request, project)
```
Dostonbek1, line 211: @Alex-Izquierdo @AlanCoding I'm not sure how important using `select_for_update()` method here is, if it could be removed then these 3 lines could be replaced with `get_object()`.

AlanCoding: It is almost certainly due to the `Conflict` exception below. The usage looks generally correct, since this is a function in a transaction. It loads `import_state` after entering, checks the data for some condition, and then sets `import_state` after starting. This is how you would be 100% sure 2 things didn't run at the same time.

You generally don't need to lump permission checks together with that, and `select_for_update` should be as limited in scope as humanly possible. I might suggest moving `transaction.atomic` from a decorator to a context manager after the permission checks. Then do `project.refresh_from_db(fields=['import_state'])` and the conflict check, and the save.

Alex-Izquierdo: I would say the lock is necessary to be sure there are no race conditions with rq jobs.

Dostonbek1:
e2e test run with updated tests: https://github.com/ansible/eda-server/actions/runs/8987218871

src/aap_eda/api/views/project.py
Dostonbek1: @Alex-Izquierdo Pushed new changes that does correct permission checking without breaking the existing implementation as much as I can. @AlanCoding looks like we still want to keep `select_for_update` in order to lock the rows, and thus we cannot move the transaction to a context manager, because `select_for_update` has to be inside a transaction. Please let me know if I missed or misunderstood something.

Alex-Izquierdo (dismissed):

AlanCoding (changes requested):
Intent of select_for_update should be respected. Use of DRF `check_object_permissions` is very nice, but, strikes me as non-trivial enough that we should have a test for having / not-having the sync_project permission.

AlanCoding (approved):
Everything looks great now, thanks!

Alex-Izquierdo (dismissed):

bzwei (approved):

Alex-Izquierdo (approved):

PR #894, opened 2024-05-07
Handle HealthyQueueNotFoundError for new processes. 

Alex-Izquierdo:
- Handle HealthyQueueNotFoundError when a new process is going to be created (restart, start, autostart) 
- Update the status message when an activation is pending because there are no healthy nodes.
- Fixes a regresion where an activation running in a unhealthy node is scheduled automatically in a new node instead of set "workers-offline" status
- Fixes a bug where we don't check the health of the queue if only one queue (default queue for k8s and single-node deployments) was set. 

Test flow:
1. create an activation until reach running state. 
2. kill the activation workers of the node where the activation is running (from now node A), the activation goes into workers-offline state
3. (optional) relaunch workers to test recoverability
4. restart activation, it should be running in node B
5. kill workers of node B, activation goes into workers-offline. 
6. Create a new activation, it should go into pending state with proper status message
7. Restart the first activation in workers offline state, it should go into pending state with proper status message.
8. (optional) relaunch some activation worker to test that both activations runs. 

Replaces https://github.com/ansible/eda-server/pull/892
Fixes: https://issues.redhat.com/browse/AAP-23378 and https://issues.redhat.com/browse/AAP-22907


src/aap_eda/tasks/orchestrator.py
```diff
@@ -191,79 +209,96 @@ def dispatch(
         ):
             if not queue_name:
                 LOGGER.info(
-                    "Scheduling request"
-                    f"{request_type_text} for {process_parent_type}"
-                    f" {process_parent_id} to the least busy queue"
-                    "; it is not currently associated with a queue.",
+                    f"Scheduling {process_parent_type} {process_parent_id} "
+                    "to the least busy queue; it is not currently associated "
+                    " with a queue.",
```
jshimkus-rh, line 214: I don't think simply dropping the rewording, in this and other messages, is a necessity.  Yes, 863 eliminated the need for the `request_type_text` but including the request type in the message is helpful.  The message changes in 892, in light of 863, should be considered.

Alex-Izquierdo: I did not rewording, the differences are because I didn't work from the existing code but from the a reverted commit, so the messages are from the previous commit. I was more focused in the logic itself.  I can recover the messages as they are, no problem to me.

Alex-Izquierdo: messages updated, please review.

Alex-Izquierdo:
@jshimkus-rh It was in WIP because it was not finished as it proves the CI fails. The description contains the details. It is ready now.

jshimkus-rh (approved):

eclarizio (approved):

src/aap_eda/tasks/orchestrator.py
```diff
@@ -191,86 +217,106 @@ def dispatch(
...
-            " the administrator."
-        )
-        LOGGER.warning(
-            "Request"
-            f"{request_type_text} for {process_parent_type}"
-            f" {process_parent_id} will be run on the least busy queue."
-        )
-        queue_name = None
+            try:
+                queue_name = get_least_busy_queue_name()
```
bzwei, line 234: This is the case when request type is not start or auto-start, which must be stop/restart/monitor. Now we cannot find its original queue. Can these operations be successfully proceeded by a new queue (node)?
In the below logic where queue is valid but not healthy, you don't proceed the such operations by another queue unless it is restart.

src/aap_eda/tasks/orchestrator.py
```diff
@@ -191,86 +217,106 @@ def dispatch(
...
+                "the administrator.",
             )
-
-            if request_type not in [
-                ActivationRequest.RESTART,
-            ]:
-                status_manager = StatusManager(
-                    get_process_parent(process_parent_type, process_parent_id),
+            try:
+                queue_name = get_least_busy_queue_name()
```
bzwei, line 294: Here you're trying another queue to proceed restart. A restart needs to stop the old activation instance first. Can it be successfully stopped by another queue?

src/aap_eda/tasks/orchestrator.py
Alex-Izquierdo: This block is to cover the case where there is a change in the settings for the queues. For instance, where a node is removed or when we have a new deployment created from backup/restore (customers do that to promote a staging env to a production or to create a test env from production). In a regular scenario, the activation is always related to a queue (expected node), so we reschedule it only when it is explicitly restarted, because the node associated might still be running the activation, so we don't to duplicate it unless the user/admin makes sure it can be restarted.

src/aap_eda/tasks/orchestrator.py
Alex-Izquierdo: No. This is an exceptional case required as workaround for the incapacity to restart an activation that is running in a unhealthy node. The root problem here is that we can not be 100% sure if the activation is running (excepting for the heartbeat). So the procedure is: set the activation as "workers-offline" so the administrator can try to recover the node or check if the activation can be really restarted. It is the lesser evil. As we can not restart it automatically without the risk of duplicate the activation (which might have catastrophic consequences) at least we allow them to restart it manually. 

In this case, we assume the old pod is cleaned up (or the node is truly dead) and we just run the "restart" operation ignoring the error in the cleanup step.

bzwei: That's fine. My concern is if the stop failed will it block the execution to continue a start.

Alex-Izquierdo: It's not the case, we already skip the exception and continue with the flow. In fact there are more cases where we want to continue with the restart if the cleanup fails, if I remember correctly.

PR #896, opened 2024-05-09
feat: allow WEBSOCKET_TOKEN_BASE_URL setting

Alex-Izquierdo:
There may be deployments where the WEBSOCKET_TOKEN_BASE_URL doesn't coincide with WEBSOCKET_BASE_URL. 
This PR allows to set a different value for WEBSOCKET_TOKEN_BASE_URL.

### DIFF
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -319,9 +319,11 @@ REST_FRAMEWORK = {
 DEPLOYMENT_TYPE = settings.get("DEPLOYMENT_TYPE", "podman")
 WEBSOCKET_BASE_URL = settings.get("WEBSOCKET_BASE_URL", "ws://localhost:8000")
 WEBSOCKET_SSL_VERIFY = settings.get("WEBSOCKET_SSL_VERIFY", "yes")
-WEBSOCKET_TOKEN_BASE_URL = WEBSOCKET_BASE_URL.replace(
-    "ws://", "http://"
-).replace("wss://", "https://")
+WEBSOCKET_TOKEN_BASE_URL = settings.get("WEBSOCKET_TOKEN_BASE_URL", None)
+if WEBSOCKET_TOKEN_BASE_URL is None:
+    WEBSOCKET_TOKEN_BASE_URL = WEBSOCKET_BASE_URL.replace(
+        "ws://", "http://"
+    ).replace("wss://", "https://")
 PODMAN_SOCKET_URL = settings.get("PODMAN_SOCKET_URL", None)
 PODMAN_MEM_LIMIT = settings.get("PODMAN_MEM_LIMIT", "200m")
 PODMAN_ENV_VARS = settings.get("PODMAN_ENV_VARS", {})


src/aap_eda/settings/default.py
```diff
@@ -319,9 +319,11 @@ def _get_databases_settings() -> dict:
 DEPLOYMENT_TYPE = settings.get("DEPLOYMENT_TYPE", "podman")
 WEBSOCKET_BASE_URL = settings.get("WEBSOCKET_BASE_URL", "ws://localhost:8000")
 WEBSOCKET_SSL_VERIFY = settings.get("WEBSOCKET_SSL_VERIFY", "yes")
-WEBSOCKET_TOKEN_BASE_URL = WEBSOCKET_BASE_URL.replace(
-    "ws://", "http://"
-).replace("wss://", "https://")
+WEBSOCKET_TOKEN_BASE_URL = settings.get("WEBSOCKET_TOKEN_BASE_URL", None)
```
hsong-rh, line 322: do we need to check for `https` if WEBSOCKET_TOKEN_BASE_URL is set?

Alex-Izquierdo: I don't think so, we are not doing validations like that in other settings. It's a responsibility of the user/installer to provide a right value.

hsong-rh: the old logic will replace `ws`, `wss` with `http` and `https` respectively, do we need to keep them? does ansible-rulebook handle them correctly if the token starts with `ws` or `wss`?

Alex-Izquierdo: I'm not sure if I'm following you @hsong-rh. 
We want to keep the current behavior not break any existing deployment/pipeline. 

The existing logic is to keep the backward compatibility and allow WEBSOCKET_TOKEN_BASE_URL to be optional. The existing logic assumes that WEBSOCKET_TOKEN_BASE_URL is the same (excepting the scheme) as WEBSOCKET_BASE_URL which is true in the most of the cases, so we don't force to define a new WEBSOCKET_TOKEN_BASE_URL variable. Just allow to override it for the cases where WEBSOCKET_TOKEN_BASE_URL is a different hostname than WEBSOCKET_BASE_URL

hsong-rh: Let me clarify myself. If WEBSOCKET_TOKEN_BASE_URL in the setting is set to `ws://localhost:8000`, the existing logic will change it into `http://localhost:8000` and send it to ansible-rulebook. With your PR  `ws://localhost:8000` is sent directly to ansible-rulebook engine. Will ansible-rulebook accept the `ws` schema? do we need to keep existing logic?

Alex-Izquierdo: @hsong-rh It is not as you are describing. If WEBSOCKET_TOKEN_BASE_URL is set to `ws://localhost:8000` is a wrong value and ansible-rulebook would fail. It should be "http or https", because the token base url if for the rest api. 

If WEBSOCKET_TOKEN_BASE_URL is not set (the current logic) we take WEBSOCKET_BASE_URL and replace the scheme from ws to http. 

We want to keep the existing behavior for backward compatibility.

hsong-rh: I see, you mean users should take the responsibility to make sure their input token url is correct.

hsong-rh (approved):

src/aap_eda/settings/default.py
Alex-Izquierdo: well, that's the same for any other variable. If WEBSOCKET_BASE_URL has a wrong value it would fail as well. This value is managed by the installer/operator/docker-compose, rarely would be override by users.

ttuffin (approved):

PR #897, opened 2024-05-09
feat: add prefix to redis keys

bzwei:
Monkey patch originally hardcoded redis keys used by rq and rq-scheduler to have EDA{uuid} prefix to avoid conflicts in a centralized redis env.

AAP-22834: Create named Redis key prefixes for EDA

Sample keys after the patch:
* EDAefeb4900-rq:clean_registries:activation
* EDAefeb4900-rq:results:activation-1
* EDAefeb4900-rq:worker:01fccd9976584eeab865776981ea4f9c
* EDAefeb4900-rq:clean_registries:default
* EDAefeb4900-rq:job:f91b8900-ec95-40d1-8e00-507a51fa9c9
* EDAefeb4900-rq:workers:default
* EDAefeb4900-rq:scheduler:scheduled_jobs
* EDAefeb4900-rq:job:activation-1
* EDAefeb4900-rq:workers
* EDAefeb4900-rq:finished.0
* EDAefeb4900-rq:workers:activation
* EDAefeb4900-rq:queues
* EDAefeb4900-rq:scheduler_instance:385c9b98c1304cfda83cf8dec92a0fd3


src/aap_eda/core/tasking/__init__.py
```diff
@@ -40,6 +44,57 @@
     None,
 ]
 
+redis_prefix = f"EDA{service_id()[0:8]}-rq"
```
Alex-Izquierdo, line 47: The max size for a key in redis is 512M, even if the chance for a collision is small, do we need to truncate it?

src/aap_eda/core/tasking/__init__.py
```diff
@@ -40,6 +45,62 @@
     None,
 ]
 
+try:
+    redis_prefix = f"EDA{service_id()[0:8]}-rq"
+except Exception:
```
Alex-Izquierdo, line 50: We need here to be cautious. I don't think catching all the exception is a good idea. If it fails by some other unexpected reason, we would generate a new UUID every restart with catastrophic results. I think we should catch only the expected exception: ProgrammingError
Even in that case, we might want to use a fixed value, so it would not change between restarts, would be better to debug and would work correctly. 


But I even in that case, I wonder if it would be better if we can figure out a way detect of migrations are applied instead of catching the exception.

src/aap_eda/core/tasking/__init__.py
```diff
@@ -41,6 +47,76 @@
 ]
 
 
+def enable_redis_prefix():
+    if rq.worker_registration.REDIS_WORKER_KEYS != "rq:workers":
+        # changes have been made
+        return
+
+    try:
+        redis_prefix = f"EDA{service_id()[0:8]}-rq"
```
Alex-Izquierdo, line 56: separator between EDA and service_id
```suggestion
        redis_prefix = f"EDA-{service_id()[0:8]}-rq"
```

Alex-Izquierdo (commented):
LGTM, seems working as expected I can not see any other rq key. Automated and manual tests pass. 

I strongly recommend to fully freeze the version constraint for rq stuff (rq, django-rq rq-scheduler) and take care explicitly of any library upgrade while this monkeypatch is active.

bzwei:
@jamesmarshall24 @msmagnanijr If this PR is approved we need to make sure the scheduler in VM and OCP deployments now has a dependency on eda-api on start-up.

bzwei:
> LGTM, seems working as expected I can not see any other rq key. Automated and manual tests pass.
> 
> I strongly recommend to fully freeze the version constraint for rq stuff (rq, django-rq rq-scheduler) and take care explicitly of any library upgrade while this monkeypatch is active.

rq is now pinned to 1.13.0. django-rq is pinned to 2.8.0. rq-scheduler had been pinned to 0.10.

Alex-Izquierdo (approved):

benthomasson:
@bzwei Please create an equivalent PR for RQ that enables configuring Redis key prefixes.    We can use that PR to make a patch until it is accepted.

benthomasson (approved):
I am fine with this solution until RQ supports prefixes.

PR #900, opened 2024-05-13
fix: PENDING has two potential status messages

jshimkus-rh:


Alex-Izquierdo (commented):
I'm not sure if it is the right solution and we may mask a bug or maybe we are missing something else. What causes the message to not be updated? In the integration test there are no workers, so when the creation is executed the dispatcher, which is called by the view, should set the expected message. Is maybe a problem of the mac environment or maybe the test is not properly isolated?

jshimkus-rh:
> In the integration test there are no workers, so when the creation is executed

Interesting.  The docs do state:
    "To run tests locally, you need to have a running instance of postgresql and redis."
But they don't say *only* postgresql and redis.
And that's not being legalistic 😄 as the difference between just postgresql/redis and a full deployment hasn't (as far as I'm aware) made a difference to date.

PR #901, opened 2024-05-14
test: add e2e multinode for e2e-dispatch workflow

Alex-Izquierdo:
Add to the e2e-dispatch workflow the tests for multinode based on docker-compose-dev implemented in https://github.com/ansible/eda-qa/pull/370
Jira: https://issues.redhat.com/browse/AAP-23959

### DIFF
--- a/.github/workflows/e2e-dispatch.yml
+++ b/.github/workflows/e2e-dispatch.yml
@@ -96,3 +96,58 @@ jobs:
         run: |
           docker compose -p eda -f docker-compose-dev.yaml logs
 
+  api-e2e-multinode-tests:
+    if: >
+      (github.repository == 'ansible/eda-server' && github.event_name == 'workflow_dispatch') ||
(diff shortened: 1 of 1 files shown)


Alex-Izquierdo:
Unfortunately this can not be tested until merge it because on-demand workflows are always read from the main branch.

Dostonbek1:
> Unfortunately this can not be tested until merge it because on-demand workflows are always read from the main branch.

Can it not be tested in fork?

Alex-Izquierdo:
> > Unfortunately this can not be tested until merge it because on-demand workflows are always read from the main branch.
> 
> Can it not be tested in fork?

It would require to rewrite the main branch as well as the workflow, not mention the need to replicate the secrets. The amount of necessary differences makes a less reliable proof. I think in this case the effort it not worth, in eda-qa repo we do the same.
In the unlikely case of some error, once the job is in main, we can detect easier the problem and fix it here, which I think it would be easier.

Dostonbek1 (approved):

hsong-rh (approved):

PR #904, opened 2024-05-14
fix: [AAP-23488] disallow deletion of "worker offline" activation

jshimkus-rh:


src/aap_eda/api/views/activation.py
```diff
@@ -110,6 +99,39 @@ def create(self, request):
...
+                    "cannot be deleted."
+                ),
+            ),
+        },
+    )
+    def destroy(self, request, *args, **kwargs):
+        activation = self.get_object()
+
+        if activation.status == ActivationStatus.WORKERS_OFFLINE:
+            return Response(status=status.HTTP_409_CONFLICT)
```
Alex-Izquierdo, line 123: I think here we should have a explicit error message.

jshimkus-rh: Explicit how?  Include the activation id? name?
A log entry?

Alex-Izquierdo: I mean we must say the reason. Example: https://github.com/ansible/eda-server/blob/main/src/aap_eda/api/views/eda_credential.py#L249

jshimkus-rh: Yep.  Figured that out. 😄

ttuffin (approved):

Alex-Izquierdo (approved):

PR #905, opened 2024-05-15
refactor: standardize boolean as string

Alex-Izquierdo:


Alex-Izquierdo:
@ansible/eda-maintainers If you know about other potential usage of str_to_bool that I may miss, please let me know to update the PR.

src/aap_eda/utils/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+
+def str_to_bool(value: str) -> bool:
+    return value.lower() in ("yes", "true", "1")
```
jshimkus-rh, line 17: Is it consistent with other ansible code to allow anything but these three values to be `false`?

Alex-Izquierdo: It is consistent with what we are currently doing here. There is not a specific requirement about it.

Dostonbek1 (approved):

jshimkus-rh (approved):

PR #907, opened 2024-05-16
enable ssl postgres

msmagnanijr:


Alex-Izquierdo (commented):
We have a docker-compose file for all configuration related with certs. 
https://github.com/ansible/eda-server/blob/main/tools/docker/docker-compose-dev-redis-tls.yaml

src/aap_eda/settings/default.py
```diff
@@ -241,7 +241,7 @@ def _get_databases_settings() -> dict:
             "PASSWORD": settings.get("DB_PASSWORD"),
             "NAME": settings.get("DB_NAME", "eda"),
             "OPTIONS": {
-                "sslmode": settings.get("PGSSLMODE", default="prefer"),
+                "sslmode": settings.get("PGSSLMODE", default="allow"),
```
Alex-Izquierdo, line 244: Why this change?

tools/docker/docker-compose-dev.yaml
```diff
@@ -217,6 +221,23 @@ services:
...
 
+  certs:
+    image: quay.io/fedora/fedora-minimal
+    volumes:
+      - ./postgres_ssl_config/certs:/etc/ssl/certs 
+    entrypoint: 
+      - sh
+      - -c
+      - |
+        sudo dnf -y install openssl
```
Alex-Izquierdo, line 232: I think this should be encapsulated in a script and used in https://github.com/ansible/eda-server/blob/main/tools/docker/docker-compose-dev-redis-tls.yaml

tools/docker/docker-compose-dev.yaml
```diff
@@ -217,6 +221,23 @@ services:
       - '../../:/app/src:z'
     restart: always
 
+  certs:
+    image: quay.io/fedora/fedora-minimal
+    volumes:
+      - ./postgres_ssl_config/certs:/etc/ssl/certs
```
Alex-Izquierdo, line 227: missing trailing `:z` flag

Alex-Izquierdo (changes requested):

tools/docker/docker-compose-dev.yaml
msmagnanijr: great, I will move it to this.

tools/docker/docker-compose-dev.yaml
msmagnanijr: good catch!

src/aap_eda/settings/default.py
msmagnanijr: https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-PROTECTION

msmagnanijr:
How to Test it?

Open your terminal and start the EDA environment with the specific Docker Compose file:

`docker-compose -f docker-compose-dev-redis-tls.yaml up -d`

After ensuring that the EDA is up, open a shell session in the PostgreSQL container:

`docker exec -u 0 -it eda-postgres-1 sh
`
Inside the container, connect to the PostgreSQL database using psql:

`psql -U postgres -d eda`


Once connected to the database, execute the following SQL query to verify the SSL status:

```
show SSL;

ssl 
-----
 on
(1 row)


SELECT * FROM pg_stat_ssl;
```

The output of the query will provide details on SSL connections. Look for the ssl column to see if it is t (true),  there is a connection using SSL along with other SSL-related information such as ssl_cipher, ssl_client_dn, etc.

In this case, it will show f (false) since we do not yet have SSL e2e working in EDA.

fincamd (approved):
This works for me running the test suite locally after setting up the docker-compose-dev-redis.yaml.
Just a small fix, I had to append `-p eda` to the docker-compose command so that the container name matches the instructions.

`docker-compose -p eda -f tools/docker/docker-compose-dev-redis-tls.yaml up -d`
`docker exec -u 0 -it eda-postgres-1 sh`
`psql -U postgres -d eda`

PR #911, opened 2024-05-22
Add pyinstrument to settings

benthomasson:


src/aap_eda/settings/default.py
```diff
@@ -193,6 +193,7 @@ def _get_secret_key() -> str:
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "crum.CurrentRequestUserMiddleware",
+    "pyinstrument.middleware.ProfilerMiddleware",
```
Alex-Izquierdo, line 196: Do we want to add this package as application dependency? I would say we want it only for development in
https://github.com/ansible/eda-server/blob/d683bcbfb50be39d51da05aa5df52eabd8813a0d/pyproject.toml#L83 and add the middleware in settings/development.py

ttuffin:
Closing due to inactivity.

PR #915, opened 2024-05-23
fix: change request body of update to role_definition_create

hsong-rh:
https://issues.redhat.com/browse/AAP-22846

This PR changes the request body of PUT action for `/role_definitions/:id/` from RoleDefinitionDetail to RoleDefintionCreate. 

image

### DIFF
--- a/src/aap_eda/api/views/dab_decorate.py
+++ b/src/aap_eda/api/views/dab_decorate.py
@@ -41,6 +41,16 @@ for viewset_cls in [
                 ),
             },
         ),
+        update=extend_schema(
+            description=f"Update a {cls_name}.",
+            request=convert_to_create_serializer(viewset_cls.serializer_class),
+            responses={
+                status.HTTP_200_OK: OpenApiResponse(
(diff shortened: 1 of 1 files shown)


hsong-rh:
@AlanCoding Please review

AlanCoding:
My only concern looking at this is that the "assignment" models should be immutable and, thus, should have a `create` method, and PUT should be invalid. The loop this modifies does include those models. So I checked out your branch to confirm:

Screenshot from 2024-05-28 12-01-44

That is the same before/after this change, and `RoleUserAssignmentCreate` isn't referenced in any new places either in the `openapi.json` data. So it seems that adding it to this decorator still won't affect the schema for a viewset, if that method isn't defined for that viewset.

So this looks fine to me, but I wanted to leave this comment to be very clear about the intention behind the API, in case it breaks at a later point and people get confused.

AlanCoding (approved):

Alex-Izquierdo (approved):

mkanoor (approved):

PR #917, opened 2024-05-28
feat: forbid write apis when resources are managed by gateway

bzwei:
forbid create/update/delete organzation, team, user apis when DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED is True

AAP-24301: Update 2.5 EA EDA API to error on endpoints that correspond to gateway-managed resources


Alex-Izquierdo (approved):

src/aap_eda/api/views/team.py
```diff
@@ -50,6 +53,10 @@
                 TeamSerializer,
                 description="Return the new team.",
             ),
+            status.HTTP_403_FORBIDDEN: OpenApiResponse(
+                None,
+                description="Create should be done through the platform ingress.",  # noqa: E501
```
mkanoor, line 58: @bzwei Wont this message be misleading when we return a 403 because of RBAC or other reasons. Should this be a more specific exception that says ExternallyManagedResourceException.

mkanoor: @bzwei extend_schema has an option called [exclude](https://drf-spectacular.readthedocs.io/en/latest/drf_spectacular.html#drf_spectacular.utils.extend_schema) Can we use that to remove the end points from the swagger UI and OpenAPI spec. 
we could set
``` 
exclude=settings.DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED
```

src/aap_eda/api/views/team.py
```diff
@@ -91,11 +102,34 @@ def filter_queryset(self, queryset):
...
+                None, description="Delete successful."
+            ),
+            status.HTTP_403_FORBIDDEN: OpenApiResponse(
+                None,
+                description="Delete should be done through the platform ingress.",  # noqa: E501
+            ),
+        },
+    )
+    def destroy(self, request, *args, **kwargs):
+        if settings.DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED:
```
mkanoor, line 127: @bzwei could this check be in a mixin that every class that wants to enable this check use in its inheritance chain. so it becomes a one line call in all the functions.

src/aap_eda/api/views/user.py
```diff
@@ -249,5 +250,11 @@ def destroy(self, request, *args, **kwargs):
         if instance == request.user:
             return Response(status=status.HTTP_403_FORBIDDEN)
 
+        if settings.DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED:
```
mkanoor, line 253: @bzwei can we use a better name for this setting maybe

CLONED_REMOTE_RESOURCE_ENABLED

or
USE_REMOTE_RESOURCES

Maybe we have a list of objects that are remote resources so if we need to add something down the road like Groups etc we can just add it to that list.

dhaustein: +1 for `USE_REMOTE_RESOURCES`

mkanoor:
@bzwei How do these remote resources get cleaned up once they are in the EDA Database? If there is no delete would there be some system task that will do periodic checks to cleanup resources by synchronizing with the Gateway.
If the Gateway has the UI for doing a create/update/delete which effects the Gateway DB why can't we teach our UI to be getting the list of these remote objects directly listed via the gateway end point. Our list is only going to be partial based on the logged in user. Why not have all the shared resources be stored in Gateway and where we need it we store a ref to the remote object e.g. If organization is remote we should have a remote_resource attribute that stores the value as 

{{ GATEWAY_HOST }}/api/v1/organization/15

So in or database we can have references to external objects and not copies. As soon as we start storing copies we have to worry about it going stale. With references we only have a link in our database.

I am guessing eventually the Credentials and Credential Types will also end up as a Remote Resource and when the UI wants to attach them to the Activation it would get the current list from the Gateway and then use a link in our database. When the workers start activation they will make a REST API call to fetch the Credential and Credential Types and do the appropriate things with it and send the data to ansible-rulebook.

src/aap_eda/api/views/user.py
bzwei: This term is specified in the jira issue, used by all products across AAP.

bzwei:
> @bzwei How do these remote resources get cleaned up once they are in the EDA Database?
I tested and noticed the local database is in sync with the gateway even when the deletion occurs. There is no extra code to insert/delete the local db. The underly dab module must have the logic to keep them in sync.

src/aap_eda/api/views/team.py
bzwei: 403 is per the Jira issue spec

hsong-rh (approved):

src/aap_eda/api/serializers/mixins.py
```diff
@@ -0,0 +1,27 @@
...
+
+from aap_eda.api import exceptions as api_exc
+
+
+class SharedResourceSerializerMixin:
+    def validate_shared_resource(self):
+        if settings.DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED:
+            view = self.context.get("view")
+            action = view.action.capitalize() if view else "Action"
+            raise api_exc.Forbidden(
```
Alex-Izquierdo, line 25: Is 403 the right response? I remember some previous discussion about it and I think the conclusion was that 403 is only for access rights, and things like this should be 409
```This response is sent when a request conflicts with the current state of the server.```

Alex-Izquierdo (approved):

src/aap_eda/api/serializers/mixins.py
bzwei: The parent issue https://issues.redhat.com/browse/AAP-24178 requests the return code to be 403.

AlanCoding:
Link companion patch https://github.com/ansible/awx/pull/15234

PR #921, opened 2024-05-31
fix: Set GNUPGHOME to a temp dir for project sync

bzwei:
AAP-24871: Project sync SCM clone failed Failed to import the gpg public key


Alex-Izquierdo (dismissed):

src/aap_eda/services/project/scm.py
```diff
@@ -211,6 +214,7 @@ def clone(
                 key_file.close()
             if gpg_key_file:
                 gpg_key_file.close()
+                gpg_home_dir.cleanup()
```
mkanoor, line 217: @bzwei how about 
```
    if gpg_home_dir:
       gpg_home_dir.cleanup()
```

src/aap_eda/services/project/scm.py
```diff
@@ -174,6 +175,8 @@ def clone(
             gpg_key_file.write("\n")
             gpg_key_file.flush()
             extra_vars["verify_commit"] = "true"
+            gpg_home_dir = tempfile.TemporaryDirectory()
+            env_vars["GNUPHOME"] = gpg_home_dir.name
```
hsong-rh, line 179: it's GNUPGHOME, right?

src/aap_eda/services/project/scm.py
bzwei: gpg_key_file and gpg_home_dir are created in the same block. No need to add an if condition

mkanoor:
Have we tested this given the first commit wasn't setting the correct parameter

bzwei:
> Have we tested this given the first commit wasn't setting the correct parameter
Yes it was tested locally. The first commit had the wrong env, the key was neither imported to the temp dir, which produced a false positive test. Now both had been fixed and tested.

hsong-rh (approved):

Alex-Izquierdo (approved):

PR #925, opened 2024-06-03
Add wait-for-migrations script to be used in eda-server container on k8s

rooftopcellist:
Add wait-for-migrations script to be used in eda-server container on k8s.

This will allow us to wait until migrations are complete before fully starting the scheduler and worker pods.  It will be run in an initContainer for those deployments.

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -11,6 +11,9 @@ RUN mkdir -p /var/lib/eda/redis-tls \
     && chmod 0775 /var/lib/eda/redis-tls
 COPY ./tools/docker/redis-tls /var/lib/eda/redis-tls
 
+
+COPY ./tools/docker/assets/wait-for-migrations /var/lib/eda/wait-for-migrations
+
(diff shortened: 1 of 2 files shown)


Alex-Izquierdo:
As we discussed privately, I think this approach can be difficult to manage as it requires changes in multiple places, specially if this is a temporal workaround.  I propose a configmap instead, so everything is contained in the operator and it can be removed easily in later versions when the script would be no longer needed.

rooftopcellist:
@Alex-Izquierdo The configmap approach won't work because of permissions issues on Openshift:

```
$ oc logs -f -c wait-for-migrations eda-scheduler-9cdbdf445-cb4mc 
chmod: changing permissions of '/usr/bin/wait-for-migrations': Read-only file system
```

Alex-Izquierdo:
Look at https://github.com/ansible/eda-server/pull/927 @rooftopcellist

Alex-Izquierdo:
closed in favor of https://github.com/ansible/eda-server/pull/927

PR #926, opened 2024-06-03
fix: raise an error if service_id is not ready

bzwei:
Scheduler and workers automatically wait on the service_id being ready so that we don't rely on  the installer or operator to maintain the start-up dependency.
Do not wait if the startup is for migration

Update:
Based on the discussion, this PR now only raise an error if the process is running as a server, a scheduler, or a worker. It does not wait on possible migration done by another process.

### DIFF
--- a/src/aap_eda/core/tasking/__init__.py
+++ b/src/aap_eda/core/tasking/__init__.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import logging
+import sys
 import uuid
 from datetime import datetime, timedelta
 from types import MethodType
@@ -47,17 +48,33 @@ _ErrorHandlersArgType = Union[
 ]
 
 
+class NoServiceIdError(Exception):
+    ...
+
+
+def get_redis_prefix():
+    try:
+        return f"EDA-{service_id()[0:8]}-rq"
+    except ProgrammingError:
+        # service_id() will fail at the first migration when
+        # dab_resource_registry_serviceid has not been created
+        if any(
+            arg in sys.argv for arg in ["runserver", "scheduler", "rqworker"]
+        ):
+            raise NoServiceIdError(
+                "service_id is required to start the service"
+            )
+        logger.info("Failed ot get service_id. Continue")
+        return ""
+
+
 def enable_redis_prefix():
     if rq.worker_registration.REDIS_WORKER_KEYS != "rq:workers":
         # changes have been made
         return
 
-    try:
-        redis_prefix = f"EDA-{service_id()[0:8]}-rq"
-    except ProgrammingError:
-        # service_id() will fail at the first migration when
-        # dab_resource_registry_serviceid has not been created
-        logger.info("Failed to get service id")
+    redis_prefix = get_redis_prefix()
+    if not redis_prefix:
         return
 
     rq.worker_registration.REDIS_WORKER_KEYS = f"{redis_prefix}:workers"


src/aap_eda/core/tasking/__init__.py
```diff
@@ -47,17 +49,35 @@
...
+    for _ in range(50):
+        try:
+            return f"EDA-{service_id()[0:8]}-rq"
+        except ProgrammingError:
+            # service_id() will fail at the first migration when
+            # dab_resource_registry_serviceid has not been created
+            if (
+                len(sys.argv) == 2
+                and sys.argv[0].endswith("aap-eda-manage")
+                and sys.argv[1] in ["migrate", "makeimgrations"]
```
Alex-Izquierdo, line 62: `aap-eda-manage` could not be called directly, so `sys.argv` could be more than 2 and `aap-eda-manage` could not be sys.argv[0]`
What about `sys.argv[-1] in ["migrate", "makemigrations"]`

Alex-Izquierdo, line 62: typo: makeimgrations

Alex-Izquierdo, line 62: I think we need more commands to be handled, we don't want to wait for the service id and fail when running "showmigrations"

src/aap_eda/core/tasking/__init__.py
```diff
@@ -47,17 +49,35 @@
...
+            ):
+                logger.info(
+                    "Failed to get service id. It will be generated shortly"
+                )
+                return ""
+            else:
+                logger.info("Waiting for AAP service_id")
+                time.sleep(6)
```
Alex-Izquierdo, line 70: this could just one, so we can configure the argument in the range function as num of seconds. That value could be a constant.

src/aap_eda/core/tasking/__init__.py
```diff
@@ -47,17 +49,35 @@
...
+    raise TimeoutError("Timeout waiting for AAP service_id")
```
Alex-Izquierdo, line 71: We should not mention "AAP" in upstream code (more than we already do, I mean xD)

src/aap_eda/core/tasking/__init__.py
Alex-Izquierdo: Second thought, I think the most reliable way could be
```
any(arg in sys.argv for arg in ["migrate", "makemigrations"])
```

Alex-Izquierdo:
I think we should not take care of any logic to wait_for and just check if we are running migrations and otherwise fail. As it is a external command, such a logic should be handled externally. 

Also, as alternative to catch the exception, we could just look if the migrations are already applied
```
from django.db.migrations.executor import MigrationExecutor
from django.db import connection

executor = MigrationExecutor(connection)
targets = executor.loader.graph.leaf_nodes()
plan = executor.migration_plan(targets)
# there are pending migrations...
if plan: 
```

Alex-Izquierdo:
Look at https://github.com/ansible/eda-server/pull/927

bzwei:
@Alex-Izquierdo This is a good idea. It requires we update the startup command for all related pods (worker, scheduler), right?

Alex-Izquierdo:
> @Alex-Izquierdo This is a good idea. It requires we update the startup command for all related pods (worker, scheduler), right?

Yes, in the docker-compose there is no need, because the compose file itself handles the dependency on the api pod. I'm talking with @rooftopcellist so he can use that command in the operator.

Alex-Izquierdo:
In this PR, I would say we don't need to apply any "wait for" logic, just detect if we run migrations commands.

bzwei:
@Alex-Izquierdo I updated the PR. This PR does not detect whether a migration has completed because it does not need to wait for all migrations be done every time. No error is raised as long as the service_id can be fetched, which is populated the first time.

src/aap_eda/core/tasking/__init__.py
```diff
@@ -2,6 +2,8 @@
 from __future__ import annotations
 
 import logging
+import sys
+import time
```
Alex-Izquierdo, line 6: time is not used

hsong-rh (approved):

PR #930, opened 2024-06-05
feat(dispatcher): return random queue with multiple candidates

Alex-Izquierdo:
The current implementation always gives priority to the first queue configured. 
If there are multiple candidates, returns one randomly.

### DIFF
--- a/src/aap_eda/tasks/orchestrator.py
+++ b/src/aap_eda/tasks/orchestrator.py
@@ -13,6 +13,7 @@
 #  limitations under the License.
 
 import logging
+import random
 from collections import Counter
 from datetime import datetime, timedelta
 from typing import Optional, Union
@@ -333,7 +334,13 @@ def get_least_busy_queue_name() -> str:
             "No healthy queue found to dispatch the request",
         )
 
-    return queue_counter.most_common()[-1][0]
+    min_count = queue_counter.most_common()[-1][1]
+    least_common = [
+        queue for queue, count in queue_counter.items() if count == min_count
+    ]
+    if len(least_common) == 1:
+        return least_common[0]
+    return random.choice(least_common)
 
 
 def get_queue_name_by_parent_id(
--- a/tests/unit/test_orchestrator.py
+++ b/tests/unit/test_orchestrator.py
@@ -274,10 +274,8 @@ def test_get_least_busy_queue_name_multiple_queues(
         "queue1",
         "queue2",
     ]
-    unexpected_queue = "queue3"
     queue = get_least_busy_queue_name()
     assert queue in expected_queues
-    assert queue != unexpected_queue
 
 
 @pytest.fixture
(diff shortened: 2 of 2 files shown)


hsong-rh (dismissed):

tests/unit/test_orchestrator.py
```diff
@@ -240,6 +267,19 @@ def test_get_least_busy_queue_name(fixture, request):
...
+    three_queues_two_candidates,
+):
+    expected_queues = [
+        "queue1",
+        "queue2",
+    ]
+    unexpected_queue = "queue3"
+    queue = get_least_busy_queue_name()
+    assert queue in expected_queues
+    assert queue != unexpected_queue
```
jshimkus-rh, line 280: The second assert is redundant.

src/aap_eda/tasks/orchestrator.py
```diff
@@ -333,7 +334,13 @@ def get_least_busy_queue_name() -> str:
...
         )
 
-    return queue_counter.most_common()[-1][0]
+    min_count = queue_counter.most_common()[-1][1]
+    least_common = [
+        queue for queue, count in queue_counter.items() if count == min_count
+    ]
+    if len(least_common) == 1:
+        return least_common[0]
+    return random.choice(least_common)
```
jshimkus-rh, line 343: Is checking the length really going to be significantly faster than just calling random.choice regardless?

Alex-Izquierdo: Sure, maybe the difference is minor but for sure it will be faster.
In OCP and single node scenarios there will be always one queue, so it is worth IMO.

tests/unit/test_orchestrator.py
Alex-Izquierdo: agree. Removed

src/aap_eda/tasks/orchestrator.py
Alex-Izquierdo: ¯\_(ツ)_/¯

```
In [2]: l = ["value"]

In [3]: import random

In [4]: def no_len(l):
   ...:     return random.choice(l)
   ...: 

In [5]: def with_len(l):
   ...:     if len(l) == 1:
   ...:         return l[0]
   ...:     return random.choice(l)
   ...: 

In [6]: %timeit no_len(l)
230 ns ± 7.37 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [7]: %timeit with_len(l)
45.5 ns ± 1.45 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

jshimkus-rh: Ooooh, 6 ns!!!  Thank the deity all we ever do is billions of these specific operations. 😄

jshimkus-rh (approved):

hsong-rh (approved):

PR #931, opened 2024-06-06
fix: set job id for scheduled jobs

Alex-Izquierdo:
We have noticed that the lock mechanism in the scheduler is not 100% reliable and when multiple schedulers are executed at the exact same time they can duplicate the jobs. Set the job_id to a static value so we can make sure we schedule always only one job. 

Jira: https://issues.redhat.com/browse/AAP-25019

### DIFF
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -442,14 +442,22 @@ RQ_QUEUES = get_rq_queues()
 RULEBOOK_QUEUE_NAME = settings.get("RULEBOOK_QUEUE_NAME", "activation")
 
 RQ_STARTUP_JOBS = []
+
+# Id of the scheduler job it's required when we have multiple instances of
+# the scheduler running to avoid duplicate jobs
 RQ_PERIODIC_JOBS = [
     {
         "func": (
             "aap_eda.tasks.orchestrator.enqueue_monitor_rulebook_processes"
         ),
         "interval": 5,
+        "id": "enqueue_monitor_rulebook_processes",
     },
-    {"func": "aap_eda.tasks.project.monitor_project_tasks", "interval": 30},
+    {
+        "func": "aap_eda.tasks.project.monitor_project_tasks",
+        "interval": 30,
+        "id": "monitor_project_tasks",
+    },
 ]
 RQ_CRON_JOBS = []
 RQ_SCHEDULER_JOB_INTERVAL = settings.get("SCHEDULER_JOB_INTERVAL", 5)


bzwei (approved):

src/aap_eda/settings/default.py
```diff
@@ -442,14 +442,22 @@ def get_rq_queues() -> dict:
...
         "func": (
             "aap_eda.tasks.orchestrator.enqueue_monitor_rulebook_processes"
         ),
         "interval": 5,
+        "id": "enqueue_monitor_rulebook_processes",
+    },
+    {
+        "func": "aap_eda.tasks.project.monitor_project_tasks",
+        "interval": 30,
+        "id": "monitor_project_tasks",
```
hsong-rh, line 459: Is this newly added `id` used as the parameter for function `monitor_project_tasks`? How does it work?

Alex-Izquierdo: @hsong-rh 
This id is the job_id for rq_redis. Works in the same way as we do in unique_enqueue:
https://github.com/ansible/eda-server/blob/main/src/aap_eda/core/tasking/__init__.py#L312

You can see the function signature here: 
https://github.com/rq/rq-scheduler/blob/master/rq_scheduler/scheduler.py#L241

hsong-rh (approved):

jshimkus-rh:
I'm concerned that the description, while not necessarily incorrect, glosses over a synchronization issue between parallel executions of, at least, enqueue_monitor_rulebook_processes.  Also it seems to me (unless I've misunderstood) that these changes relative to the periodic tasks rely on the same mechanism the description calls out as problematic.

Alex-Izquierdo:
@jshimkus-rh I think you might be mixing two different mechanisms, the scheduler lock system when it schedules the job, and the mechanism of the rqworker to fetch/update the job. We have not observed issues at the second one so far. 
In this case we were doing it wrong regardless of the lock system of the scheduler when we execute https://github.com/ansible/eda-server/blob/main/src/aap_eda/core/management/commands/scheduler.py#L111
We should have had the job_id defined from the beginning, just we didn't hit the issue until start to paly with multiple schedulers. 

I agree that the description might be inaccurate, it suggests that is a problem of the scheduler, but the scheduler ensures that you can not have two instances setting the same job, here the problem is that prior to run the scheduler we initially created twice the job. 

 In this screenshot you can see how looks redis between [this line](https://github.com/ansible/eda-server/blob/main/src/aap_eda/core/management/commands/scheduler.py#L150) and [this line](https://github.com/ansible/eda-server/blob/main/src/aap_eda/core/management/commands/scheduler.py#L152)
 
Screenshot from 2024-06-06 13-47-09

jshimkus-rh:
> @jshimkus-rh I think you might be mixing two different mechanisms, the scheduler lock system when it schedules the job, and the mechanism of the rqworker to fetch/update the job. We have not observed issues at the second one so far. 

I don't think so.  And I do agree with the following...

> I believe we should have had the job_id defined from the beginning, just we didn't hit the issue until start to play with multiple schedulers.

But that's also the source of my angst.  I don't like the idea of depending on an external entity to prevent us from doing something that we shouldn't be doing; we should do that.
But I don't see the need to start a holy war about it. I'll just have to live with it.

PR #932, opened 2024-06-07
feat: run scheduled tasks to sync shared resources from gateway

bzwei:
Resync the resources every 15 ~~seconds~~ minutes. Enabled only when DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED is turned no. Need to configure the RESOURCE_SERVER's URL, SECRET_KEY, and RESOURCE_JWT_USER_ID before use.

AAP-23724: EDA: Schedule Sync Task

STEPS to prepare the dev environment:
1. start the gateway
2. find the jwt_user_id through API `https://<gateway>/api/gateway/v1/users` (for the first time)
    find the ansible_id of user _system
3. create a secret_key on the gateway shell (inside the gateway container) (for the first time)
    `django-admin generate_service_secret eda`
4. export the following ENVS and start eda-server:
    `EDA_RESOURCE_SERVER__URL=https://<gateway ip, not localhost>`
    `EDA_RESOURCE_JWT_USER_ID=<ansible_id from step 2>`
    `EDA_RESOURCE_SERVER__SECRET_KEY=<output from step 3>`
    There will be errors in the default worker logs. Proceed to step 5.
5. run service migrate data on the gateway shell (for the first time)
    `django-admin migrate_service_data --api-slug eda --username admin`
    Worker errors should go away.


src/aap_eda/settings/default.py
```diff
@@ -617,6 +617,13 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
 DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED = settings.get(
     "DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED", False
 )
+if DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED:
+    RQ_PERIODIC_JOBS.append(
+        {
+            "func": "aap_eda.tasks.shared_resources.resync_shared_resources",
+            "interval": 900,
```
mkanoor, line 624: @bzwei Should this be slightly lower than 15 minutes so we don't overlap when the new one is about to start. Also how do we specify the job timeout?

rochacbruno: I am working on adding a lock to SyncExecutor that will not allow 2 instances to execute at the same time.

src/aap_eda/settings/default.py
```diff
@@ -636,6 +643,17 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
 # Organization and object roles will come from create_initial_data
 ANSIBLE_BASE_ROLE_PRECREATE = {}
 
+# --------------------------------------------------------
+# GATEWAY RESOURCE API CLIENT
+# --------------------------------------------------------
+RESOURCE_SERVER = {
```
mkanoor, line 649: @bzwei Since all our keys have an EDA prefix would the RESOURCE_SERVER__URL have to be called EDA_RESOURCE_SERVER__URL when passing in via env var. Should DAB parameters have their own prefix. Can dynaconf support multiple prefix EDA,DAB
https://github.com/ansible/eda-server/blob/01746bec37107c64bd0e11a58adb6600015e6951/src/aap_eda/settings/default.py#L65

mkanoor: @bzwei @rochacbruno @Alex-Izquierdo Would it be better to call this **EDA_DAB_RESOURCE_SERVER__URL** to keep the DAB parameters separate.  Or **DAB_RESOURCE_SERVER__URL**. Since these are things specific to DAB. Or would the SyncExecutor() have issues locating these [parameters](https://github.com/ansible/django-ansible-base/blob/9f0f05168dbcb18c1f542aec480b80eedb5ffc3e/ansible_base/resource_registry/tasks/sync.py#L68) since it won't have the correct prefix.

src/aap_eda/settings/default.py
```diff
@@ -617,6 +617,13 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
 DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED = settings.get(
     "DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED", False
 )
+if DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED:
```
mkanoor, line 620: @bzwei Can we check if the required parameters are set after DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED is set. Since those other parameters like

1. RESOURCE_SERVER
2. RESOURCE_JWT_USER_ID
Should be set otherwise we have a bad configuration and we should stop the startup process.

src/aap_eda/settings/default.py
```diff
@@ -617,6 +617,13 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
...
+    RQ_PERIODIC_JOBS.append(
```
Alex-Izquierdo, line 621: @bzwei It needs to define the job_id to be a static value, otherwise we will have issues with multiple running schedulers.

src/aap_eda/settings/default.py
Alex-Izquierdo: We already take DAB settings [here](https://github.com/ansible/eda-server/blob/main/src/aap_eda/settings/default.py#L603) So if we are overwriting the RESOURCE_SERVER from DAB, the double score makes no sense. The double score is for dynaconf to handle nested values. 
So in this case I see two options. 
1. We expose separated values, and RESOURCE_SERVER["url"] can be as it is now (but without double score)
```
RESOURCE_SERVER = {
    "URL": settings.get("RESOURCE_SERVER_URL", "https://localhost"),
    ....
}
```

We expose RESOURCE_SERVER as object, in that way the envvar `EDA_RESOURCE_SERVER__URL` would override only `RESOURCE_SERVER["url]`. 

```
default_resource_server = {
    "URL": "https://localhost",
    "SECRET_KEY": ""),
    "VALIDATE_HTTPS": False),
}
RESOURCE_SERVER = settings.get("RESOURCE_SERVER", default_resource_server)
```

Since it is not a complex object and we need to define all these values in the most of cases, I vote to keep them separately as it is now. 

All the settings exposed has automatically added the EDA prefix, adding the "DAB" prefix to allow us to know that it is a setting that is passed to DAB, seems a bit redundant to me, it is information for the development but has no meaning in the context of the application. For EDA, it's not the "DAB_RESOURCE_SERVER" it's just the resource server. 

To me it's fine as it is called now, in the settings file we have comments indicating that those values are for DAB.

src/aap_eda/settings/default.py
```diff
@@ -636,6 +643,17 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
 # Organization and object roles will come from create_initial_data
 ANSIBLE_BASE_ROLE_PRECREATE = {}
 
+# --------------------------------------------------------
+# GATEWAY RESOURCE API CLIENT
+# --------------------------------------------------------
+RESOURCE_SERVER = {
+    "URL": settings.get("RESOURCE_SERVER__URL", "https://localhost"),
+    "SECRET_KEY": settings.get("RESOURCE_SERVER__SECRET_KEY", ""),
```
Alex-Izquierdo, line 651: Should this be our already defined SECRET KEY?

bzwei: no, it must be generated by the gateway

src/aap_eda/settings/default.py
Alex-Izquierdo: After more discussions and get more context, it seems the double underscore would be to be aligned with other components, so I guess in that case, the current approach would be fine as well.

bzwei: @Alex-Izquierdo I don't see the imported DAB settings include RESOURCE_SERVER. Can you point me the origin? If the RESOURCE_SERVER is already imported we should not overwrite.
> We already take DAB settings [here](https://github.com/ansible/eda-server/blob/main/src/aap_eda/settings/default.py#L603) So if we are overwriting the RESOURCE_SERVER from DAB, the double score makes no sense.

Alex-Izquierdo (commented):
@bzwei could you extend the docstring here to document these new variables? They seem important enough to clarify them. 
https://github.com/ansible/eda-server/blob/main/src/aap_eda/settings/default.py#L14

src/aap_eda/settings/default.py
```diff
@@ -636,6 +636,39 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
...
+    ):
+        RQ_PERIODIC_JOBS.append(
+            {
+                "func": "aap_eda.tasks.shared_resources.resync_shared_resources",  # noqa E501
+                "interval": 900,
+                "id": "resync_shared_resources",
+            }
+        )
+    else:
+        raise ImproperlyConfigured(
```
bzwei, line 666: Should we raise this error in another PR? What happens if the installer or operator has not set these ENVS?

mkanoor: It's better to catch it as early as possible otherwise imagine the amount of time spent trying to debug this. They will have the whole env setup and then the syncs won't happen. I don't think this is an optional feature if that field has been setup the others have to be set too.

Alex-Izquierdo: I agree with @mkanoor 
What we can do is merge this PR in sync with our mates @jamesmarshall24 @dsavineau and @rooftopcellist to minimize the window of errors.

bzwei:
Added
> @bzwei could you extend the docstring here to document these new variables? They seem important enough to clarify them. https://github.com/ansible/eda-server/blob/main/src/aap_eda/settings/default.py#L14

src/aap_eda/settings/default.py
```diff
@@ -83,6 +83,16 @@
                              "target": "/var/run/containers/storage",
                              "type": "bind"}]'
 
+
+Django Ansible Base settings:
+For Resource Server the following are required when
+DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED is turned on:
+* RESOURCE_SERVER__URL - The URL to connect to the resource server
+* RESOURCE_SERVER__SECRET_KEY - The secrect key to connect to the server
```
Alex-Izquierdo, line 91: Typo:
```suggestion
* RESOURCE_SERVER__SECRET_KEY - The secret key to connect to the server
```

I think the description may be incorrect. I guess we don't need the secret key to perform the connection but to be able to encrypt/decrypt data crossed with the resource server. Am I wrong?

Alex-Izquierdo (approved):

mkanoor (approved):

bzwei:
@jamesmarshall24 @rooftopcellist This PR expects EDA_RESOURCE_SERVER__URL, EDA_RESOURCE_SERVER__SECRET_KEY , EDA_RESOURCE_SERVER__VALIDATE_HTTPS, EDA_RESOURCE_JWT_USER_ID, and optionally EDA_RESOURCE_SERVICE_PATH

PR #933, opened 2024-06-10
fix: [AAP-24814] prevent request enqueuing when parent deleted

jshimkus-rh:


jshimkus-rh:
Tests to come.

Alex-Izquierdo:
Hi @jshimkus-rh What I noticed is that once the activation is deleted I saw in every loop of the monitor the message:
```
May 30 14:10:44 ip-10-0-2-14.ec2.internal automation-eda-controller-worker-1[4947]: 2024-05-30 14:10:44,224 aap_eda.tasks.orchestrator INFO     Dispatching request auto_start for activation 77
May 30 14:10:44 ip-10-0-2-14.ec2.internal automation-eda-controller-worker-1[4947]: 2024-05-30 14:10:44,225 aap_eda.tasks.orchestrator WARNING  activation 77 no longer exists, request auto_start can not be dispatched.
```

It seems that the problem is after run `_queue_auto_start`

I think there is a bug in [this PR](https://github.com/ansible/eda-server/pull/656/files#diff-71687df2e87df81216aba65e07df0f26fd2fca9fc6fa2eb1aaa275db0319b5cbR47), because in `_queue_auto_start` we couldn't add the request if the activation was deleted, but later we removed that constraint (the activation is a foreign key, but the process_parent_id don't), so we put in the queue the request even if the activation doesn't exist. 

I'm afraid that just cancelling the job could not be enough. What do you think about handling the issue in `_queue_auto_start` and check for the existence of the process_parent before add it to the queue?

jshimkus-rh:
> 
> It seems that the problem is after run `_queue_auto_start`

Right.  And cancelling that before it gets a chance to run prevents the problem.
Certainly there's no need to let a job run just to have it fail.

Ultimately, as pointed out in that PR, it's the requests_queue push that's the culprit.

Alex-Izquierdo (approved):

src/aap_eda/tasks/activation_request_queue.py
```diff
@@ -12,19 +12,37 @@
...
+            f"Unknown parent type {parent_type}",
+        )
+
     ActivationRequestQueue.objects.create(
         process_parent_type=parent_type,
         process_parent_id=parent_id,
         request=request,
     )
 
+    # Check that the parent referenced still exists.
```
Alex-Izquierdo, line 42: Shouldn't we do this before creating the registry?

jshimkus-rh: Doesn't matter much as we have two potential asynchronous operations both manipulating the db; this enqueue (on the default worker) and the actual delete (on an activation worker).  Doing it as late as possible narrows the chances of encountering issues from that.

If we really want the whole thing to be bullet-proof w/o a complicated synchronization approach we should modify the handling of auto-start handling to consider the determination the activation no longer exists to be dispositive and drop the auto-start from the requests queue.

Alex-Izquierdo: This is not exclusively related with the autostart. This is to recover a functionality that we already had and we lost [with this change](https://github.com/ansible/eda-server/pull/656/files#diff-71687df2e87df81216aba65e07df0f26fd2fca9fc6fa2eb1aaa275db0319b5cbR47). 
Although cancelling the autostart job might be good, the real issue is caused because we are creating the record when we should not do it because the process_parent does not exist. Prior to this change, this was guaranteed due to the foreign key, which is why we caught the exception in _queue_auto_start. The activation no longer exists, so there is no need to autostart anything.

With this change, we miss the constraint and we have to take care of it explicitly (that's because it was decided to implement the relationship "manually" instead of using Django [generic relations](https://docs.djangoproject.com/en/5.0/ref/contrib/contenttypes/#generic-relations)). It could be the source of more bugs, but I think we already take care of non-existing process_parents. For 100% consistency, we should take care of the on_delete_cascade; we can do that in a different PR.

But what we cannot do is raise an IntegrityError after creating the object.
IntegrityError means the opposite: the record cannot be created because the related object does not exist, and we have to raise the exception before creating the object.

Alex-Izquierdo: I missed the atomic decorator.
all is ok then. :)

Alex-Izquierdo (approved):

src/aap_eda/tasks/activation_request_queue.py
```diff
@@ -12,19 +12,37 @@
...
+    if not model.objects.filter(id=parent_id).exists():
+        raise IntegrityError
```
Alex-Izquierdo, line 44: @jshimkus-rh in the exception we should say the reason: parent_id does not exist.

Alex-Izquierdo (approved):

bzwei (approved):

PR #935, opened 2024-06-11
fix: change to use ALLOW_LOCAL_RESOURCE_MANAGEMENT

bzwei:
Change DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED to ALLOW_LOCAL_RESOURCE_MANAGEMENT. Default to True which means EDA manages shared resource locally.

fixes AAP-25224


src/aap_eda/settings/default.py
```diff
@@ -654,17 +653,11 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
...
+RESOURCE_JWT_USER_ID = settings.get("RESOURCE_JWT_USER_ID", None)
+RESOURCE_SERVICE_PATH = settings.get("RESOURCE_SERVICE_PATH", None)
 
-if DIRECT_SHARED_RESOURCE_MANAGEMENT_ENABLED:
-    if (
-        RESOURCE_SERVER["URL"]
-        and RESOURCE_SERVER["SECRET_KEY"]
-        and RESOURCE_JWT_USER_ID
-    ):
+if not ALLOW_LOCAL_RESOURCE_MANAGEMENT:
```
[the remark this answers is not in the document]
bzwei: RESOURCE_SERVER will always evaluate to True. Unless we want to evaluate RESOURCE_SERVER["SECRET_KEY"]

src/aap_eda/settings/default.py
```diff
@@ -654,29 +653,17 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
...
+RESOURCE_SERVICE_PATH = settings.get("RESOURCE_SERVICE_PATH", None)
+
+if RESOURCE_SERVER["URL"] and RESOURCE_SERVER["SECRET_KEY"]:
+    RQ_PERIODIC_JOBS.append(
+        {
+            "func": "aap_eda.tasks.shared_resources.resync_shared_resources",
+            "interval": 900,
+            "id": "resync_shared_resources",
+        }
+    )
```
mkanoor, line 666: @bzwei Should we set ALLOW_LOCAL_RESOURCE_MANAGEMENT to False here? Which would mean that ALLOW_LOCAL_RESOURCE_MANAGEMENT is a setting that is controlled by the presence of RESOURCE_SERVER["URL"]
Do we also need to check if the RESOURCE_SERVER is set?

bzwei: @rochacbruno's previous commented not to depend on ALLOW_LOCAL_RESOURCE_MANAGEMENT. How do we check RESOURCE_SERVER since it is a dict with keys already. Here thus I directly check the keys.

src/aap_eda/settings/default.py
```diff
@@ -654,29 +653,17 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
...
-                "id": "resync_shared_resources",
-            }
-        )
-    else:
-        raise ImproperlyConfigured(
-            "RESOURCE_SERVER__URL, RESOURCE_SERVER__SECRET_KEY, "
-            "and RESOURCE_JWT_USER_ID settings must be properly configured"
-        )
+RESOURCE_JWT_USER_ID = settings.get("RESOURCE_JWT_USER_ID", None)
+RESOURCE_SERVICE_PATH = settings.get("RESOURCE_SERVICE_PATH", None)
```
Alex-Izquierdo, line 657: These values should not be validated as well? I understand that are mandatory if we configure the resource server, is it?

Alex-Izquierdo (commented):
Should we update also the docker-compose files, isn't?

src/aap_eda/settings/default.py
rochacbruno: Those 2 variables are now optional on latest DAB https://github.com/ansible/django-ansible-base/pull/442

src/aap_eda/settings/default.py
rochacbruno: Those 2 variables are meant to be independent.

- `RESOURCE_SERVER` - dict of settings to authenticate to the server
- `ALLOW_LOCAL_RESOURCE_MANAGEMENT` - Flag indicating if local service can mutate local settings or only RESOURCE_Server can do it.

src/aap_eda/settings/default.py
```diff
@@ -654,29 +653,17 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
...
+
+if RESOURCE_SERVER["URL"] and RESOURCE_SERVER["SECRET_KEY"]:
```
rochacbruno, line 659: Can `RESOURCE_SERVER` be an empty dict? if so it will raise `KeyError`

Maybe check using `.get` method?
```suggestion
if RESOURCE_SERVER.get("URL") and RESOURCE_SERVER.get("SECRET_KEY"):
```

rochacbruno (approved):

bzwei:
@Alex-Izquierdo Good catch. Updated
> Should we update also the docker-compose files, isn't?

Alex-Izquierdo (approved):

mkanoor (approved):

PR #945, opened 2024-06-18
feat: enable v1 root drf browsable and fix staticfiles

Alex-Izquierdo:
Enable the v1 root for DRF browsable and adds the missing collecstatic step. 
It requires updates in the installer/operator/ui_image to run the colllecstatic command and serve those static files. 
Jira: https://issues.redhat.com/browse/AAP-25508
Screenshot from 2024-06-18 18-42-08


rooftopcellist (dismissed):

mkanoor (dismissed):

Alex-Izquierdo:
Don't merge this for now, please. I want to sync the required changes in other places like the installer to avoid broken builds.

mkanoor (approved):

tests/integration/api/test_v1_root.py
```diff
@@ -0,0 +1,99 @@
...
+                "/event-streams/",
+                "/users/me/awx-tokens/",
+                "/credential-types/",
+                "/eda-credentials/",
+                "/decision-environments/",
+                "/organizations/",
+                "/teams/",
+            ],
+            True,
+            id="with_shared_resource",
```
jshimkus-rh, line 48: Is there a purpose to `id`?  There's not a third entry in the parameter list.

hsong-rh (approved):

tests/integration/api/test_v1_root.py
Alex-Izquierdo: @jshimkus-rh 
`id` is a parameter for pytest. It's the name of the test case. 
```
tests/integration/api/test_v1_root.py::test_v1_root[with_shared_resource] PASSED [ 45%]
tests/integration/api/test_v1_root.py::test_v1_root[no_shared_resource] PASSED [ 45%]
```

jshimkus-rh: Learn something new everyday.  I guess I never looked that close at the output from stuff I parameterized.

jshimkus-rh (approved):

Dostonbek1 (changes requested):
Looking at the JIRA ticket expected behavior, this PR doesn't seem to consider the following case:
```
/api/eda should provide version options for the API
```

Alex-Izquierdo:
> Looking at the JIRA ticket expected behavior, this PR doesn't seem to consider the following case:
> 
> ```
> /api/eda should provide version options for the API
> ```

It is out of the scope of this PR. As you can see in the description of the jira, only includes the v1/

Dostonbek1 (approved):

Alex-Izquierdo:
> Another minor comment, it seems a little redundant to have the hostname before every endpoint listed in v1 root view, e.g. `http://localhost:8000/api/eda/users/me/`. Since it is obvious what the hostname is, maybe we can drop it; it would make the endpoints more readable too. Thoughts?

It seems to me a valid point. I think it is due to the way I generate automatically all the links. I was trying to merge this PR in sync with other required changes to avoid breaking build. New changes in the PR would complicate this. Do you agree to merge as it is and later address the host in a later PR?

PR #946, opened 2024-06-18
chore: fully parameterize image specs in compose files

jshimkus-rh:


tools/docker/docker-compose-dev-redis-tls.yaml
```diff
@@ -216,10 +216,10 @@ services:
       retries: 3
       start_period: 5s
     depends_on:
-          - certs
+      - certs
 
   redis:
-    image: 'quay.io/fedora/redis-6:latest'
+    image: ${EDA_REDIS_IMAGE:-quay.io/fedora/redis-6}:${EDA_REDIS_VERSION:-latest}
```
mkanoor, line 222: @jshimkus-rh Why this image why not a multi arch image. The fedora image does not support amr64. We can run linux on arm64 in beaker vms.

Alex-Izquierdo: Ideally we should use images as closer as possible to RHEL. If the cpu is a problem, in the case of redis the docker.io base image could not be a big deal. But I would try to look first at some image in quay.io that can meet our requirements if it's possible, docker.io also has api rate limits and it is another credential that we have to handle in CI workflows, if we can avoid that, much better.

jshimkus-rh: This PR parameterizes the existing compose files.
Changing a/the specific default(s) is a separate concern.

Alex-Izquierdo (approved):

PR #948, opened 2024-06-21
fix: build the JOB URL on the server side

mkanoor:
With the advent of controller behind our server the URL to access the job details for JobTemplate and WorkflowTemplate has changed

We have an api end point that is specificed in the RH AAP Credential Type this is sent to the ansible-rulebook so that it can talk to the controller via the api. This end point could point to a controller directly or to a controller behind our server.

To directly access the controller we only need the host and port so the API end point would look like
e.g. https://<<your_controller>>/

This will map the Job URL for a Job Template to look like
     https://<<your_controller>>/#/jobs/playbook/23/details/
This will map the Job URL for a Workflow Template to look like
     https://<<your_controller>>/#/jobs/workflow/24/details/

When the controller is behind our server the API end point is e.g. https://<<your_server>>/api/controller

This will map the Job URL for a Job Template to look like
     https://<<your_server>>/execution/jobs/playbook/25/details/
This will map the Job URL for a Workflow Template to look like
     https://<<your_server>>/execution/jobs/workflow/26/details/

For this to work correctly we need ansible-rulebook to pass in the controller_job_id and the action_type. Currently it doesn't send up the controller_job_id which is being fixed in a separate [PR](https://github.com/ansible/ansible-rulebook/pull/698).

https://issues.redhat.com/browse/AAP-25604

### DIFF
--- a/src/aap_eda/wsapi/consumers.py
+++ b/src/aap_eda/wsapi/consumers.py
@@ -4,6 +4,7 @@ import logging
 import typing as tp
 from datetime import datetime
 from enum import Enum
+from urllib.parse import urlparse, urlunparse
 
 import yaml
 from channels.db import database_sync_to_async
@@ -224,10 +225,11 @@ class AnsibleRulebookConsumer(AsyncWebsocketConsumer):
         audit_rule = models.AuditRule.objects.filter(
             rule_uuid=message.rule_uuid, fired_at=message.rule_run_at
         ).first()
+
+        activation_instance = models.RulebookProcess.objects.filter(
+            id=message.activation_id
+        ).first()
         if audit_rule is None:
-            activation_instance = models.RulebookProcess.objects.filter(
-                id=message.activation_id
-            ).first()
             activation_org = models.Organization.objects.filter(
                 id=activation_instance.organization.id
             ).first()
@@ -259,11 +261,27 @@ class AnsibleRulebookConsumer(AsyncWebsocketConsumer):
         ).first()
 
         if audit_action is None:
+            inputs = {}
+            aap_credential_type = models.CredentialType.objects.filter(
+                name=DefaultCredentialType.AAP
+            )
+            if aap_credential_type:
+                credentials = (
+                    activation_instance.get_parent().eda_credentials.filter(
+                        credential_type_id=aap_credential_type[0].id
+                    )
+                )
+                if credentials:
+                    inputs = yaml.safe_load(
+                        credentials[0].inputs.get_secret_value()
+                    )
+
+            url = self._get_url(message, inputs)
             audit_action = models.AuditAction.objects.create(
                 id=message.action_uuid,
                 fired_at=message.run_at,
                 name=message.action,
-                url=message.url,
+                url=url,
                 status=message.status,
                 rule_fired_at=message.rule_run_at,
                 audit_rule_id=audit_rule.id,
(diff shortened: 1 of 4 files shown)


src/aap_eda/wsapi/messages.py
```diff
@@ -46,6 +46,7 @@ class ActionMessage(Message):
     delay: Optional[float]
     message: Optional[str]
     kind: Optional[str]
+    controller_job_id: Optional[str]
```
hsong-rh, line 49: What's the difference between this new `controller_job_id` and the existing `job_id`?

src/aap_eda/wsapi/consumers.py
```diff
@@ -417,3 +435,37 @@ def get_eda_system_vault_passwords(
...
         return vault_passwords
+
+    def _get_url(self, message: ActionMessage, inputs: dict) -> str:
+        if message.action not in ("run_job_template", "run_workflow_template"):
+            return ""
+        url = message.url
+        if not message.controller_job_id:
+            return url
+
+        api_url = inputs["host"]
```
Alex-Izquierdo, line 446: It seems that this method can be called with empty `inputs`. Should not be handled the potential absence of the "host" key?

tests/integration/wsapi/test_consumer.py
```diff
@@ -448,6 +448,119 @@ async def test_multiple_rules_for_one_event(
         assert await get_audit_event_action_count(event) == 2
 
 
+test_data = [
```
Alex-Izquierdo, line 451: Can you use a more specific name?

src/aap_eda/wsapi/messages.py
mkanoor: @Alex-Izquierdo I thought they are the same but it predates us and is a unique job identifier for every action that we run. Initially I thought it is the job id from controller only to realize its a UUID and not the controller job id

Alex-Izquierdo: Thanks for the clarification, but the question was made by @hsong-rh xD

src/aap_eda/wsapi/consumers.py
hsong-rh: For the input of credential type of AAP, the `hosts` is required

Alex-Izquierdo: the function can still be called with empty "inputs": https://github.com/ansible/eda-server/pull/948/files#diff-179935aa48d00084c8f2722f9b007cad6d09b24425ffe4121cc75f6c37b03c42R279

src/aap_eda/wsapi/consumers.py
```diff
@@ -417,3 +435,37 @@ def get_eda_system_vault_passwords(
...
+        path = urlparts.path.rstrip("/")
+        if path == "":
+            path = "/"
+        if path in settings.API_PATH_TO_UI_PATH_MAP:
+            path = settings.API_PATH_TO_UI_PATH_MAP[path]
+
+        if message.action == "run_job_template":
+            slug = f"{path}/jobs/playbook/{message.controller_job_id}/details/"
+        elif message.action == "run_workflow_template":
+            slug = f"{path}/jobs/workflow/{message.controller_job_id}/details/"
```
jshimkus-rh, line 458: Should be `if/else`.  The check at entry guarantees it's one or the other.
As written it gives the impression that `slug` may not be set before use in `urlunparse`.

Alex-Izquierdo (approved):

jshimkus-rh (approved):

PR #949, opened 2024-06-21
fix: add router_basename for models with compound words

bzwei:
It helps to reverse the url to access the resource

Needed by https://github.com/ansible/django-ansible-base/blob/devel/ansible_base/lib/abstract_models/common.py#L21

AAP-25487: Role user assignment response missing content_object except for Activations

### DIFF
--- a/src/aap_eda/core/models/credential_type.py
+++ b/src/aap_eda/core/models/credential_type.py
@@ -20,6 +20,8 @@ __all__ = ("CredentialType",)
 
 
 class CredentialType(BaseOrgModel, UniqueNamedModel):
+    router_basename = "credentialtype"
+
     class Meta:
         db_table = "core_credential_type"
         constraints = [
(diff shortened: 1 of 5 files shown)


Alex-Izquierdo (commented):
Why shouldn't we have this field defined in all our models?

bzwei:
> Why shouldn't we have this field defined in all our exposed models?
Adding this field to single word models does not change the resulting base name, but yes we can do this to all exposed models.

src/aap_eda/core/models/credential_type.py
```diff
@@ -20,6 +20,8 @@
 
 
 class CredentialType(BaseOrgModel, UniqueNamedModel):
+    router_basename = "credentialtype"
```
mkanoor, line 23: @bzwei Should this be a meta property?

bzwei: no. it is a class variable. I followed the example of DAB

mkanoor (approved):

hsong-rh (approved):

PR #950, opened 2024-06-24
fix: validate registry credential for completeness

mkanoor:
For container registry credential you need
 * username and password 
 or
  * password (token)

If one of the above conditions is not met we will treat it as a Bad Request and not create the Decision Environment

https://issues.redhat.com/browse/AAP-23517

### DIFF
--- a/src/aap_eda/api/serializers/decision_environment.py
+++ b/src/aap_eda/api/serializers/decision_environment.py
@@ -49,7 +49,7 @@ class DecisionEnvironmentCreateSerializer(serializers.ModelSerializer):
         required=False,
         allow_null=True,
         validators=[
-            validators.check_credential_types_for_registry,
+            validators.check_credential_registry_username_password,
         ],
     )
 
--- a/src/aap_eda/core/validators.py
+++ b/src/aap_eda/core/validators.py
@@ -77,14 +77,6 @@ def check_credential_types_for_activation(eda_credential_id: int) -> int:
     return eda_credential_id
 
 
-def check_credential_types_for_registry(eda_credential_id: int) -> int:
-    check_credential_types(
-        eda_credential_id, [enums.DefaultCredentialType.REGISTRY]
-    )
-
-    return eda_credential_id
-
-
 def check_credential_types_for_gpg(eda_credential_id: int) -> int:
     check_credential_types(
         eda_credential_id, [enums.DefaultCredentialType.GPG]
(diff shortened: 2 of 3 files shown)


src/aap_eda/core/validators.py
```diff
@@ -235,3 +235,23 @@ def check_credential_types(
...
+def check_credential_registry_username_password(
+    eda_credential_id: int,
+) -> int:
+    credential = get_credential_if_exists(eda_credential_id)
+    name = credential.credential_type.name
+    if name != enums.DefaultCredentialType.REGISTRY.value:
+        raise serializers.ValidationError(
+            "The type of credential can only be one of "
+            f"['{enums.DefaultCredentialType.REGISTRY.value}']"
+        )
```
jshimkus-rh, line 249: We should be able to simply use `enums.DefaultCredentialType.REGISTRY`.

If not that's an issue but unrelated to these changes.

src/aap_eda/core/validators.py
```diff
@@ -235,3 +235,23 @@ def check_credential_types(
...
+    inputs = yaml.safe_load(credential.inputs.get_secret_value())
+    username = inputs.get("username", "")
+    password = inputs.get("password", "")
+    if not username and not password:
+        raise serializers.ValidationError(
+            "Need username and password or just token in credential"
+        )
```
jshimkus-rh, line 256: This check would allow username by itself to be considered valid.

src/aap_eda/api/serializers/decision_environment.py
```diff
@@ -49,7 +49,7 @@ class DecisionEnvironmentCreateSerializer(serializers.ModelSerializer):
         required=False,
         allow_null=True,
         validators=[
-            validators.check_credential_types_for_registry,
```
hsong-rh, line 52: Should we remove this unused function [`check_credential_types_for_registry`](https://github.com/ansible/eda-server/pull/950/files#diff-0fc6d501fb19fb05efa10f75a58fc762f5aec1ecf8ecd80fa488d619f346b626R80)?

tests/integration/api/test_decision_environment.py
```diff
@@ -92,6 +92,55 @@ def test_create_decision_environment(
...
+            status.HTTP_400_BAD_REQUEST,
+            "Need username and password or just token in credential",
+        ),
+        (
+            {"host": "quay.io", "password": "token1"},
+            status.HTTP_201_CREATED,
+            "",
+        ),
+    ],
+)
```
jshimkus-rh, line 109: Add a set to test username w/o password.  Result should be status.HTTP_400_BAD_REQUEST.

mkanoor: Done

jshimkus-rh (approved):

hsong-rh (approved):

PR #951, opened 2024-06-25
fix: allow to update project organization

bzwei:
AAP-25825: The PATCH /projects<id> api does not update the organization_id

### DIFF
--- a/tests/integration/api/test_project.py
+++ b/tests/integration/api/test_project.py
@@ -586,14 +586,16 @@ def test_partial_update_project(
     new_project: models.Project,
     default_scm_credential: models.EdaCredential,
     default_gpg_credential: models.EdaCredential,
-    admin_client: APIClient,
+    superuser_client: APIClient,
 ):
     assert new_project.eda_credential_id is None
     assert new_project.signature_validation_credential_id is None
     assert new_project.verify_ssl is False
 
+    new_org = models.Organization.objects.create(name="new org")
     new_data = {
         "name": "new-project-updated",
+        "organization_id": new_org.id,
         "eda_credential_id": default_scm_credential.id,
         "signature_validation_credential_id": default_gpg_credential.id,
         "scm_branch": "main",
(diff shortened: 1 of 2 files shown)


tests/integration/api/test_project.py
```diff
@@ -586,29 +586,32 @@ def test_partial_update_project(
...
+        "organization_id": new_org.id,
         "eda_credential_id": default_scm_credential.id,
         "signature_validation_credential_id": default_gpg_credential.id,
         "scm_branch": "main",
         "scm_refspec": "ref1",
         "verify_ssl": True,
         "proxy": "http://user:$encrypted$@myproxy.com",
     }
-    response = admin_client.patch(
+    response = superuser_client.patch(
```
mkanoor, line 606: @bzwei why did we have to switch to superuser_client, wouldn't admin_client have sufficed. Are we short circuiting the RBAC check

jshimkus-rh (approved):

tests/integration/api/test_project.py
bzwei: pytest admin_client is created with roles in default org only.
This test is mainly focus on bringing back the missing field, not for rbac check.

hsong-rh (approved):

Dostonbek1:
I'm not quite sure if allowing changing organization for project makes sense to me. Following the RBAC hierarchy Organization -> Project, we are essentially migrating a project from one org to another. Even if we allow that, should the permission for that be more strict, e.g. only Admin user of both orgs can perform this change, not just anyone with change_project permission? @AlanCoding could you chime in, in case I'm missing something here?

bzwei:
If project does not allow to switch organization, what about other types such as DecisionEnviornment, EdaCredential, ad CredentialType? They all allow to switch organizations by the PATCH method. We need to be consistent.

Dostonbek1 (approved):
@bzwei Looking at the Controller behavior, we do allow changing the organization for projects, so the changes in this PR look good to me. As for RBAC policy, the user has to have `change` permission on the new organization that the project is moving to, which I think is the expected behavior.

PR #953, opened 2024-06-26
chore: upgrade deps for CVE's

Alex-Izquierdo:
Upgrade `requests sqlparse pydantic` deps to address the following CVE's:
* https://github.com/ansible/eda-server/security/dependabot/35
* https://github.com/ansible/eda-server/security/dependabot/33
* https://github.com/ansible/eda-server/security/dependabot/32

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -632,7 +632,7 @@ bcrypt = ["bcrypt"]
 
 [[package]]
 name = "django-ansible-base"
-version = "2024.9.17.0.dev9+gc8fbc1e"
+version = "2024.9.27.0.dev13+g5f1ca89"
 description = "A Django app used by ansible services"
 optional = false
 python-versions = ">=3.9"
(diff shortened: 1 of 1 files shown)


Dostonbek1:
Should we upgrade `requests` in [pyproject.toml](https://github.com/ansible/eda-server/blob/11122324aef27a73d6383489f25f0c166aa1c0de/pyproject.toml#L60) file too?

Alex-Izquierdo:
@Dostonbek1 
> Should we upgrade `requests` in [pyproject.toml](https://github.com/ansible/eda-server/blob/11122324aef27a73d6383489f25f0c166aa1c0de/pyproject.toml#L60) file too?

I think as it is now is fine. Requests is a nested dependency so we should rely on the constraints of the dependencies that depends on that. 

In our side it is defined just for the test dependencies, so that would not be critical. But even in that case the constraint is a wildcard and seems correct:
`requests = { version = "*", python = "<4.0" }`

Dostonbek1 (approved):

Alex-Izquierdo:
Closed in favor of https://github.com/ansible/eda-server/pull/1075

PR #955, opened 2024-06-27
fix: exclude system user from user list view

bzwei:
AAP-23939: EDA/AWX org-level role assignments return 400 for gateway users "_system" and "admin"

### DIFF
--- a/src/aap_eda/api/views/user.py
+++ b/src/aap_eda/api/views/user.py
@@ -241,6 +241,9 @@ class UserViewSet(
 
     def filter_queryset(self, qs):
         qs = visible_users(self.request.user, queryset=qs)
+        if not self.kwargs or "pk" not in self.kwargs:
+            # TODO: hardcode for now, will need to use settings.SYSTEM_USERNAME
+            qs = qs.exclude(username="_system")
         return super().filter_queryset(qs)
 
     def get_serializer_class(self):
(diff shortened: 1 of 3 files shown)


tests/integration/api/test_user.py
```diff
@@ -339,6 +349,7 @@ def test_list_users(
             "resource_type": admin_user.resource.resource_type,
         },
     }
+    assert results[0]["id"] != system_user.id
```
Alex-Izquierdo, line 352: This should not check for all the results instead of just the first one?

src/aap_eda/api/views/user.py
```diff
@@ -241,6 +241,9 @@ class UserViewSet(
 
     def filter_queryset(self, qs):
         qs = visible_users(self.request.user, queryset=qs)
+        if not self.kwargs or "pk" not in self.kwargs:
+            # TODO: hardcode for now, will need to use settings.SYSTEM_USERNAME
+            qs = qs.exclude(username="_system")
```
jshimkus-rh, line 246: Can we add `settings.SYSTEM_USERNAME` and default that to `_system`?

tests/integration/api/test_user.py
bzwei: there is a `len(results) == 1` check above

src/aap_eda/api/views/user.py
bzwei: No. It causes some other problem that needs to be addressed post GA.

Alex-Izquierdo (approved):

jshimkus-rh (approved):

Dostonbek1 (approved):
tested, lgtm.

PR #957, opened 2024-07-01
chore: support workers startup order with podman-compose

Alex-Izquierdo:
We rely on the healthcheck and depends_on::condition in our docker-compose files to ensure the correct startup order. Since podman-compose doesn't support it, add the wait_for_migrations command before the rq workers to ensure they are executed after the api pod and avoid errors like:
```
[eda-default-worker]    | django.db.utils.ProgrammingError: relation "core_activation_request_queue" does not exist
[eda-default-worker]    | LINE 1: ...re_activation_request_queue"."activation_id" FROM "core_acti...
```


tools/docker/docker-compose-mac.yml
```diff
@@ -96,10 +96,12 @@ services:
...
       - label=disable
     command:
-      - aap-eda-manage
-      - rqworker
-      - --worker-class
-      - aap_eda.core.tasking.DefaultWorker
+      - /bin/bash
+      - -c
+      - >-
+        aap-eda-manage wait_for_migrations -t ${EDA_WAIT_FOR_MIGRATIONS_TIMEOUT:-30};
```
hsong-rh, line 102: Should we add the same `wait_for_migration` line for [eda-ws](https://github.com/ansible/eda-server/pull/957/files#diff-356c78596b94c6e2943ec4bb62a483fa30a7ac113de4eede37310b0d38cbdfc0L83) in this file?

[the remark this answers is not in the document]
Alex-Izquierdo: added

Alex-Izquierdo: added

Alex-Izquierdo:
It seems the e2e fails because the import project task reach the timeout. I have seen other failures today, I feel CI is having today a bad network performance.

hsong-rh (approved):

jshimkus-rh (approved):

Alex-Izquierdo:
The multinode tests always fails with the same error related with project import. I have no idea why and it seems the change in the PR has not relation with it, but it fails. I need to investigate further why is failing.

Alex-Izquierdo:
It is a performance issue, increasing the timeout in the test suite solves the problem: https://github.com/ansible/eda-server/actions/runs/9764139814/job/26951732995

PR #958, opened 2024-07-01
fix: [AAP-26461] Disallow null organization id for project update

jshimkus-rh:


mkanoor:
@jshimkus-rh This seems to be similar for DE https://github.com/ansible/eda-server/blob/3c68bc2e64254c93fd6b2e74016705375d18d579/src/aap_eda/api/serializers/decision_environment.py#L45 Should we fix both of them Project/Decision Environment

jshimkus-rh:
> @jshimkus-rh This seems to be similar for DE
> 
> https://github.com/ansible/eda-server/blob/3c68bc2e64254c93fd6b2e74016705375d18d579/src/aap_eda/api/serializers/decision_environment.py#L45
> 
> Should we fix both of them Project/Decision Environment

@mkanoor @lgalis Looked at the controller UI and found that `organization` is only required for projects.  I'll query and see, but at the moment DE appears to be correct.

Alex-Izquierdo (approved):

Dostonbek1 (approved):

jshimkus-rh:
@mkanoor I'm merging this change as it has the necessary approvals.  We have another jira  for the other models and I'll put out a PR for that when/if we get blocker approval.

PR #963, opened 2024-07-08
feat: add filtering for superusers in users list view

Dostonbek1:
This PR considers the following 3 new cases in users list view:

1. `?is_superuser=true` - returns only superusers
2. `?is_superuser=false` - returns only non-superusers
3. no query param - returns all the users

### DIFF
--- a/src/aap_eda/api/views/user.py
+++ b/src/aap_eda/api/views/user.py
@@ -28,6 +28,7 @@ from rest_framework.viewsets import GenericViewSet
 
 from aap_eda.api import filters, serializers
 from aap_eda.core import models
+from aap_eda.utils import str_to_bool
 
 from .mixins import (
     CreateModelMixin,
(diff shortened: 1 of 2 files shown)


mkanoor:
@Dostonbek1 If Organization/User/Teams are going to come from Gateway why do we need this change in EDA. Is this when we won't be using the EDA product without Gateway?

src/aap_eda/api/views/user.py
```diff
@@ -259,6 +251,27 @@ def get_serializer_class(self):
     def get_response_serializer_class(self):
         return serializers.UserDetailSerializer
 
+    @extend_schema(
+        description="List all users",
```
jshimkus-rh, line 255:
```suggestion
        description="List users",
```

Alex-Izquierdo (approved):

jshimkus-rh (approved):

PR #964, opened 2024-07-09
fix: check to match the image_url of decision environment with its credential's host

hsong-rh:
[AAP-26580](https://issues.redhat.com/browse/AAP-26580)

The `image_url` will be validated with the registry's host when DE is created or updated. UI shows the errors like:
During creating
image
During updating
image

### DIFF
--- a/src/aap_eda/core/validators.py
+++ b/src/aap_eda/core/validators.py
@@ -45,17 +45,31 @@ def check_if_de_exists(decision_environment_id: int) -> int:
     try:
         de = models.DecisionEnvironment.objects.get(pk=decision_environment_id)
         if de.eda_credential_id:
-            models.EdaCredential.objects.get(pk=de.eda_credential_id)
-    except models.EdaCredential.DoesNotExist:
-        raise serializers.ValidationError(
-            f"EdaCredential with id {de.eda_credential_id} does not exist"
-        )
+            check_if_de_valid(de.image_url, de.eda_credential_id)
     except models.DecisionEnvironment.DoesNotExist:
         raise serializers.ValidationError(
             f"DecisionEnvironment with id {decision_environment_id} "
             "does not exist"
         )
     return decision_environment_id
+
+
+def check_if_de_valid(image_url: str, eda_credential_id: int):
+    credential = get_credential_if_exists(eda_credential_id)
+    inputs = yaml.safe_load(credential.inputs.get_secret_value())
+    host = inputs.get("host")
+
+    if not host:
+        raise serializers.ValidationError(
+            f"Credential {credential.name} needs to have host information"
+        )
+
+    if not image_url.startswith(host):
+        msg = (
+            f"DecisionEnvironment image url: {image_url} does "
+            f"not match with the credential host: {host}"
+        )
+        raise serializers.ValidationError(msg)
 
 
 def get_credential_if_exists(eda_credential_id: int) -> models.EdaCredential:
--- a/src/aap_eda/services/activation/activation_manager.py
+++ b/src/aap_eda/services/activation/activation_manager.py
@@ -1022,7 +1022,19 @@ class ActivationManager(StatusManager):
 
     def _get_container_request(self) -> ContainerRequest:
         try:
-            return self.db_instance.get_container_request()
+            request = self.db_instance.get_container_request()
+
+            if request.credential and not request.image_url.startswith(
+                request.credential.host
+            ):
+                msg = (
+                    f"DecisionEnvironment image url: {request.image_url} does "
+                    "not match with the credential host: "
+                    f"{request.credential.host}"
+                )
+                raise engine_exceptions.ContainerRequestError(msg)
+
+            return request
         except ContainerableInvalidError:
             msg = (
                 f"Activation {self.db_instance.id} not valid, "
--- a/src/aap_eda/services/activation/activation_manager.py
+++ b/src/aap_eda/services/activation/activation_manager.py
@@ -1022,19 +1022,7 @@ class ActivationManager(StatusManager):
 
     def _get_container_request(self) -> ContainerRequest:
         try:
-            request = self.db_instance.get_container_request()
-
-            if request.credential and not request.image_url.startswith(
-                request.credential.host
-            ):
-                msg = (
-                    f"DecisionEnvironment image url: {request.image_url} does "
-                    "not match with the credential host: "
-                    f"{request.credential.host}"
-                )
-                raise engine_exceptions.ContainerRequestError(msg)
-
-            return request
+            return self.db_instance.get_container_request()
         except ContainerableInvalidError:
             msg = (
                 f"Activation {self.db_instance.id} not valid, "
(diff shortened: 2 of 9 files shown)


mkanoor:
@hsong-rh Can this be done when the Decision Environment is being created/updated. Since that is the first validation error we should be able to catch

src/aap_eda/services/activation/activation_manager.py
```diff
@@ -1022,7 +1022,19 @@ def _get_queue_name(self) -> str:
...
+
+            if request.credential and not request.image_url.startswith(
+                request.credential.host
+            ):
+                msg = (
+                    f"DecisionEnvironment image url: {request.image_url} does "
+                    "not match with the credential host: "
+                    f"{request.credential.host}"
+                )
+                raise engine_exceptions.ContainerRequestError(msg)
```
Alex-Izquierdo, line 1035: Maybe I'm missing something, Why this validation must happen at the container request? It seems to me a mixture of responsibilities. I would say it should happen at processing the request, the fixture itself seems to reveal this point:
```
    aap_credential_type = models.CredentialType.objects.get(
        name=DefaultCredentialType.AAP,
    )
    credential = models.EdaCredential.objects.create(
        name="mismatch_credential",
        inputs={
            "host": "mismatch.io",
            "username": "Fred",
            "password": "secret",
        },
        credential_type=aap_credential_type,
    )
    return models.DecisionEnvironment.objects.create(
        name="test-decision-environment",
        image_url="localhost:14000/test-image-url",
        eda_credential_id=credential.id,
    )
 ```

Comment (hsong-rh) [reply src/aap_eda/services/activation/activation_manager.py]: Changed to use DE's validator

File: tests/integration/api/test_decision_environment.py [hunk 2]
```diff
@@ -245,6 +253,66 @@ def test_partial_update_decision_environment(
...
+                "password": "secret",
+                "host": "registry.com",
+            },
+            status.HTTP_400_BAD_REQUEST,
+            "does not match with the credential host",
+        ),
+    ],
+)
+@pytest.mark.django_db
+def test_partial_update_decision_environment_with_image_url_and_host(
```
Alex-Izquierdo, line 285: Should we have a test as well when creating a DE?

Alex-Izquierdo (commented):
Can the credential be modified after it has been associated with some DE?

tests/integration/api/test_decision_environment.py
hsong-rh: Added already at https://github.com/ansible/eda-server/pull/964/files#diff-66e5a86ad1ff6de7358f68e2e87359606462164415f5456570e127256188e313R110

chrismeyersfsu:
Chatted with @mkanoor outside of here. I was wrong about  this not being verifiable statically. You can.

**Example Happy Path:**
* EDA Cred w/ url my-registry.com
* EDA Decision env with container my-registry.com/foo/bar:latest

User is allowed to create a decision env :+1: 

**Example Malicous Path #1:**
* EDA Cred w/ url my-registry.com
* EDA Decision env with container my-malicious-registry.com/foo/bar:latest

User NOT allowed to create the decision env.

**Example Malicous Path #2:**
* EDA Cred w/ url my-registry.com
* EDA Decision env with container my-registry.com/foo/bar:latest
* User edits Cred to be `my-other-registry.com`

What should happen here?

hsong-rh:
> **Example Malicous Path #2:**
> 
> * EDA Cred w/ url my-registry.com
> * EDA Decision env with container my-registry.com/foo/bar:latest
> * User edits Cred to be `my-other-registry.com`
> 
> What should happen here?

@chrismeyersfsu In this scenario, new start or restart activations with the DE will fail.
Restarting an activation with the changed Cred will popup the error message window:
image
Starting a new activation with the changed Cred:
image

src/aap_eda/core/validators.py
```diff
@@ -45,7 +45,7 @@ def check_if_de_exists(decision_environment_id: int) -> int:
     try:
         de = models.DecisionEnvironment.objects.get(pk=decision_environment_id)
         if de.eda_credential_id:
-            models.EdaCredential.objects.get(pk=de.eda_credential_id)
+            check_if_de_valid(de.image_url, de.eda_credential_id)
     except models.EdaCredential.DoesNotExist:
```
bzwei, line 49: seems we longer need to catch this exception since `get_credential_if_exists()` does it already

hsong-rh: removed

bzwei (dismissed):

src/aap_eda/core/validators.py
```diff
@@ -45,11 +45,7 @@ def check_if_de_exists(decision_environment_id: int) -> int:
     try:
         de = models.DecisionEnvironment.objects.get(pk=decision_environment_id)
         if de.eda_credential_id:
-            models.EdaCredential.objects.get(pk=de.eda_credential_id)
```
mkanoor, line 48: @hsong-rh What happened to this check why did it get removed. If we have a bogus id there we still need to catch it during validation

mkanoor: Never mind I see that other function is doing the check 👍

src/aap_eda/core/validators.py
```diff
@@ -58,6 +54,19 @@ def check_if_de_exists(decision_environment_id: int) -> int:
     return decision_environment_id
 
 
+def check_if_de_valid(image_url: str, eda_credential_id: int):
+    credential = get_credential_if_exists(eda_credential_id)
+    inputs = yaml.safe_load(credential.inputs.get_secret_value())
+    host = inputs.get("host")
+
+    if host and not image_url.startswith(host):
```
benthomasson, line 62: Raise an exception if host is empty

benthomasson (changes requested):

src/aap_eda/core/validators.py
mkanoor: @benthomasson Host is a required parameter in the object https://github.com/ansible/eda-server/blob/a7ffc3397288f944088d6bbf44dfb9a72a7cb7bb/src/aap_eda/core/management/commands/create_initial_data.py#L210 Without host we won't be able to create a registry credential in the DB

benthomasson: I'd still like this to be enforced here in the check.

hsong-rh: Ok, I'll add it

src/aap_eda/core/validators.py
```diff
@@ -58,6 +54,24 @@ def check_if_de_exists(decision_environment_id: int) -> int:
     return decision_environment_id
 
 
+def check_if_de_valid(image_url: str, eda_credential_id: int):
+    credential = get_credential_if_exists(eda_credential_id)
+    inputs = yaml.safe_load(credential.inputs.get_secret_value())
+    host = inputs.get("host")
```
mkanoor, line 60: @hsong-rh What happens if the "host" key is missing, do you want to use empty string if its missing.

hsong-rh: I changed to raise an exception if `hosts` is null

src/aap_eda/core/validators.py
hsong-rh: @benthomasson the raising exception is added when host is empty

benthomasson (approved):

bzwei (approved):

mkanoor (approved):

PR #966, opened 2024-07-10
feat: change to using filter class for is_superuser filter

Dostonbek1:
This is a followup to the PR https://github.com/ansible/eda-server/pull/963 to switch to using `UserFilter` to achieve the same purpose and to keep the code cleaner.


Alex-Izquierdo (approved):
Good catch!

Alex-Izquierdo (approved):

jshimkus-rh (approved):

PR #971, opened 2024-07-12
chore: fix dependency for podman

mkanoor:
All of our workers have a dependency on the api-worker which has dependency to postgres and redis.
We dont need to have default worker and activation workers have a direct dependency to postgres and redis. They can have a direct dependency to api-worker.

In podman with the --requires if we have a direct dependency to redis and postgres from default and activation workers they might come up before even redis is queued up and break the dependency chain. Causing them to not start.

```
[eda-scheduler]         | Error: unable to start container c979979377dd402dacee7856328c205810362bf3abdb5c85d0e291ecb2ca6e45: generating dependency graph for container c979979377dd402dacee7856328c205810362bf3abdb5c85d0e291ecb2ca6e45: container 4f8bfe666edea37dd185d51066e09d5fcb512db3dcd2608f3acc2b5f106f9ad9 depends on container 92cc8f498d3dba3a99ca1b649fa62ba03fd5663891dc32217cd79ad2573e2315 not found in input list: no such container
[eda-activation-worker] | Error: unable to start container 2edd9cab6bcf841db30689f6379a992b0e2ecb6b7befb3a5eca9d77f945f89f8: generating dependency graph for container 2edd9cab6bcf841db30689f6379a992b0e2ecb6b7befb3a5eca9d77f945f89f8: container 4f8bfe666edea37dd185d51066e09d5fcb512db3dcd2608f3acc2b5f106f9ad9 depends on container 92cc8f498d3dba3a99ca1b649fa62ba03fd5663891dc32217cd79ad2573e2315 not found in input list: no such container
```

### DIFF
--- a/tools/docker/docker-compose-stage.yaml
+++ b/tools/docker/docker-compose-stage.yaml
@@ -54,15 +54,34 @@ services:
     depends_on:
      - podman-pre-setup
 
-  eda-ui:
-    image: ${EDA_UI_IMAGE:-quay.io/ansible/eda-ui}:${EDA_UI_VERSION:-main}
-    environment: *common-env
+  postgres:
+    image: ${EDA_POSTGRES_IMAGE:-quay.io/sclorg/postgresql-15-c9s}:${EDA_POSTGRES_VERSION:-latest}
+    environment:
+      POSTGRESQL_USER: eda
+      POSTGRESQL_PASSWORD: secret
+      POSTGRESQL_ADMIN_PASSWORD: secret
+      POSTGRESQL_DATABASE: eda
     ports:
-      - '${EDA_UI_PORT:-8443}:443'
-    depends_on:
-      eda-api:
-        condition: service_healthy
+      - '${EDA_PG_PORT:-5432}:5432'
+    volumes:
+      - 'postgres_data:/var/lib/pgsql/data'
+    healthcheck:
+      test: [ 'CMD', 'pg_isready', '-U', 'postgres' ]
+      interval: 5s
+      timeout: 5s
+      retries: 3
+      start_period: 5s
 
+  redis:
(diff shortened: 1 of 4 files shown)


jshimkus-rh (approved):

mkanoor:
Another thing that I have noticed is the way Podman-compose works it reads the file from top to bottom and dispatches asyncio tasks for each one of the services. And our current layout we add Postgres and redis at the end while eda-ui is at the top. So I am wondering if I should change the order in the docker-compose so that the core dependencies Postgres, redis, Eda-api come before the rest of the services. If Podman-compose loaded the service list and then created a graph and started the leaf nodes that don't have any dependent services it would be ideal. docker-compose might be doing that but not podman-compose. So with the re-arranged service nodes both docker-compose and Podman-compose should work.

Dostonbek1 (approved):

hsong-rh (approved):

tools/docker/docker-compose-stage.yaml
```diff
@@ -201,6 +179,15 @@ services:
     ports:
       - '${EDA_PROXY_PORT:-3128}:3128'
 
+  eda-ui:
+    image: ${EDA_UI_IMAGE:-quay.io/ansible/eda-ui}:${EDA_UI_VERSION:-main}
+    environment: *common-env
+    ports:
+      - '${EDA_UI_PORT:-8443}:443'
+    depends_on:
```
Alex-Izquierdo, line 187: Do we really need the conditional for the UI? The UI is agnostic and it is not really dependent on the api.

mkanoor: @Alex-Izquierdo I have seen that the nginx embedded inside the UI image tries to reach out and fails. Not sure if that is when you have an idling browser window that tries to send in the /me call when the system is just coming back up.

Alex-Izquierdo (approved):

PR #972, opened 2024-07-16
feat: add config/ endpoint to display eda-server version

Dostonbek1:
- add a new endpoint `api/eda/v1/config/` to retrieve eda-server version

JIRA: https://issues.redhat.com/browse/AAP-22119

NOTE: UI will need changes to read version data from this new endpoint. cc @lgalis


src/aap_eda/api/views/root.py
```diff
@@ -70,3 +71,12 @@ def list_urls(urls):
...
 
     return list_urls(urls)
+
+
+class ApiV1ConfigView(APIView):
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request):
+        from aap_eda import __version__
+
```
mkanoor, line 81: @Dostonbek1 We might need to send back more information like deployment type etc since that helps the UI team block out UI elements specific to deployment type.
https://github.com/ansible/eda-server/blob/55e8897c3be3e7f537fe0b1c9ee219303fbc6d07/src/aap_eda/settings/default.py#L340 maybe we can create a Dataclass with the required fields and send that out as json.

Dostonbek1: yup, agreed. I was mainly trying to capture the minimum requirement asked in this ticket but I think adding extra data shouldn't be too hard and we can include them in this PR. Here is the extra data we can include:
```
- version
- deployment type (podman or kubernetes)
- controller url
```
can you think of anything else we should add?

mkanoor: @Dostonbek1 The current config end point in AAP has the following data
/api/v2/config/
```
{
    "time_zone": "UTC",
    "license_info": {
        "license_type": "open",
        "valid_key": true,
        "subscription_name": "OPEN",
        "product_name": "AWX"
    },
    "version": "24.1.1.dev19+gb022b50966.d20240404",
    "eula": "",
    "analytics_status": "off",
....

```

mkanoor: @Dostonbek1 Can we check with the UI team which end point they would like to call and also where should the version string be coming from. The installer team might be using a different value. This should work for both upstream and downstream.

mkanoor: @Dostonbek1 Controller URL is being deprecated in favor of RH AAP Credential type it won't be there going forward

src/aap_eda/utils/__init__.py
```diff
@@ -12,6 +12,16 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import pkg_resources
```
mkanoor, line 15: @Dostonbek1 This seems to have been deprecated in favor of importlib https://docs.python.org/3/library/importlib.metadata.html#module-importlib.metadata
```
from importlib.metadata import version 
version('psycopg')
'3.2.1'
version('aap-eda')
'0.2.0'
```

mkanoor: If the python package is not installed it throws **PackageNotFoundError(name)** Maybe at that point we could directly try loading the pyproject.toml and directly read the version

mkanoor:
@Dostonbek1 Should we add [Timezone](https://github.com/ansible/eda-server/blob/55e8897c3be3e7f537fe0b1c9ee219303fbc6d07/src/aap_eda/settings/default.py#L297) to be consistent with AAP

src/aap_eda/utils/__init__.py
```diff
@@ -11,7 +11,36 @@
...
 def str_to_bool(value: str) -> bool:
     return value.lower() in ("yes", "true", "1")
+
+
+def get_eda_version():
+    """Return EDA version as defined in the package or pyproject.toml."""
+    try:
+        return version("aap-eda")
+    except PackageNotFoundError:
+        return get_version_from_file()
```
mkanoor, line 31: @Dostonbek1 Can this just be "N/A". Are you seeing this only during pytest? In all other scenarios
* Container Images have the package aap-eda package installed
* The RPM packages should have the aap-eda package installed or does it have a different name.

Dostonbek1: I mean it could but as we saw in container instance, if aap-eda package is not available for whatever reason then we would resort to getting it directly from pyproject.toml file.

src/aap_eda/utils/__init__.py
```diff
@@ -11,7 +11,33 @@
...
```
Alex-Izquierdo, line 29: The logic of `get_version_from_file` is not reliable because it assumes that there is a repo or somehow a pyproject file without a package installed which is a weird scenario. 

I suggest to simplify the logic and rely only on importlib, which is the expected scenario, and returns a string like "package not installed properly" in case of `PackageNotFoundError`

In next iterations we can implement version in `aap_eda/__init__`, that should be a better fallback for `importlib.metadata.version` But I think there is no reason to support such a scenario where the package is not installed.

src/aap_eda/utils/__init__.py
```diff
@@ -11,7 +11,33 @@
...
+from pathlib import Path
+
+import toml
 
 
 def str_to_bool(value: str) -> bool:
     return value.lower() in ("yes", "true", "1")
+
+
+def get_eda_version():
```
Alex-Izquierdo, line 24: type hints, please?

tests/integration/api/test_root.py
```diff
@@ -109,3 +112,14 @@ def test_api_root(admin_client):
     assert "v1" in response.data["available_versions"]
     assert api_url_v1 in response.data["current_version"]
     assert api_url_v1 in response.data["available_versions"]["v1"]
+
+
+@pytest.mark.django_db
```
Alex-Izquierdo, line 117: I think this test should be placed in their own file for the config endpoint.

Alex-Izquierdo (dismissed):

Alex-Izquierdo (approved):

mkanoor (approved):

PR #973, opened 2024-07-18
fix: add missing openapi response schema to /status

Alex-Izquierdo:


src/aap_eda/core/views.py
jshimkus-rh, line 1: Might as well use the imported status for the values in HealthView.

Alex-Izquierdo: No follow, sorry.

jshimkus-rh: The HealthView class, in the same file, is using the literals `200` and `500`.  Let's change that to use the same `status` attributes.

Alex-Izquierdo: I think that's out of the scope of this PR.

jshimkus-rh: That then makes the critique the new code is inconsistent with the existing code.

Alex-Izquierdo: There are more inconsistencies like this in the whole app. Do you want to address all them here? Or it is just that file? Do you want to keep the old literal?

Alex-Izquierdo: HealthView changed.

jshimkus-rh (approved):

Dostonbek1 (approved):

PR #974, opened 2024-07-18
fix: correct trailing slash for api docs endpoints and update tests

Alex-Izquierdo:
Currently only works https://localhost:8443/api/eda/v1/docs, docs/ returns 404
This PR fixes the correct handling of the trailing slash to keep it consistent with the rest of the application, so /docs is a 301 to -> /docs/ and /docs/ works as expected
Also adds sanity tests for all these urls.

### DIFF
--- a/tests/integration/api/test_openapi.py
+++ b/tests/integration/api/test_openapi.py
@@ -1,0 +1,52 @@
+#  Copyright 2024 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+
+import pytest
+import yaml
+from django.conf import settings
+from rest_framework import status
+
+from tests.integration.constants import api_url_v1
+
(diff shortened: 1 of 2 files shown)


Dostonbek1 (approved):
I observed this a while ago but never got around to pushing a fix for it. Thank you for putting up the PR. 👍

tests/integration/api/test_openapi.py
```diff
@@ -0,0 +1,51 @@
...
+    "path",
+    [
+        pytest.param("/openapi.json", id="openapi_json"),
+        pytest.param("/openapi.yaml", id="openapi_yaml"),
+    ],
+)
+@pytest.mark.django_db
+def test_v1_openapi(admin_client, path):
+    response = admin_client.get(f"{api_url_v1}{path}")
+    assert response.status_code == 200
```
jshimkus-rh, line 35: I'd prefer the use of the `status` value from `rest_framework` over a hardcoded value.

tests/integration/api/test_openapi.py
```diff
@@ -0,0 +1,51 @@
...
+@pytest.mark.parametrize(
+    "path",
+    [
+        pytest.param("/docs/", id="swagger_ui"),
+        pytest.param("/redoc/", id="redoc"),
+    ],
+)
+def test_v1_openapi_ui(admin_client, path):
+    response = admin_client.get(f"{api_url_v1}{path}")
+    assert response.status_code == 200
```
jshimkus-rh, line 50: I'd prefer the use of the `status` value from `rest_framework` over a hardcoded value.

tests/integration/api/test_openapi.py
Alex-Izquierdo: changed

tests/integration/api/test_openapi.py
Alex-Izquierdo: changed

jshimkus-rh (approved):

PR #976, opened 2024-07-19
fix: set ANSIBLE_LOCAL_TEMP for ansible-runner

bzwei:
AAP-26860: unable to create local directories (/.ansible/tmp)

### DIFF
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -93,6 +93,7 @@ To configure a Resource Server for syncing of managed resources:
 
 """
 import os
+import tempfile
 from datetime import timedelta
 
 import dynaconf
@@ -696,3 +697,20 @@ SAFE_PLUGINS_FOR_PORT_FORWARD = settings.get(
 API_PATH_TO_UI_PATH_MAP = settings.get(
     "API_PATH_UI_PATH_MAP", {"/api/controller": "/execution", "/": "/#"}
 )
+
+
+ANSIBLE_LOCAL_TEMP = settings.get("ANSIBLE_LOCAL_TEMP", tempfile.gettempdir())
+if not os.access(ANSIBLE_LOCAL_TEMP, os.W_OK):
+    raise ImproperlyConfigured(
+        f"ANSIBLE_LOCAL_TEMP {ANSIBLE_LOCAL_TEMP} is not writable, "
+        "please define ANSIBLE_LOCAL_TEMP with a writable directory."
+    )
+
+ANSIBLE_REMOTE_TEMP = settings.get(
+    "ANSIBLE_REMOTE_TEMP", tempfile.gettempdir()
+)
+if not os.access(ANSIBLE_REMOTE_TEMP, os.W_OK):
+    raise ImproperlyConfigured(
+        f"ANSIBLE_REMOTE_TEMP {ANSIBLE_LOCAL_TEMP} is not writable, "
+        "please define ANSIBLE_REMOTE_TEMP with a writable directory."
+    )
--- a/src/aap_eda/api/vault.py
+++ b/src/aap_eda/api/vault.py
@@ -16,6 +16,7 @@ import shutil
 import tempfile
 
 import pexpect
+from django.conf import settings
 
 
 class AnsibleVaultNotFound(Exception):
@@ -41,8 +42,11 @@ def encrypt_string(password: str, plaintext: str, vault_id: str) -> str:
     tmp.write(password)
     tmp.flush()
     label = f"{vault_id}@{tmp.name}"
+    env = os.environ | {"ANSIBLE_LOCAL_TEMP": settings.ANSIBLE_LOCAL_TEMP}
 
-    child = pexpect.spawn(f"ansible-vault encrypt_string --vault-id {label}")
+    child = pexpect.spawn(
+        f"ansible-vault encrypt_string --vault-id {label}", env=env
+    )
     child.expect("Reading plaintext input from stdin*")
     child.sendline(plaintext)
     child.sendcontrol("D")
@@ -75,7 +79,8 @@ def encrypt_string(password: str, plaintext: str, vault_id: str) -> str:
 
 
 def decrypt(password: str, vault_string: str) -> str:
-    child = pexpect.spawn("ansible-vault decrypt")
+    env = os.environ | {"ANSIBLE_LOCAL_TEMP": settings.ANSIBLE_LOCAL_TEMP}
+    child = pexpect.spawn("ansible-vault decrypt", env=env)
     child.expect("Vault password: ")
     child.sendline(password)
     child.expect("Reading ciphertext input from stdin")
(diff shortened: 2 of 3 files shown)


jshimkus-rh (dismissed):

Alex-Izquierdo (dismissed):

mkanoor (changes requested):
@bzwei This needs to be fixed in multiple places, we are missing it for vaults

mkanoor:
@bzwei the ansible-vault can also fail here https://github.com/ansible/eda-server/blob/24cd4b3e3c3ca19e83ddf3d6ef73a37a82ceea07/src/aap_eda/api/vault.py#L44
```
2024-06-27 16:09:57,986 - ansible_rulebook.engine - ERROR - Source error End Of File (EOF). Exception style platform.
<pexpect.pty_spawn.spawn object at 0x7f37ebc0fe20>
command: /app/.local/bin/ansible-vault
args: ['/app/.local/bin/ansible-vault', 'decrypt', '--vault-id', 'EDA_SERVER@/tmp/tmpz_ulnc4x']
buffer (last 100 chars): b''
before (last 100 chars): b"ible-local-63puar1zl3'. [Errno 13] Permission denied: '/app/.ansible/tmp/ansible-local-63puar1zl3'\r\n"
after: <class 'pexpect.exceptions.EOF'>
```
The above is from ansible-rulebook which does the decrypt using ansible-vault and on the eda-server side using the same executable we are doing the encryption

jshimkus-rh (dismissed):

Alex-Izquierdo (dismissed):

src/aap_eda/settings/default.py
```diff
@@ -696,3 +696,6 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
 API_PATH_TO_UI_PATH_MAP = settings.get(
     "API_PATH_UI_PATH_MAP", {"/api/controller": "/execution", "/": "/#"}
 )
+
+
+ANSIBLE_LOCAL_TEMP = settings.get("ANSIBLE_LOCAL_TEMP", "/tmp")
```
mkanoor, line 701: @bzwei I think we should be using the python temp directory function, if the tmpfs is marked as readonly this will still fail.

mkanoor: https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir

mkanoor (dismissed):

src/aap_eda/settings/default.py
```diff
@@ -696,3 +697,10 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
 API_PATH_TO_UI_PATH_MAP = settings.get(
     "API_PATH_UI_PATH_MAP", {"/api/controller": "/execution", "/": "/#"}
 )
+
+
+ANSIBLE_LOCAL_TEMP = settings.get("ANSIBLE_LOCAL_TEMP", tempfile.gettempdir())
+if not os.access(ANSIBLE_LOCAL_TEMP, os.W_OK):
+    raise ImproperlyConfigured(
+        f"ANSIBLE_LOCAL_TEMP ${ANSIBLE_LOCAL_TEMP} is not writable"
+    )
```
Alex-Izquierdo, line 706: This can end up in a misleading message because `ANSIBLE_LOCAL_TEMP` could not be defined and the default value not writable. Also the message has a misleading `$` char, maybe a typo from bash syntax? 

I suggest instead:

```suggestion
    raise ImproperlyConfigured(
        f"{ANSIBLE_LOCAL_TEMP} is not writable, please define ANSIBLE_LOCAL_TEMP with a writable directory."
    )
```

src/aap_eda/settings/default.py
```diff
@@ -696,3 +697,11 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
 API_PATH_TO_UI_PATH_MAP = settings.get(
     "API_PATH_UI_PATH_MAP", {"/api/controller": "/execution", "/": "/#"}
 )
+
+
+ANSIBLE_LOCAL_TEMP = settings.get("ANSIBLE_LOCAL_TEMP", tempfile.gettempdir())
+if not os.access(ANSIBLE_LOCAL_TEMP, os.W_OK):
+    raise ImproperlyConfigured(
+        f"ANSIBLE_LOCAL_TEMP ${ANSIBLE_LOCAL_TEMP} is not writable, "
```
Alex-Izquierdo, line 705: Why the dolar sign? I think it is a typo

bzwei:
Turn to draft until we're allowed to merge it

src/aap_eda/settings/default.py
```diff
@@ -696,3 +697,20 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
...
+
+
+ANSIBLE_LOCAL_TEMP = settings.get("ANSIBLE_LOCAL_TEMP", tempfile.gettempdir())
+if not os.access(ANSIBLE_LOCAL_TEMP, os.W_OK):
+    raise ImproperlyConfigured(
+        f"ANSIBLE_LOCAL_TEMP {ANSIBLE_LOCAL_TEMP} is not writable, "
+        "please define ANSIBLE_LOCAL_TEMP with a writable directory."
+    )
+
+ANSIBLE_REMOTE_TEMP = settings.get(
```
Alex-Izquierdo, line 709: I'm concerned to use the same default value for ANSIBLE_REMOTE_TEMP and ANSIBLE_LOCAL_TEMP just in case there might be conflicts, have different purposes. 

I don't see ANSIBLE_REMOTE_TEMP in https://docs.ansible.com/ansible/latest/reference_appendices/config.html
Seems that is [going to be deprecated](https://github.com/ansible/ansible/blob/stable-2.5/changelogs/CHANGELOG-v2.5.rst#minor-changes-4)

I think it is better to define the variable in the inventory instead of an envvar and use a tempdir created on demand. Thoughts?

bzwei: Actual tests proved that we need to set ANSIBLE_REMOTE_TEMP.

Since we're already a creating temp folder for project sync, we can just create new temp folders for ANSIBLE too. This will eliminate the needs to set temp folder for ANSIBLE in the settings.

Alex-Izquierdo: I'm not saying you don't have to set the remote temp, I'm proposing to define the variable in the inventory/playbook, but in anycase it should be better ANSIBLE_REMOTE_TMP, because ANSIBLE_REMOTE_TEMP [will be deprecated](https://github.com/ansible/ansible/blob/stable-2.5/changelogs/CHANGELOG-v2.5.rst#minor-changes-4#:~:text=ANSIBLE_REMOTE_TMP).  It would be a more robust solution in the long term. 
Plus a temp dir instead of a static setting.

Ref: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/sh_shell.html#parameter-remote_tmp

PR #715, opened 2024-03-06
Add eda to the service mesh network

chrismeyersfsu:
* Join a common named network that components can discover one another and talk on. Related to https://github.com/ansible/aap-gateway/pull/235

PR #978, opened 2024-07-20
Bring -stage docker compose in line with dev file

AlanCoding:
Original objective: the [aap-dev environment](https://github.com/ansible/aap-gateway/tree/devel/aap-dev) was spinning up 2 containers for eda-task and other types of containers. Overall, it runs way too many containers, and I wanted to make it simple and more understandable.

I had a conversation in Slack about this:

https://redhat-internal.slack.com/archives/C02F01HBCBZ/p1721225412790029

And @mkanoor suggested that I directly run `docker compose up`, which is what I would prefer anyway! The root of the problem is that the -dev file hard-codes 2 replicas for things. I am not attempting to solve that here.

The `tools/docker/docker-compose-stage.yaml` lets me control the number of replicas, which is what I'd like. So I tried to run aap-dev with that, and I hit some problems, mainly that the settings that the aap-dev Makefile produces weren't getting picked up, and this is because the -dev file templates that, but the -stage file doesn't. @chrismeyersfsu also added a network to the -dev file which was missing in -stage.

There isn't much reason the -dev and -stage files should be any difference _except_ for `volumes:`, which I admit, can't easily be addressed by templates inside the YAML itself. So merging the files might not be possible. I get that.

So for now I just want do to a one-off update, and I'll link the aap-dev patch to put the whole picture together.


Alex-Izquierdo (commented):
The main goal of the -stage file is to run the application easily with a prebuilt images and closer to a production setup. 

I don't see problems making `eda_host` configurable or adding the `service-mesh` network but if you want to build the image from the source code and apply by default `aap_eda.settings.development` I think you should use the -dev file instead or overwrite the desired values. Once `DJANGO_SETTINGS_MODULE` or the image_url can be overwritten through envvars, you can still use the -stage file without change its default values.

AlanCoding:
> if you want to build the image from the source code and apply by default aap_eda.settings.development 

Oh good catch! That was completely unintentional. I meant to make it configurable but did not mean to change the default value. I pushed a commit, and want to ask if this form might be acceptable.

AlanCoding:
@chrismeyersfsu In https://github.com/ansible/eda-server/pull/715, I know why the `service-mesh` network was added (to talk with other components hooked in aap-dev) but what was the `default` network for? It would be nice to leave a comment on these to tell what it is for.

tools/docker/docker-compose-stage.yaml
```diff
@@ -84,6 +84,9 @@ services:
       start_period: 5s
   eda-api:
     image: ${EDA_IMAGE:-quay.io/ansible/eda-server}:${EDA_IMAGE_VERSION:-main}
+    build:
```
Alex-Izquierdo, line 87: @AlanCoding following up my [last comment](https://github.com/ansible/eda-server/pull/978#pullrequestreview-2191159938), one of the purpose of the -stage is to run the application with prebuilt images, if you want to build the image from the source code, you have the -dev image. This change would also break some CI pipelines. I think this file should not build the image and it is not necessary because you could override the variable to use your local one if you want that. 

What we can do, is to remove `EDA_IMAGE_VERSION` and use just `EDA_IMAGE` for the whole url as we already do in the -dev image.

AlanCoding: > What we can do, is to remove EDA_IMAGE_VERSION and use just EDA_IMAGE for the whole url as we already do in the -dev image.

I did find that type-inconsistent. I agree, if I'm expecting to be able to swap the dockerfile, then I need to be able to give the image in the same format. We could use a separate variable for the full image, like `EDA_IMAGE=$(EDA_IMAGE_BASE):$(EDA_IMAGE_VERSION)`. But I will just make the change like you suggest.

https://github.com/ansible/eda-server/blob/726a3a9ddff4c7d0336da04e7fd5e2eb96be7e80/tools/docker/docker-compose-dev.yaml#L114

This will lead to a little more legwork of building images with the dev dockerfile and then running them with -stage, but that is fine, given that it's intended.

Alex-Izquierdo: I think that replicating what we do in -dev file is enough.
`image: ${EDA_IMAGE:-quay.io/ansible/eda-server:main}`

AlanCoding: On second look, this is too big of a change for me to changing the meaning of `EDA_IMAGE` because of how it is referenced in the Makefile. There is an existing `EDA_IMAGE_TAG` which would indicate the image without the version tag, but the contract right now doesn't use that variable.

tools/docker/docker-compose-stage.yaml
```diff
@@ -14,8 +14,8 @@ x-environment:
   - EDA_CONTROLLER_TOKEN=${EDA_CONTROLLER_TOKEN:-some-secret-token}
   - EDA_CONTROLLER_SSL_VERIFY=${EDA_CONTROLLER_SSL_VERIFY:-no}
   - EDA_PROTOCOL=http
-  - EDA_HOST=eda-api:8000
-  - EDA_SERVER=http://eda-api:8000
+  - EDA_HOST=http://${EDA_HOST:-eda-api:8000}
```
AlanCoding, line 17: oh this seems wrong

AlanCoding:
Hopefully this patch is in good shape now. It will require light iteration, but should give me everything I need for aap-dev.

.gitignore
```diff
@@ -23,3 +23,6 @@ __pycache__/
 Taskfile.yml
 Taskfile.yaml
 /tools/deploy/environment.properties
+
+# Special file placed for gateway aap-dev testing
+src/aap_eda/settings/dev_hacked_settings.py
```
Alex-Izquierdo, line 28: If this file is only for development testing should be [ignored locally](https://git-scm.com/docs/gitignore) instead here?

Alex-Izquierdo (dismissed):

.gitignore
AlanCoding: That's fine, adding it to the aap-dev instructions.

Alex-Izquierdo (approved):

tools/docker/docker-compose-stage.yaml
```diff
@@ -14,8 +14,8 @@ x-environment:
   - EDA_CONTROLLER_TOKEN=${EDA_CONTROLLER_TOKEN:-some-secret-token}
   - EDA_CONTROLLER_SSL_VERIFY=${EDA_CONTROLLER_SSL_VERIFY:-no}
   - EDA_PROTOCOL=http
-  - EDA_HOST=eda-api:8000
-  - EDA_SERVER=http://eda-api:8000
+  - EDA_HOST=${EDA_HOST:-eda-api:8000}
+  - EDA_SERVER=http://${EDA_HOST:-eda-api:8000}
```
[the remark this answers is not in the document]
AlanCoding: I don't know, this is what the `tools/docker/docker-compose-dev.yaml` file has.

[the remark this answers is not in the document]
jshimkus-rh: At least it's consistent.

jshimkus-rh (approved):

PR #981, opened 2024-07-30
fix(AAP-28076): remove org id from credential type

Alex-Izquierdo:
We return an empty list of credential types for every user that doesn't belong to the default organization. 
As we agreed in internal discussions, the solution is to remove the organization relationship and the permissions should be CRUD for system_admin, read for everyone else. This is aligned with AWX. 

Jira: https://issues.redhat.com/browse/AAP-28076


Alex-Izquierdo:
E2E tests needs to be updated. It's late for me, I will do it tomorrow and link here the respective successful test run.

mkanoor:
@Alex-Izquierdo A customer can create their own Credential Types, and that could be per Organization, would these Custom Credentials be visible to every Org now with this change.

Alex-Izquierdo:
@mkanoor That's correct. Credential types are understood as system wide content, so each new credential type is visible to every org and only can be created by system admins.

Dostonbek1 (dismissed):
This overall looks good to me, ran it locally and permission class works as expected.

hsong-rh:
@Alex-Izquierdo This change only allows the admin to create credential types, do we need to change documentation to match this?

Alex-Izquierdo:
> @Alex-Izquierdo This change only allows the admin to create credential types, do we need to change documentation to match this?

@hsong-rh that's correct, I'm managing it internally.

hsong-rh (approved):

Alex-Izquierdo:
succesful E2E test run: https://github.com/ansible/eda-server/actions/runs/10204136436

Alex-Izquierdo:
Please if you can reapprove again @hsong-rh @Dostonbek1 
Just changed the migrations

hsong-rh (approved):

hsong-rh (approved):

Dostonbek1 (approved):

PR #982, opened 2024-07-31
feat(AAP-28111): add more managed roles on object and org levels

Dostonbek1:
This PR attempts to add more managed roles on both object and org levels. The following is the list of new roles with this change.

**Object-level roles _NOTE_: Use role consists `view` and `change` permissions:**

- Project Use - Has use permissions to a single project
- Activation Use - Has use permissions to a single activation
- Eda Credential Use - Has use permissions to a single eda credential
- Decision Environment Use - Has use permissions to a single decision environment
- Webhook Use - Has use permissions to a single webhook

**Org-level roles:**

- Organization Project Admin - Has all permissions to projects within an organization
- Organization Activation Admin - Has all permissions to activations within an organization
- Organization Eda Credential Admin - Has all permissions to eda credentials within an organization
- Organization Decision Environment Admin - Has all permissions to decision environments within an organization
- Organization Webhook Admin - Has all permissions to webhooks within an organization

This PR also renames org-level roles to have "Organization" prefix, e.g. "Admin" -> "Organization Admin".

JIRA: [AAP-28111](https://issues.redhat.com/browse/AAP-28111) & [AAP-27990](https://issues.redhat.com/browse/AAP-27990) & [AAP-27848](https://issues.redhat.com/browse/AAP-27848)


Dostonbek1:
One outstanding question I have on this is whether we can rename the existing org-level roles to have "Organization " prefix for more clarity, e.g. Organization Admin, Organization Viewer, without breaking backward compatibility. Thoughts? @AlanCoding @Alex-Izquierdo

AlanCoding:
> One outstanding question I have on this is whether we can rename the existing org-level roles to have "Organization " prefix for more clarity, e.g. Organization Admin, Organization Viewer, without breaking backward compatibility. Thoughts?

Oh do we still have "Admin" as a role name?

https://github.com/ansible/eda-server/blob/main/src/aap_eda/core/management/commands/create_initial_data.py#L35

I thought we require consistent names to play nicely with the resource provider. So yeah strong agreement that it should be "Organization Admin" type naming.

Alex-Izquierdo (dismissed):
LGTM

Alex-Izquierdo:
on-hold for now. Removing my approval for security.

AlanCoding (approved):
My only concern is if migration from prior version (older version, using the old role model) will then see duplicates created because the migration logic looks for same-name roles to copy over to DAB RBAC. This could be fixed by having that data migration rename "Admin" role if it finds it, for example. That may be important, but isn't urgent to address right now.

Dostonbek1:
@AlanCoding I agree. I thought about changing the query to look for 2 name patterns, e.g. "Admin" and "Organization Admin", and it finds either then we call `.update_or_create` instead of `.get_or_create`. I think that should handle this case pretty well.

Alex-Izquierdo:
LGTM but have you test it with a existing instance? That is not covered by our tests and I'm concerned about potential migration issues.

Dostonbek1:
@Alex-Izquierdo Yes, I have tested the org level role definition migrations and confirmed that the existing org level roles get migrated with new names with "Organization" prefix and do not get duplicated.

Alex-Izquierdo (dismissed):
LGTM

hsong-rh (approved):
LGTM

Alex-Izquierdo (approved):

PR #984, opened 2024-08-01
feat: (AAP-28247) - add endpoint to fetch Activations by event streams

hsong-rh:
https://issues.redhat.com/browse/AAP-28247

Add an endpoint `webhooks/:id/activations` to display activations by the event stream.


tests/integration/api/test_activation_with_webhook.py
```diff
@@ -246,6 +246,30 @@ def test_create_activation_with_webhooks(
...
+def test_list_activations_by_webhook(
+    admin_client: APIClient,
+    default_activation: models.Activation,
+    new_activation: models.Activation,
+    default_webhook: models.Webhook,
+):
+    activation_1 = default_activation
+    activation_2 = new_activation
+
+    activation_1.webhooks.add(default_webhook)
```
mkanoor, line 259: @hsong-rh Can we add a test for no activations connected to EventStreams

hsong-rh: Added

mkanoor (dismissed):

src/aap_eda/core/migrations/0042_alter_activation_webhooks.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 4.2.7 on 2024-08-01 18:59
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("core", "0041_alter_credentialtype_options_and_more"),
```
Dostonbek1, line 8: there is a migration conflict from the previous merge, can you fix it?

hsong-rh: Fixed

src/aap_eda/api/views/webhook.py
```diff
@@ -203,3 +208,36 @@ def partial_update(self, request, *args, **kwargs):
...
+        ],
+    )
+    @action(
+        detail=False,
+        queryset=models.Activation.objects.order_by("id"),
+        filterset_class=filters.ActivationFilter,
+        url_path="(?P<id>[^/.]+)/activations",
+    )
+    def activations(self, request, id):
+        webhook = get_object_or_404(models.Webhook, id=id)
```
Dostonbek1, line 236: `get_object_or_404` function bypasses permission check by default, I think instead we should use `self.get_object()` or change the queryset to `models.Webhook.access_qs(request.user)` or use an approach similar to [this](https://github.com/ansible/eda-server/blob/1ccc5e415696ef5b1072e3cb9265dcf375d00c5b/src/aap_eda/api/views/activation.py#L182C1-L189C14).

Dostonbek1 (changes requested):

src/aap_eda/api/views/webhook.py
hsong-rh: Fixed

Dostonbek1 (approved):

Alex-Izquierdo (approved):

PR #986, opened 2024-08-02
feat: (AAP-28246) - only delete the event stream when no activations using it

hsong-rh:
https://issues.redhat.com/browse/AAP-28246

A conflict exception(409) will be raised when activations use the event stream. UI shows:
image


hsong-rh:
Some changed files are duplicates with https://github.com/ansible/eda-server/pull/984. They will disappear when it is merged.

mkanoor (approved):

Dostonbek1 (approved):
lgtm. Just one minor comment, would it make sense to provide force delete option here similar to how we do for credentials or DE?

mkanoor:
@Dostonbek1 Since this is wired to Activations I don't think we would be able to force it since those Activations depend on it for incoming data. In the call we had talked about cascading delete that can be implemented in the future

PR #989, opened 2024-08-05
AAP-27921 Use DAB-provided Redis client.

jshimkus-rh:
This change uses the DAB-provided Redis client support which encapsulates requisite HA capabilities.

It does not currently take specific advantage of these capabilities; that is to come later.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,7 @@ django-ansible-base = { git = "https://github.com/ansible/django-ansible-base.gi
     "channel_auth",
     "rbac",
 ] }
+django-redis = "5.4.*"
 jinja2 = ">=3.1.3,<3.2"
 django-split-settings = "^1.2.0"
 pexpect = "^4.9.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,8 +45,8 @@ rq-scheduler = "^0.10"
 django-ansible-base = { git = "https://github.com/ansible/django-ansible-base.git", branch = "devel", extras = [
     "channel_auth",
     "rbac",
+    "redis_client",
 ] }
-django-redis = "5.4.*"
 jinja2 = ">=3.1.3,<3.2"
 django-split-settings = "^1.2.0"
 pexpect = "^4.9.0"
--- a/src/aap_eda/core/tasking/__init__.py
+++ b/src/aap_eda/core/tasking/__init__.py
@@ -183,6 +183,17 @@ class Job(_Job):
         super().__init__(id, connection, serializer)
 
 
+# django-rq's rqworker command does not support --connection-class so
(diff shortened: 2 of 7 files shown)


pyproject.toml
```diff
@@ -46,6 +46,7 @@ django-ansible-base = { git = "https://github.com/ansible/django-ansible-base.gi
     "channel_auth",
     "rbac",
 ] }
+django-redis = "5.4.*"
```
john-westcott-iv, line 49: Instead of pulling this in directly, I'd modify the ansible-base to include the extra `redis_client` and that should pull in django-redis and anything required in the future.

src/aap_eda/core/tasking/__init__.py
```diff
@@ -162,6 +184,19 @@ def __init__(
         super().__init__(id, connection, serializer)
 
 
+# django-rq's rqworker command does not support --connection-class so
+# we cannot specify the DAB redis client that way.  Even if it did we
+# couldn't use it as DAB requires a url parameter that Redis does not.
+# If the connection a worker is given is not from DAB we replace it
+# with one that is.
+def _get_necessary_client_connection(connection: Connection) -> Connection:
+    if type(connection) not in [DABRedis, DABRedisCluster]:
```
Alex-Izquierdo, line 193: Is there a reason to not use `isinstance` instead of `type`?

tests/integration/core/test_tasking.py
```diff
@@ -54,3 +63,38 @@ def test_unique_enqueue_new_job(default_queue, eda_caplog):
...
     assert job.kwargs["number"] == 2
     assert "Enqueing unique job" in eda_caplog.text
+
+
+def test_worker_dab_client(default_queue: Queue):
+    """Test that workers end up with a DABRedis client connection."""
+
+    # Verify if given no connection the worker gets a DABRedis.
+    worker = DefaultWorker([default_queue])
+    assert type(worker.connection) is DABRedis
```
Alex-Izquierdo, line 73: Same question about usage of `type` instead `isinstance`

Alex-Izquierdo (approved):

src/aap_eda/core/tasking/__init__.py
jshimkus-rh: @Alex-Izquierdo No.  It would be my first choice but then I saw there is some EDA code that uses type().

tests/integration/core/test_tasking.py
jshimkus-rh: @Alex-Izquierdo No.  It would be my first choice but then I saw there is some EDA code that uses type().

Alex-Izquierdo (approved):
LGTM

PR #995, opened 2024-08-08
feat: rsyslog audit logging (WIP)

zkayyali812:
Going to take a step back on this, and see if this can be reworked simpler to no longer require rsyslog.

Adding rsyslog audit logging for EDA features.

This PR is still a work in progress.
Please feel free to review/add comments though in the meantime :) 

Still need to ensure this works when containerized

Jira: 28245


tools/docker/Dockerfile
```diff
@@ -20,12 +20,15 @@ RUN DNF=dnf \
...
-    java-17-openjdk-headless" \
+    java-17-openjdk-headless \
+    supervisor" \
+    && $DNF -y install  https://dl.fedoraproject.org/pub/epel/epel-release-latest-9.noarch.rpm \
     && $DNF -y install dnf-plugins-core \
     && $DNF -y config-manager --set-enabled crb \
     && $DNF -y install $INSTALL_PACKAGES \
     && $DNF -y clean all \
     && rm -rf /var/cache/dnf
+COPY ./tools/docker/supervisord/supervisord.conf /etc/supervisord.conf
```
Alex-Izquierdo, line 31: I'm concerned about the potential approach that this might be taking if are going to run supervisor in the container image. 
Running multiple processes in the same container is not considered a good practice and in EDA we have tried to avoid it from the beginning.

src/aap_eda/core/management/commands/run_rsyslog_configurer.py
```diff
@@ -0,0 +1,25 @@
+import logging
+
+from django.core.management.base import BaseCommand, CommandError
+
+from aap_eda.utils.external_logging import reconfigure_rsyslog
+
+logger = logging.getLogger("aap.gateway.rsyslog_configurer")
```
Alex-Izquierdo, line 7: I guess that since this PR is still WIP you already have in mind to change it, just in case I remind we should not point to gateway stuff in any case.

zkayyali812: Great catch :) will fix

tools/docker/Dockerfile
zkayyali812: In gateway and in EDA, supervisor is used to run rsyslog.

Trying to follow their approach here to lift and shift these requirements.

Do you have alternative suggestions about a different approach?

zkayyali812: After further thought and discussion, I will refactor this to rework this to a simpler approach

zkayyali812:
Closing PR

PR #997, opened 2024-08-09
feat (AAP-29049) - before deleting credentials check if event streams attached

hsong-rh:
https://issues.redhat.com/browse/AAP-29049

The UI will raise a warning as shown in the screenshot. Because we support the `force` flag, the deletion will succeed after checking `Yes`.

image


mkanoor (approved):

Dostonbek1:
@hsong-rh Does this account for resources that the user has no access to?

hsong-rh:
> @hsong-rh Does this account for resources that the user has no access to?

We use `objects.filter` to retrieve referenced resources, which should check permissions. So the answer should be no.

Dostonbek1 (approved):

PR #1001, opened 2024-08-12
fix: check for valid settings before creating EventStreams

mkanoor:
JIRA: https://issues.redhat.com/browse/AAP-29132

Before we create event streams we have to check the settings are proper. 
If the EDA_WEBHOOK_BASE_URL is missing we should not allow users to create event streams that use non mTLS webhooks.

If the EDA_WEBHOOK_MTLS_BASE_URL is missing we should not allow users to create event streams that use mTLS webhook.

Screenshot 2024-08-12 at 12 29 13 PM
Screenshot 2024-08-12 at 12 28 47 PM


tests/integration/dab_rbac/test_crud_permissions.py
```diff
@@ -109,8 +110,9 @@ def test_add_permissions(
                     related_perm = "view"
                 give_obj_perm(default_user, related_obj, related_perm)
 
-    response = user_client.post(url, data=post_data, format="json")
-    assert response.status_code == 201, response.data
+    with override_settings(WEBHOOK_BASE_URL="https://www.example.com/"):
```
Dostonbek1, line 113: @mkanoor Would it better if we set this as a decorator for the test function? Not sure how much difference it makes, but since this test is parameterized, this override gets run for every parameter.

Dostonbek1 (approved):

msmagnanijr (approved):

PR #1012, opened 2024-08-16
Redis ha/aap 28152

jshimkus-rh:
Add redis availability checking to Activation CRUD.

https://issues.redhat.com/browse/AAP-28152

### DIFF
--- a/src/aap_eda/api/views/activation.py
+++ b/src/aap_eda/api/views/activation.py
@@ -40,6 +40,8 @@ from aap_eda.tasks.orchestrator import (
     stop_rulebook_process,
 )
 
+from .mixins import RedisDependencyMixin
+
 logger = logging.getLogger(__name__)
 
 resource_name = "RulebookActivation"
@@ -48,6 +50,7 @@ resource_name = "RulebookActivation"
 class ActivationViewSet(
     mixins.DestroyModelMixin,
     viewsets.GenericViewSet,
+    RedisDependencyMixin,
 ):
     queryset = models.Activation.objects.all()
     serializer_class = serializers.ActivationSerializer
@@ -70,7 +73,8 @@ class ActivationViewSet(
             status.HTTP_400_BAD_REQUEST: OpenApiResponse(
                 description="Invalid data to create activation."
             ),
-        },
+        }
+        | RedisDependencyMixin.redis_unavailable_response(),
     )
     def create(self, request):
         context = {"request": request}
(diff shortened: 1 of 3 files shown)


src/aap_eda/api/views/activation.py
```diff
@@ -79,6 +83,14 @@ def create(self, request):
         )
         serializer.is_valid(raise_exception=True)
 
+        # If we're expected to run this activation we need redis
```
hsong-rh, line 86: Should we use the `is_enabled` field validator to check this? something like following in the serializer file?:
```
is_enabled = serializers.BooleanField(
    validators=[validators.check_if_redis_is_available],
)
```

jshimkus-rh: @hsong-rh I view a distinction between receiving a valid value and not being able to act on that value because Redis isn't available.

We could get the same functional result using a validator but, as described, that would require embedding system state checks into what is ostensibly a field value check.  If we were to opt to do it via the serializer I think a better choice would be to incorporate it as part of the validate() as that already takes a more holistic approach.

That having been said there are other usages (e.g., enable) where the decision needs to be made by the view and there's a strong development/maintenance argument to be made that the determination should be performed at a consistent level (in this case, views) in the processing.

hsong-rh: @jshimkus-rh Maybe `validate()` is a better choice. The logic in the view should be as simple as possible; all validation should happen in the serializer. That's my understanding.

hsong-rh: 👌

src/aap_eda/api/views/activation.py
```diff
@@ -79,6 +83,14 @@ def create(self, request):
...
+        # to be available.
+        if serializer.validated_data.get(
+            "is_enabled",
+            models.activation.DEFAULT_ENABLED,
```
Dostonbek1, line 90: is this line correct? Should it be `models.activation.is_enabled`?

jshimkus-rh: @Dostonbek1 This is accessing the activation.py package for the default value as there's no Activation instance at this point and the request may not actually contain is_enabled.
Only after calling serializer.create() would we be guaranteed of having is_enabled (on the created activation) but I wanted to avoid doing anything with the DB (even though it's wrapped in a transaction).

I considered this an area where I might get "just do it in the transaction" so I'm prepared to change if people want. :-)

Dostonbek1 (approved):

hsong-rh (approved):

PR #1016, opened 2024-08-20
fix: (AAP-29630) - save extra_var when creating an activation with vault/AAP credentials

hsong-rh:
https://issues.redhat.com/browse/AAP-29630


jshimkus-rh (approved):
The custom_ credential_type in tests/integration/api/test_activation_with_credential.py is only used by one test. Maybe it doesn't need to be a fixture?

Either way is good.

Alex-Izquierdo (approved):

matburt (approved):

PR #1017, opened 2024-08-20
chore: bump django-ansible-base package

Dostonbek1:


jshimkus-rh (approved):

ttuffin (dismissed):

Alex-Izquierdo:
@Dostonbek1 I'm not familiar yet with the new pipeline to run e2e-ui, those failures are expected?

Dostonbek1:
@Alex-Izquierdo the failures are not related to the API, there was a merge on the UI that introduced those failures; they are working on reverting it now. https://github.com/ansible/ansible-ui/pull/3023

Alex-Izquierdo (dismissed):

zkayyali812 (approved):

Dostonbek1:
e2d test run with an updated test suite: https://github.com/ansible/eda-server/actions/runs/10563148958

AlanCoding:
Look at the e2e test failures here, I would narrow in on:

https://github.com/ansible/eda-server/blob/8e06d7565be72103fef6e9d7dd3187538519b2a4/.github/workflows/scheduled.yml#L30-L38

The tests are getting 400 errors because the policies related to local resource management. Follow the config for the "stage" target...

You can set settings like:

https://github.com/ansible/eda-server/blob/8e06d7565be72103fef6e9d7dd3187538519b2a4/tools/docker/docker-compose-stage.yaml#L30-L32

The settings we want to change are:
 - `ALLOW_LOCAL_ASSIGNING_JWT_ROLES` to True
 - `ALLOW_SHARED_RESOURCE_CUSTOM_ROLES` to True

And I think those would look like `EDA__ALLOW_LOCAL_ASSIGNING_JWT_ROLES`.

Then you may want to apply those to all of your standalone environments. This may require a follow-up change to the aap-dev environment to set the back to false, but no one will even notice that's broken.

jshimkus-rh (dismissed):

AlanCoding:
Looks like the 1 e2e failure is unrelated, and the failures we hoped to fix are addressed by the rebase :+1:

zkayyali812 (approved):

hsong-rh (approved):

PR #1023, opened 2024-08-27
chore(AAP-29868): add new vars related to roles

Dostonbek1:
This PR adds two new variables in settings which default to `False` in settings file as it is the expected behavior for the product but `True` in docker-compose files for dev environment and test suite. These vars are required for bumping django-ansible-base here https://github.com/ansible/eda-server/pull/1017, in order to not break the e2e tests. 

[AAP-29868](https://issues.redhat.com/browse/AAP-29868)


AlanCoding (approved):
I understand this may need to be combined with the DAB update, but it can probably be merged on its own with no ill-effects.

hsong-rh (approved):

zkayyali812 (approved):

PR #1029, opened 2024-08-27
chore: Create and POST jUnit results file

zkayyali812:
https://issues.redhat.com/browse/AAP-28571

e2e test run - https://github.com/ansible/eda-server/actions/runs/10597875343

This PR adds the CI onboarding of our unit tests. This PR should not merge until it is confirmed to be functioning with our organization github secrets.


mkanoor (approved):

zkayyali812:
Hey @Alex-Izquierdo ,

My hope is that this would also work with forks, but I share your worry here.

I see in other places we reference in our github actions, and they are working in our forks - https://github.com/ansible/eda-server/blob/main/.github/workflows/e2e-dispatch.yml#L69

My expectation and hope is that this would function similarly.

Alex-Izquierdo:
> Hey @Alex-Izquierdo ,
> 
> My hope is that this would also work with forks, but I share your worry here.
> 
> I see in other places we reference secrets in our github actions, and they are working in our forks - https://github.com/ansible/eda-server/blob/main/.github/workflows/e2e-dispatch.yml#L69
> 
> My expectation and hope is that this would function similarly.

e2e-dispatch doesn't work in fork unless you define the secrets. It follows a different flow and it can be only executed on-demand through a label.

jneedle (approved):

Alex-Izquierdo (approved):

PR #1033, opened 2024-08-29
Add setting to force TLS on for Redis connections

dhageman:
This will add a new setting "MQ_TLS" to force TLS on for Redis connections. 

The default value for this setting is 'False'.  Switching it to 'True' will force TLS to be used for Redis connections.

The current logic of enabling TLS when provided a client certificate remains.

### DIFF
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -63,7 +63,7 @@ Redis queue settings:
 *   - Takes precedence over host and port
 * MQ_HOST - Redis queue hostname (default: "127.0.0.1")
 * MQ_PORT - Redis queue port (default: 6379)
-* MQ_TLS - Force TLS on when True (default: False)
+* MQ_TLS - Force TLS on when True (default: None)
 * MQ_DB - Redis queue database (default: 0)
 * MQ_USER - Redis user (default: None)
 * MQ_USER_PASSWORD - Redis user passed (default: None)
@@ -366,7 +366,7 @@ REDIS_USER_PASSWORD = settings.get("MQ_USER_PASSWORD", None)
 REDIS_CLIENT_CACERT_PATH = settings.get("MQ_CLIENT_CACERT_PATH", None)
 REDIS_CLIENT_CERT_PATH = settings.get("MQ_CLIENT_CERT_PATH", None)
 REDIS_CLIENT_KEY_PATH = settings.get("MQ_CLIENT_KEY_PATH", None)
-REDIS_TLS = settings.get("MQ_TLS", False)
+REDIS_TLS = settings.get("MQ_TLS", None)
 DEFAULT_REDIS_DB = 0
 REDIS_DB = settings.get("MQ_DB", DEFAULT_REDIS_DB)
 RQ_REDIS_PREFIX = settings.get("RQ_REDIS_PREFIX", "eda-rq")
@@ -393,8 +393,15 @@ def _rq_common_parameters():
             "HOST": REDIS_HOST,
             "PORT": REDIS_PORT,
         }
-        if REDIS_CLIENT_CERT_PATH or REDIS_TLS:
+        if REDIS_TLS:
             params["SSL"] = True
+        else:
+            # This check should be deprecated in favor of just using
+            # MQ_TLS as the determining factor for if TLS is enabled.
+            if REDIS_CLIENT_CERT_PATH and REDIS_TLS is None:
+                params["SSL"] = True
+            else:
+                params["SSL"] = False
     return params
 
 
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -396,7 +396,7 @@ def _rq_common_parameters():
         if REDIS_TLS:
             params["SSL"] = True
         else:
-            # TODO: Deprecate implicit setting based on cert path in favor of 
+            # TODO: Deprecate implicit setting based on cert path in favor of
             #       MQ_TLS as the determinant.
             if REDIS_CLIENT_CERT_PATH and REDIS_TLS is None:
                 params["SSL"] = True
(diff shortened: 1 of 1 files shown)


src/aap_eda/settings/default.py
```diff
@@ -365,6 +366,7 @@ def _get_databases_settings() -> dict:
 REDIS_CLIENT_CACERT_PATH = settings.get("MQ_CLIENT_CACERT_PATH", None)
 REDIS_CLIENT_CERT_PATH = settings.get("MQ_CLIENT_CERT_PATH", None)
 REDIS_CLIENT_KEY_PATH = settings.get("MQ_CLIENT_KEY_PATH", None)
+REDIS_TLS = settings.get("MQ_TLS", False)
```
jshimkus-rh, line 369: Should default to None for backwards compatibility and respecting the provided, if any, value.

src/aap_eda/settings/default.py
```diff
@@ -391,7 +393,7 @@ def _rq_common_parameters():
             "HOST": REDIS_HOST,
             "PORT": REDIS_PORT,
         }
-        if REDIS_CLIENT_CERT_PATH:
+        if REDIS_CLIENT_CERT_PATH or REDIS_TLS:
```
jshimkus-rh, line 396:
```suggestion
        if (REDIS_CLIENT_CERT_PATH and (REDIS_TLS is None)) or REDIS_TLS:
```
This provides backwards compatibility where TLS is not specified and  respects the setting of TLS, if specified.  That is, if TLS is specified False it supersedes the previous implicit setting.

mkanoor: @jshimkus-rh Since this is a new field and its in the same file settings/default.py why does it have to be None, are you thinking about the settings.yml file that the customer might have instead of the python file. Even in that case it would be inheriting the default values defined in the settings.py first before it overrides values from the settings.yml.

src/aap_eda/settings/default.py
dhageman: I will comment in the thread below to keep the replies concise.

src/aap_eda/settings/default.py
jshimkus-rh: @mkanoor As currently coded if REDIS_TLS is actually specified (via settings/env vars) as False the code would ignore it and enable TLS regardless if REDIS_CLIENT_CERT_PATH were specified.
If it is set we must assume it was purposefully done so and respect the value.

dhageman: The logic behind setting REDIS_TLS to False by default is to align with the current behavior.  That behavior is to not use TLS unless we activate it by setting REDIS_CLIENT_CERT_PATH.  Therefore, REDIS_TLS should be set to False by default.  The only reason to deviate from that is if there is another way to activate TLS for EDA.  I am not the most familiar with the code base but that didn't seem like a possibility when I reviewed the code.

jshimkus-rh: @dhageman We should respect (backwards compatibility) the implicit setting of TLS = True if TLS is not separately specified.  
With this change we now have to ability to have TLS enablement specifically chosen by the user and we must assume that, for whatever purpose they may have, if it is provided and False that is what the user wants.

dhageman: @jshimkus-rh - Okay, the so the condition you want to address if someone explicitly sets the REDIS_CLIENT_CERT_PATH to <something> and REDIS_TLS to False that you want to NOT enable TLS?

jshimkus-rh: @dhageman 
> Okay, the so the condition you want to address if someone explicitly sets the REDIS_CLIENT_CERT_PATH to "something" and REDIS_TLS to False that you want to NOT enable TLS?

That is correct.

dhageman: I really think this highlights a bigger issue.  There should only be one option to enable TLS so it is explicitly clear what is occurring.  The use of enabling TLS via REDIS_CLIENT_CERT_PATH should be deprecated and eventually removed.  I wonder if the better approach is to document the behavior today and then move toward making REDIS_TLS the only option to enable TLS going forward.  Thoughts?

jshimkus-rh: @dhageman That is an option.  I didn't propose it in order to avoid the delay it would  add to getting this change in asap.  As long as it doesn't _require_ any changes to existing installers, etc. we can get this in right away and pursue anything further later.

dhageman: I have updated the pull request and put in a comment on there on the ideal path forward.  We should get this in the next round of deprecation notices.  This will work in the interim.

src/aap_eda/settings/default.py
```diff
@@ -391,8 +393,15 @@ def _rq_common_parameters():
             "HOST": REDIS_HOST,
             "PORT": REDIS_PORT,
         }
-        if REDIS_CLIENT_CERT_PATH:
+        if REDIS_TLS:
```
jshimkus-rh, line 396: I know I didn't mention this yesterday, and I'm not asking for it to be changed, but this could all be collapsed to:
```suggestion
        params["SSL"] = REDIS_TLS or (REDIS_CLIENT_CERT_PATH and REDIS_TLS is None)
```

src/aap_eda/settings/default.py
```diff
@@ -391,8 +393,15 @@ def _rq_common_parameters():
...
             params["SSL"] = True
+        else:
+            # This check should be deprecated in favor of just using
```
jshimkus-rh, line 399: I'd like this marked as TODO.
```suggestion
            # TODO: Deprecate implicit setting based on cert path in favor of MQ_TLS as the determinant.
```

src/aap_eda/settings/default.py
dhageman: Right!  My goal was to keep the logic "exploded" to make it super obvious what needs to be done.

jshimkus-rh (approved):

jshimkus-rh (approved):

matburt (approved):

rooftopcellist (approved):

PR #1040, opened 2024-09-04
Allow reverse-sync if resource server is defined

AlanCoding:
AAP-30009

Specifically with this commit:

https://github.com/ansible/django-ansible-base/commit/2af5f34dd4887eeba65f48e01e488745e90f8f04#diff-8623b469b6e5d0b1173a741c5702169d5a06e643a8d416c860127373d662217d

This should now be safe to add, because any standalone variation will not have the URL of the resource server defined, which means that it still will not sync.

### DIFF
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -665,6 +665,11 @@ ALLOW_LOCAL_RESOURCE_MANAGEMENT = settings.get(
     "ALLOW_LOCAL_RESOURCE_MANAGEMENT", True
 )
 
+# If we have a resource server defined, apply local changes to that server
+RESOURCE_SERVER_SYNC_ENABLED = settings.get(
+    "RESOURCE_SERVER_SYNC_ENABLED", True
+)
+
 # ---------------------------------------------------------
 # DJANGO ANSIBLE BASE RESOURCES REGISTRY SETTINGS
 # ---------------------------------------------------------


AlanCoding:
DAB is on a devel pin here:

https://github.com/ansible/eda-server/blob/cb986825ce2f974660f1af67afd98e1957254b92/pyproject.toml#L44

h-kataria:
@elyezer please review

h-kataria:
@jneedle can we get a review on this

Dostonbek1:
@AlanCoding should we include dab bump with this PR for it to work with your changes in dab? I see that eda-server is using [this tag ](https://github.com/ansible/eda-server/blob/c0d8d524bdf6e5d0f7a2512cc520b96bc7e8476f/poetry.lock#L624)which was 3 days ago.

dhaustein:
@AlanCoding  What was the reason for NOT adding `SOCIAL_AUTH_PIPELINE`? There is no need for that in eda-server?

elyezer (commented):
These changes should not make a difference until the related DAB changes are used here. Seems safe to merge unless the bump on DAB gets added here.

relrod:
> Seems safe to merge unless the bump on DAB gets added here.

Why would updating DAB break this?

AlanCoding:
We have a recent conversation in Slack (@relrod and I) where the direction we were going is to close this, and only do the auth changes for right now. This is to help de-risk event 1, which does not need this. That will obligate us to resurrect this PR for event 2. Immediately, the side-track event 2 pipelines can set this on the installer side.

PR #1043, opened 2024-09-05
fix(AAP-30273): allow updating is_superuser field when local resource management is disallowed

Dostonbek1:
JIRA: [AAP-30273](https://issues.redhat.com/browse/AAP-30273)

### DIFF
--- a/src/aap_eda/api/serializers/mixins.py
+++ b/src/aap_eda/api/serializers/mixins.py
@@ -22,7 +22,7 @@ class SharedResourceSerializerMixin:
     def validate_shared_resource(self, data=None):
         if not settings.ALLOW_LOCAL_RESOURCE_MANAGEMENT:
             view = self.context.get("view")
-            action = view.action.capitalize() if view else "action"
+            action = view.action.capitalize() if view else "Action"
 
             # exception where we should allow updating is_superuser field
             if action == "Partial_update" and isinstance(
--- a/src/aap_eda/api/serializers/mixins.py
+++ b/src/aap_eda/api/serializers/mixins.py
@@ -15,13 +15,22 @@
 from django.conf import settings
 
 from aap_eda.api import exceptions as api_exc
+from aap_eda.core import models
 
 
 class SharedResourceSerializerMixin:
-    def validate_shared_resource(self):
+    def validate_shared_resource(self, data=None):
         if not settings.ALLOW_LOCAL_RESOURCE_MANAGEMENT:
             view = self.context.get("view")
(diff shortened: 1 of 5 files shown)


src/aap_eda/api/serializers/mixins.py
```diff
@@ -15,13 +15,22 @@
...
+from aap_eda.core import models
 
 
 class SharedResourceSerializerMixin:
-    def validate_shared_resource(self):
+    def validate_shared_resource(self, data=None):
         if not settings.ALLOW_LOCAL_RESOURCE_MANAGEMENT:
             view = self.context.get("view")
-            action = view.action.capitalize() if view else "Action"
+            action = view.action.capitalize() if view else "action"
```
Alex-Izquierdo: Should it be "Action"? (capitalized)

AlanCoding (changes requested):
By inheriting from `SharedResourceSerializerMixin`, as opposed to its child class `UserUpdateSerializerBase`, it looks like you've missed out on `validate_is_superuser` which is necessary for permission checking.

The test `test_modify_superuser_as_org_admin` has nice-looking coverage for exactly this, but since the permission check is performed in the serializer, swapping out the serializer can leave this edit unprotected from access rules. So it would probably be a good idea to test that it still gives a 403 with a non-admin client fixture.

h-kataria:
@elyezer please review

src/aap_eda/api/views/user.py
```diff
@@ -196,9 +196,10 @@ def perform_create(self, serializer):
         },
     ),
     partial_update=extend_schema(
-        exclude=not settings.ALLOW_LOCAL_RESOURCE_MANAGEMENT,
         description="Partial update of a user.",
-        request=serializers.UserCreateUpdateSerializer,
+        request=serializers.UserCreateUpdateSerializer
```
Alex-Izquierdo, line 200: I would prefer a conventional if-else block for readability.

Alex-Izquierdo (approved):

src/aap_eda/api/views/user.py
Dostonbek1: already changed. Thanks for catching it.

AlanCoding (approved):
:ship:

elyezer (commented):
From an initial look this doesn't seem that it may cause installation problems. If this does not change behavior when the `ALLOW_LOCAL_RESOURCE_MANAGEMENT` is set to `False` then we should be good to merge.

Alex-Izquierdo (approved):

PR #1046, opened 2024-09-06
fix(AAP-30640): update use roles to mitigate permission escalation

Dostonbek1:
This PR updates managed Use roles (e.g. Decision Environment Use) to limit their capabilities and avoid granting unwanted permissions. 

JIRA: [AAP-30640](https://issues.redhat.com/browse/AAP-30640) & [AAP-30627](https://issues.redhat.com/browse/AAP-30627)


AlanCoding (approved):
I think I see the logic in this. It's still weird that it's read-only, but that's the decision for related checks.

bzwei (approved):

mkanoor (approved):

hsong-rh (approved):

PR #1047, opened 2024-09-09
fix: [AAP-30505] add pubsub and suspended prefixes

jshimkus-rh:
Re-ordered settings alphabetically by related type.


bzwei:
@jshimkus-rh LGTM. The AAP issue is current not a blocker, so the commit target should be to devel branch.

hsong-rh (approved):

PR #1048, opened 2024-09-10
feat: AAP-30754 - Add analytics collector to EDA

hsong-rh:
https://issues.redhat.com/browse/AAP-30754

depends: https://github.com/ansible/eda-server/pull/1049


src/aap_eda/analytics/analytics_collectors.py
```diff
@@ -0,0 +1,422 @@
...
+
+
+@register(
+    "audit_rule_table",
+    "1.0",
+    format="csv",
+    description="Data on audit_rules",
+)
+def audit_rules_table(since, full_path, until, **kwargs):
+    activation_instance_ids = models.RulebookProcess.objects.filter(
```
Alex-Izquierdo, line 102: I think this query can return duplicated ids. Should we use `distinct`?

pyproject.toml
```diff
@@ -56,6 +56,7 @@ psycopg = "^3.1.17"
 xxhash = "*"
 pyjwt = { version = "*", extras = ["crypto"] }
 ecdsa = "*"
+insights-analytics-collector = {git = "https://github.com/RedHatInsights/insights-analytics-collector", rev = "main"}
```
Alex-Izquierdo, line 59: should we use a pinned version?

Alex-Izquierdo (commented):
@hsong-rh  What if we used the ORM to generate all the queries? This approach would improve maintainability and help prevent potential issues when changes are made to the model later on. Just make sure that the datetime values are properly formatted.

Alex-Izquierdo:
Even I would suggest to explore https://palewi.re/docs/django-postgres-copy/ which adds COPY support for django.

src/aap_eda/analytics/analytics_collectors.py
hsong-rh: The ids are the primary keys; they shouldn't have duplicates

pyproject.toml
hsong-rh: I found it has been published in Pypi

hsong-rh:
> Even I would suggest to explore https://palewi.re/docs/django-postgres-copy/ which adds COPY support for django.

Good suggestions, will work on it in the future.

src/aap_eda/analytics/analytics_collectors.py
Alex-Izquierdo: My point is that you are using OR in the query, so it can return the same id twice if it matches both conditions.

src/aap_eda/analytics/analytics_collectors.py
```diff
@@ -0,0 +1,438 @@
...
+
+@register(
+    "activations_table",
+    "1.0",
+    format="csv",
+    description="Data on activations",
+)
+def activations_table(since, full_path, until, **kwargs):
+    activation_query = """COPY (SELECT core_activation.id,
+       core_activation.name,
```
Alex-Izquierdo, line 50: What about my previous comment? https://github.com/ansible/eda-server/pull/1048#pullrequestreview-2296837617

I really think we should try to avoid raw queries if we can, the maintenance will be easier and much less error prone.

src/aap_eda/analytics/analytics_collectors.py
hsong-rh: added

src/aap_eda/analytics/analytics_collectors.py
hsong-rh: Changed to use Django ORM generated query

poetry.lock
```diff
@@ -1,9 +1,10 @@
-# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
+# This file is automatically @generated by Poetry and should not be changed by hand.
```
Alex-Izquierdo, line 1: please upgrade poetry, is very old.

hsong-rh: updated

hsong-rh:
moved to https://github.com/ansible/eda-server/pull/1065

PR #1049, opened 2024-09-11
feat: introduce application_settings

bzwei:
Application settings are predefined and registered in create_inital_data.py. They can be set or get as application_settings.XYZ The settings data are stored in core_setting DB table

AAP-30766: Create table schema for EDA settings


src/aap_eda/core/management/commands/create_initial_data.py
```diff
@@ -1054,6 +1056,37 @@
     },
 ]
 
+APPLICATION_SETTINGS_ARGS = [
```
bzwei, line 1059: @hsong-rh We can add more settings that are needed for analytics. Let me know the name and type

hsong-rh: @bzwei we need 3 more settings:

INSIGHTS_CERT_PATH, string
AUTOMATION_ANALYTICS_LAST_GATHER: string (timestamp)
AUTOMATION_ANALYTICS_LAST_ENTRIES: dict

Alex-Izquierdo (commented):
Maybe I don't understand the intent of the PR, I feel it confusing. In addition to my other questions regarding typing and structs,  why not merge SettingsRegistry into ApplicationSetttings to be a single proxy interface for the model that could be used as well at the initial data?

src/aap_eda/core/models/setting.py
```diff
@@ -0,0 +1,47 @@
...
+__all__ = ("Setting",)
+
+ENCRYPTED_STRING = "$encrypted$"
+
+
+class Setting(models.Model):
+    class Meta:
+        db_table = "core_setting"
+
+    key = models.CharField(max_length=255)
```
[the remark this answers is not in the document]
Alex-Izquierdo: Also, should it be better "name"? to keep naming consistency.

Alex-Izquierdo, line 31: Since we are going always to fetch a setting by his name, I think the table should have an index for that field.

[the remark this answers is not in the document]
bzwei: This is to keep the schema same as what is in controller.

src/aap_eda/conf/registry.py
```diff
@@ -0,0 +1,56 @@
...
+        if setting in self._registry:
+            raise ImproperlyConfigured(
+                f"Setting {setting} is already registered."
+            )
+        self._registry[setting] = kwargs
+
+    def unregister(self, setting):
+        self._registry.pop(setting, None)
+
+    def get_registered_settings(self, read_only=None):
```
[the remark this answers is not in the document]
bzwei: copied from controller's code

src/aap_eda/conf/registry.py
```diff
@@ -0,0 +1,145 @@
...
+    def register(self, registry_data: RegistryData) -> None:
+        if registry_data.name in self._registry:
+            raise ImproperlyConfigured(
+                f"Setting {registry_data.name} is already registered."
+            )
+        self._registry[registry_data.name] = registry_data
+        if registry_data.read_only:
+            update_method = models.Setting.objects.update_or_create
+        else:
+            update_method = models.Setting.objects.get_or_create
```
Alex-Izquierdo, line 104: This should not be the contrary?
If is read_only, get_or_create, if is not read-only, update_or_create

hsong-rh:
@bzwei We need another env variable named INSTALL_UUID. As @Ladas mentioned, it is generated via migration in the controller. [https://github.com/ansible/awx/blob/d411354f368640d4f17df8652f703f815d5fac15/awx/main/migrations/0069_v350_generate_unique_install_uuid.py#L25](https://www.google.com/url?q=https://github.com/ansible/awx/blob/d411354f368640d4f17df8652f703f815d5fac15/awx/main/migrations/0069_v350_generate_unique_install_uuid.py%23L25&sa=D&source=docs&ust=1726239794475188&usg=AOvVaw1Le8WMpwWV3nhvfnwcOgmf)

src/aap_eda/conf/registry.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+
+    def unregister(self, setting: str) -> None:
+        self._registry.pop(setting, None)
+
+    def get_registered_settings(self, read_only: bool = False) -> list[str]:
+        setting_names = []
+
+        for setting, data in self._registry.items():
+            if data.read_only and read_only:
```
Alex-Izquierdo, line 116: If the setting is read_only and the param read_only is True, the setting is skipped. According to the function signature, I understand that read_only=True would return only read-only values. 

If the intent is to skip them, should not be clearer to call the param "skip_read_only"?

src/aap_eda/settings/default.py
```diff
@@ -765,3 +765,8 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
 MAX_PG_NOTIFY_MESSAGE_SIZE = int(
     settings.get("MAX_PG_NOTIFY_MESSAGE_SIZE", 6144)
 )
+
+AUTOMATION_ANALYTICS_URL = settings.get("AUTOMATION_ANALYTICS_URL", "")
+INSIGHTS_CERT_PATH = settings.get(
+    "INSIGHTS_CERT_PATH", "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
```
hsong-rh, line 771: The default should be an empty string. Otherwise, it will complain no certificate file is found when the default path is invalid.

src/aap_eda/conf/registry.py
bzwei: If it is read_only, we allow internal update via coding (product upgrading or set ENV). If not, we shall keep the value set by user.

src/aap_eda/conf/registry.py
bzwei: The parameter is coped from controller, but your point is valid

src/aap_eda/conf/registry.py
```diff
@@ -0,0 +1,145 @@
...
+        self._registry[registry_data.name] = registry_data
+        if registry_data.read_only:
+            update_method = models.Setting.objects.update_or_create
+        else:
+            update_method = models.Setting.objects.get_or_create
+        update_method(
+            key=registry_data.name, defaults={"value": registry_data.default}
+        )
+
+    def unregister(self, setting: str) -> None:
```
Alex-Izquierdo, line 109: This is method is not used. What would be its purpose? I don't see why we would want to do that.

bzwei: again. it is copied. I will remove

bzwei:
@hsong-rh We can use service_id for INSTALL_UUID
> @bzwei We need another env variable named INSTALL_UUID. As @Ladas mentioned, it is generated via migration in the controller. we need to follow the same pattern:[https://github.com/ansible/awx/blob/d411354f368640d4f17df8652f703f815d5fac15/awx/main/migrations/0069_v350_generate_unique_install_uuid.py#L25](https://www.google.com/url?q=https://github.com/ansible/awx/blob/d411354f368640d4f17df8652f703f815d5fac15/awx/main/migrations/0069_v350_generate_unique_install_uuid.py%23L25&sa=D&source=docs&ust=1726239794475188&usg=AOvVaw1Le8WMpwWV3nhvfnwcOgmf)

hsong-rh:
> @hsong-rh We can use service_id for INSTALL_UUID
> 
@bzwei how to get it in EDA?

bzwei:
> > @hsong-rh We can use service_id for INSTALL_UUID
> 
> @bzwei how to get it in EDA?
`from ansible_base.resource_registry.models.service_identifie import service_id`

hsong-rh:
@bzwei are you going to add pytests for the changes?

src/aap_eda/conf/registry.py
```diff
@@ -0,0 +1,190 @@
...
+    def is_setting_encrypted(self, key: str) -> bool:
+        return self._registry[key].encrypted
+
+    def is_setting_read_only(self, key: str) -> bool:
+        return self._registry[key].read_only
+
+    def get_setting_type(self, setting: str) -> type:
+        return self._registry[setting].type
+
+    def db_update_setting(self, key: str, value: Any) -> None:
```
Alex-Izquierdo, line 138: We should not validate here the type and read_only properties too?

src/aap_eda/conf/registry.py
```diff
@@ -0,0 +1,190 @@
...
+        return setting_names
+
+    def is_setting_encrypted(self, key: str) -> bool:
+        return self._registry[key].encrypted
+
+    def is_setting_read_only(self, key: str) -> bool:
+        return self._registry[key].read_only
+
+    def get_setting_type(self, setting: str) -> type:
+        return self._registry[setting].type
```
Alex-Izquierdo, line 136: Minor: this should use "key" for consistency with the other methods, right?
```suggestion
    def get_setting_type(self, key: str) -> type:
        return self._registry[key].type
```

src/aap_eda/conf/registry.py
```diff
@@ -0,0 +1,190 @@
...
+    ...
+
+
+@dataclass
+class RegistryData(object):
+    name: str
+    default: Any
+    type: type = str
+    read_only: bool = False
+    encrypted: bool = False
```
Alex-Izquierdo, line 44: since all the values are actually encrypted, I feel this name confusing. Would not be better to name "is_secret" ?

src/aap_eda/core/models/setting.py
```diff
@@ -0,0 +1,30 @@
...
+
+__all__ = ("Setting",)
+
+
+class Setting(models.Model):
+    class Meta:
+        db_table = "core_setting"
+        indexes = [models.Index(fields=["key"])]
+
+    key = models.CharField(max_length=255)
```
Alex-Izquierdo, line 27: Is there a reason to not enforce the uniqueness of the key at the DB layer? I don't see a reason why we should allow duplicates here.

bzwei: Later when this table is expanded to store settings per user we will see key duplication. I added the uniqueness for now. Can lift it up when the time comes.

bzwei:
> @bzwei are you going to add pytests for the changes?

Yes when the code is finalized. Currently I am keeping on refactoring it.

bzwei:
I moved the seeding code to create_init_data otherwise it will fail the building docker image process when the api is added.

Alex-Izquierdo (commented):
LGTM

bzwei:
@hsong-rh unit tests added

hsong-rh (approved):
LGTM

Alex-Izquierdo (approved):
LGTM but I think we should have tests that check the data is correctly persisted in DB

bzwei:
> LGTM but I think we should have tests that check the data is correctly persisted in DB

test_application_setting sets and reads the data via database.

PR #1050, opened 2024-09-12
feat: config AUTHENTICATION_BACKENDS to use PrefixedUserAuthBackend

bzwei:
AAP-30375: Configure EDA to authenticate prefixed usernames

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -621,7 +621,7 @@ bcrypt = ["bcrypt"]
 
 [[package]]
 name = "django-ansible-base"
-version = "2024.9.2.0.dev8+g5f718b4"
+version = "2024.9.12.0.dev4+ga44d6bf"
 description = "A Django app used by ansible services"
 optional = false
 python-versions = ">=3.9"
@@ -653,7 +653,7 @@ testing = ["cryptography", "pytest", "pytest-django"]
 type = "git"
 url = "https://github.com/ansible/django-ansible-base.git"
 reference = "devel"
-resolved_reference = "5f718b41107f942967155dea0434e4802e3b5380"
+resolved_reference = "a44d6bfab27ffb33a2bd04604b0cb88c7dd7762d"
 
 [[package]]
 name = "django-crum"
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -327,6 +327,23 @@ REST_FRAMEWORK = {
     "EXCEPTION_HANDLER": "aap_eda.api.exceptions.api_fallback_handler",
 }
 
+
+def _config_authentication_backends():
+    from django.conf import settings as djsettings
+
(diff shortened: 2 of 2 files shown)


bzwei:
@AlanCoding please review

poetry.lock
```diff
@@ -621,7 +621,7 @@ bcrypt = ["bcrypt"]
 
 [[package]]
 name = "django-ansible-base"
-version = "2024.9.2.0.dev8+g5f718b4"
+version = "2024.9.12.0.dev4+ga44d6bf"
```
AlanCoding, line 624: I don't see a 9.12 tag

bzwei: This is auto generated by `poetry lock --no-update` command. The pyproject.toml has dependency `django-ansible-base = { git = "https://github.com/ansible/django-ansible-base.git", branch = "devel"`

src/aap_eda/settings/default.py
```diff
@@ -327,6 +327,23 @@ def _get_databases_settings() -> dict:
...
+
+def _config_authentication_backends():
+    from django.conf import settings as djsettings
+
+    backend = (
+        "ansible_base.lib.backends.prefixed_user_auth.PrefixedUserAuthBackend"
+    )
+    backends = djsettings.AUTHENTICATION_BACKENDS or []
+    if backend not in backends:
+        backends.insert(0, backend)
```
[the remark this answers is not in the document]
bzwei: I changed the insert to append. In my test `djsettings.AUTHENTICATION_BACKENDS` only has the default `django.contrib.auth.backends.ModelBackend`, after the append it reads
```
['django.contrib.auth.backends.ModelBackend',
'ansible_base.lib.backends.prefixed_user_auth.PrefixedUserAuthBackend'
 ]
 ```

poetry.lock
bzwei: @AlanCoding should we lock the dab dependency to the latest release tag?

AlanCoding (approved):
Looks great

bzwei:
It is agreed that this PR should be merged into main per the discussion at https://redhat-internal.slack.com/archives/C053CJ4JFP1/p1726180638452939. It becomes a blocker for https://github.com/ansible/aap-gateway/pull/545.

poetry.lock
Alex-Izquierdo: @bzwei I don't think so, In the past we have it pinned but it has been more annoying that helpful, because DAB is in constant development, so to enable any new change we have to bump the version. So we finally decided to work from HEAD. 

cc @Dostonbek1 

I created https://github.com/ansible/eda-server/pull/1045 to help us to update the poetry.lock so we don't have to take care of it.

Alex-Izquierdo (approved):

hsong-rh (approved):

PR #1052, opened 2024-09-16
[AAP-30722] add Redis cluster-related retry and timeout settings

jshimkus-rh:
Adds settings to EDA allowing manipulation of timeouts and retries for Redis cluster deployments.

### DIFF
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -136,16 +136,16 @@ def _get_secret_key() -> str:
 SECRET_KEY = _get_secret_key()
 
 
-def _get_debug() -> bool:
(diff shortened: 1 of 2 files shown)


Alex-Izquierdo (approved):
LGTM, but I think the new options, which are for redis, should have the MQ prefix for consistency, don't think you?

jshimkus-rh:
> LGTM, but I think the new options, which are for redis, should have the MQ prefix for consistency, don't think you?

@Alex-Izquierdo Makes sense.

Alex-Izquierdo (approved):

Alex-Izquierdo (approved):

rooftopcellist (approved):

PR #1059, opened 2024-09-18
Use DAB value as fallback value for service path setting

AlanCoding:
I kept hitting this error even when it should have been fixed:

```
  File "/app/venv/lib64/python3.11/site-packages/ansible_base/resource_registry/signals/handlers.py", line 103, in sync_to_resource_server_post_save
    sync_to_resource_server(instance, action)
  File "/app/venv/lib64/python3.11/site-packages/ansible_base/resource_registry/utils/sync_to_resource_server.py", line 73, in sync_to_resource_server
    client = get_resource_server_client(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/venv/lib64/python3.11/site-packages/ansible_base/resource_registry/rest_client.py", line 26, in get_resource_server_client
    return ResourceAPIClient(
           ^^^^^^^^^^^^^^^^^^
  File "/app/venv/lib64/python3.11/site-packages/ansible_base/resource_registry/rest_client.py", line 66, in __init__
    self.base_url = f"{service_url}/{service_path.strip('/')}/"
                                     ^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'strip'
```

My finding here is that this is (fortunately) a one-off problem in the eda-server settings processing for just this setting. I don't have any problem allowing the dynaconf setting override, so I made this patch to continue allowing that. But if there is _not_ an override present, then we need to fall back to the DAB value, and the traceback above is because it incorrectly gives the `None` value, after the DAB defaults are already loaded about 100 lines earlier in this file.

### DIFF
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -733,7 +733,15 @@ RESOURCE_SERVER = {
     "VALIDATE_HTTPS": settings.get("RESOURCE_SERVER__VALIDATE_HTTPS", False),
 }
 RESOURCE_JWT_USER_ID = settings.get("RESOURCE_JWT_USER_ID", None)
-RESOURCE_SERVICE_PATH = settings.get("RESOURCE_SERVICE_PATH", None)
+
+try:
+    service_path_default = RESOURCE_SERVICE_PATH  # noqa
+except NameError:
+    service_path_default = None
+
+RESOURCE_SERVICE_PATH = settings.get(
+    "RESOURCE_SERVICE_PATH", service_path_default
+)
 ANSIBLE_BASE_MANAGED_ROLE_REGISTRY = settings.get(
     "ANSIBLE_BASE_MANAGED_ROLE_REGISTRY", {}
 )


src/aap_eda/settings/default.py
```diff
@@ -728,7 +728,15 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
     "VALIDATE_HTTPS": settings.get("RESOURCE_SERVER__VALIDATE_HTTPS", False),
 }
 RESOURCE_JWT_USER_ID = settings.get("RESOURCE_JWT_USER_ID", None)
-RESOURCE_SERVICE_PATH = settings.get("RESOURCE_SERVICE_PATH", None)
+
+try:
+    service_path_default = RESOURCE_SERVICE_PATH  # noqa
+except NameError:
+    service_path_default = None
```
Alex-Izquierdo, line 735: Would it be clearer if this logic is encapsulated in the default value of dynaconf? That would be in line of how we do it for other settings
For example:
```
def _get_dab_resouce_service_path() -> Optional[str]:
    try:
        service_path_default = RESOURCE_SERVICE_PATH  # noqa
    except NameError:
        service_path_default = None
RESOURCE_SERVICE_PATH = settings.get(
    "RESOURCE_SERVICE_PATH", _get_dab_resouce_service_path()
)
```

Alex-Izquierdo (commented):
Right now only blocker things should go into main, we are working temporally in devel branch. Can you push there?

src/aap_eda/settings/default.py
AlanCoding: I tried this but it never felt right to me, because the method will be catching a `NameError` for a global variable, and that seems a little fragile. The way we load the settings into the current module namespace inherently makes things awkward. I put up another form of this in https://github.com/ansible/eda-server/pull/1061 that I still think reads slightly better.

AlanCoding:
Closing, re-opened, against devel.

AlanCoding:
Re-opening, because we want this against the `main` branch

zkayyali812 (approved):
/lgtm!

Alex-Izquierdo (approved):

AlanCoding:
"e2e tests on demand" failed with:

```
2024-09-27 12:50:35,287 Pod 5e58db8a2a6f7abe74ac022b68416137de8e799c86c9ee7f5d848b5eae91f953 has successfully exited. Activation completed. It will attempt to restart in 5 seconds according to the restart policy always.It may take longer if there is no capacity available.
Traceback (most recent call last):
  File "/home/runner/work/eda-server/eda-server/eda-qa/eda_qa/utils/activations.py", line 62, in wait_for_activation
    wait_for(
  File "/opt/hostedtoolcache/Python/3.9.20/x64/lib/python3.9/site-packages/wait_for/__init__.py", line 278, in wait_for
    raise TimedOutError(timeout_msg) from exc
wait_for.TimedOutError: Could not do 'Activation 5 did not reach the expected state(s) ['running'] within 150 seconds. Look at the activation output for more details.' at /home/runner/work/eda-server/eda-server/eda-qa/eda_qa/api/activations.py:76 in time
```

Alex-Izquierdo:
flaxy test, rerunning @AlanCoding  ;)

PR #1061, opened 2024-09-19
[devel] Use DAB value as fallback value for service path setting

AlanCoding:
Replaces https://github.com/ansible/eda-server/pull/1059 for devel branch, and includes the review comment.


AlanCoding:
I re-tested this with aap-dev, because it's really hard to be sure it works otherwise. Confirmed:

 - create user `rick` in the resource provider
 - edit rick username in https://localhost/api/eda/v1/users/, a PATCH to change username and other stuff
 - confirm this is reflected in the user details in the resource provider

AlanCoding:
Moving back to main branch

PR #1064, opened 2024-09-19
feat: add settings/system/ API endpoint

bzwei:
Only admin is allowed to list or update settings.

AAP-30794: new eda settings API endpoint

replace #1060


dhaustein:
This might need a companion PR in eda-qa to regenerate the API client at least.

hsong-rh (approved):

bzwei:
> This might need a companion PR in eda-qa to regenerate the API client at least.

should we create an ATF jira issue for integration test?

dhaustein:
@bzwei 
> should we create an ATF jira issue for integration test?
No I meant eda-qa, the upstream test suite. Since you are adding a new endpoint, we might need to update the test suite's api client and perhaps add a test or two.

src/aap_eda/api/serializers/setting.py
```diff
@@ -0,0 +1,40 @@
...
+    elif v_type == int:
+        field = serializers.IntegerField(required=False, read_only=read_only)
+    elif v_type == bool:
+        field = serializers.BooleanField(required=False, read_only=read_only)
+    elif v_type == dict:
+        field = serializers.JSONField(required=False, read_only=read_only)
+    else:
+        raise TypeError(f"unsupported type {v_type}")
+    setattr(SettingSerializer, key, field)
+    SettingSerializer._declared_fields[key] = field
```
Alex-Izquierdo, line 40: Wouldn't it make more sense to add this logic at the initialization of the SettingSerializer? 
Following my reasoning maybe is more flexible to just pass the data at the instantiation of the Serializer and let decide to the serializer what is the right field given the type. 

Here you can see some examples: https://joel-hanson.medium.com/advanced-serializer-usage-dynamically-modifying-fields-e7c3bc28efa6

What do you think?

bzwei: Creating fields outside the initializer runs only once while creating them inside the initializer runs for every instance. Nevertheless I moved the code to the initializer because it looks more elegant.

Alex-Izquierdo (approved):

src/aap_eda/api/serializers/setting.py
Alex-Izquierdo: in that case would be a singleton a better pattern?

hsong-rh (approved):

PR #1066, opened 2024-09-24
Redis cluster error handling

john-westcott-iv:
Tactical solution to make the workers and schedulers more resilient to Redis Cluster issues.
Ideally, we would make contributions to the `rq` library instead of needing to take these measures in the EDA code.


src/aap_eda/core/management/commands/scheduler.py
```diff
@@ -149,4 +152,47 @@ def handle(self, *args, **options) -> None:
...
+            try:
+                super().handle(*args, **options)
+            except (
+                redis.exceptions.TimeoutError,
+                redis.exceptions.ClusterDownError,
+                redis.exceptions.ConnectionError,
+            ) as e:
+                # If we got one of these exceptions but are not on a Cluster go
+                # ahead and raise it normally.
+                if type(self.connection) is not DABRedisCluster:
```
jshimkus-rh, line 167: Using 'not isinstance(self.connection, DABRedisCluster)' would provide some future-proofing.

src/aap_eda/core/management/commands/scheduler.py
```diff
@@ -149,4 +152,47 @@ def handle(self, *args, **options) -> None:
...
+                    raise
+
+                # There are a lot of different exceptions that inherit from
+                # ConnectionError.  So we need to make sure if we got that its
+                # an actual ConnectionError. If not, go ahead and raise it.
+                # Note:  ClusterDownError and TimeoutError are not subclasses
+                #        of ConnectionError.
+                if (
+                    issubclass(redis.exceptions.ConnectionError, e)
+                    and type(e) is not redis.exceptions.ConnectionError
```
jshimkus-rh, line 177: The issubclass usage is reversed.  As written it's checking that ConnectionError is a subclass of the exception instance.  This allows subclasses of ConnectionError to pass through rather than be raised. This possibly relates to the disconnect logic question.
Further, issubclass documentation indicates that it is a check on classes not between instances and classes.  So it's use here would be incorrect regardless.
If we're intending to raise ConnectionError subclasses an 'isinstance(e, ConnectionError) and type(e) is not ConnectionError' would do so.

src/aap_eda/core/tasking/__init__.py
jshimkus-rh, line 1: The same comments made against scheduler.py apply here except that the reversed usage of issubclass is possibly the source of the worker not exiting.

Alex-Izquierdo (approved):
Tested in containerized.

jshimkus-rh:
> /lgtm! This was validated on Containerized install.
> 
> I have not yet been able to validate on OCP

Jessica and John are working on that.

PR #1072, opened 2024-09-27
feat: [AAP-32189] Collect event streams used by each activation

hsong-rh:
https://issues.redhat.com/browse/AAP-32189

The collected data is stored at event_streams_by_activation_table.csv in the tar file. It looks like:

image

### DIFF
--- a/src/aap_eda/analytics/analytics_collectors.py
+++ b/src/aap_eda/analytics/analytics_collectors.py
@@ -1,3 +1,17 @@
+#  Copyright 2024 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
 import os
 import platform
 from datetime import datetime
@@ -206,6 +220,49 @@ def event_streams_table(
 ):
     query = _get_query(models.EventStream.objects, since, until)
     return _copy_table("event_streams", query, full_path)
+
+
+@register(
+    "event_streams_by_activation_table",
+    "1.0",
+    format="csv",
+    description="Data on event_streams used by each activation",
+)
+def event_streams_by_activation_table(
+    since: datetime, full_path: str, until: datetime, **kwargs
+):
+    activations = models.Activation.objects.filter(
+        Q(created_at__gt=since, created_at__lte=until)
+        | Q(modified_at__gt=since, modified_at__lte=until)
+    ).distinct()
+
+    event_streams = models.EventStream.objects.none()
+    for activation in activations:
+        event_streams |= activation.event_streams.all()
+
+    if not bool(event_streams):
+        return
+
+    event_streams = event_streams.extra(
+        select={
+            "activation_id": "core_activation_event_streams.activation_id",
+            "event_stream_id": "core_eventstream.id",
+        }
+    ).values(
+        "activation_id",
+        "event_stream_id",
+        "name",
+        "event_stream_type",
+        "eda_credential_id",
+        "owner",
+        "events_received",
+        "last_event_received_at",
+        "organization_id",
+    )
+
+    query = f"COPY ({event_streams.query}) TO STDOUT WITH CSV HEADER"
+
+    return _copy_table("event_streams_by_activation", query, full_path)
 
 
 @register(
--- a/src/aap_eda/analytics/analytics_collectors.py
+++ b/src/aap_eda/analytics/analytics_collectors.py
@@ -20,7 +20,7 @@ import distro
 from ansible_base.resource_registry.models.service_identifier import service_id
 from django.conf import settings
 from django.db import connection
-from django.db.models import Manager, Q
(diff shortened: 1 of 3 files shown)


jshimkus-rh (approved):

src/aap_eda/analytics/analytics_collectors.py
```diff
@@ -167,6 +181,49 @@ def event_streams_table(
...
+    ).values(
+        "activation_id",
+        "event_stream_id",
+        "name",
+        "event_stream_type",
+        "eda_credential_id",
+        "owner",
+        "events_received",
+        "last_event_received_at",
+        "organization_id",
```
zkayyali812, line 219: Im wondering if we actually intend to be capturing all these values for our analytics? Or if instead we should only capture a subset of these fields? 

Fields like name, owner and organization_id, I'm not sure we would want to submit as part of our analytics dat, as I'm thinking there is a chance this could be sensitive data? 

Im thinking we would need to confirm with @Ladas to be sure of the exact fields we want to collect and send?

hsong-rh: @zkayyali812 Yes, you're right, we need to confirm with @Ladas and @josephpisciotta . 
The reason I added some extra data is for some other MVPs stories, such as: 
```
- I wish to know the amount of events fired into eventstreams;
- I wish to know the number of event streams customers have;
- I wish to know the number of event streams per rulebook activation;
```

This single collected table may be enough for these cases.

Ladas: yeah, we shouldn't collect owner. Even if it's just user_id, it's considered PII

hsong-rh: @Ladas The owners' info was removed.

src/aap_eda/analytics/analytics_collectors.py
```diff
@@ -163,8 +176,84 @@ def decision_environments_table(
...
+        "test_content_type",
+        "test_content",
+        "test_headers",
+        "test_error_message",
+        "uuid",
+        "url",
+        "created_at",
+        "modified_at",
+        "events_received",
+        "last_event_received_at",
```
zkayyali812, line 203: I dont think we should be collecting the following fields for analytics due to potential PII concerns-

1. name
2. additional_data_headers
3. test_content perhaps? Does this have potential to be sensitive/identifiable?
4. test_headers (if potential to be sensitive/identifiable)
5. url

src/aap_eda/analytics/analytics_collectors.py
```diff
@@ -163,8 +176,84 @@ def decision_environments_table(
...
+        }
+    ).values(
+        "activation_id",
+        "event_stream_id",
+        "name",
+        "event_stream_type",
+        "eda_credential_id",
+        "events_received",
+        "last_event_received_at",
+        "organization_id",
```
zkayyali812, line 251: We should not collect `name` here i expect, as this may have sensitive/identifiable data which we should not collect for analytics

tests/integration/analytics/test_analytics_collectors.py
```diff
@@ -424,6 +422,52 @@ def test_event_streams_table_collector(
...
+
+            assert header == [
+                "activation_id",
+                "event_stream_id",
+                "name",
+                "event_stream_type",
+                "eda_credential_id",
+                "events_received",
+                "last_event_received_at",
+                "organization_id",
```
zkayyali812, line 457: Maybe we should do the inverse in this test, and ensure `name` is not collected here

zkayyali812 (changes requested):
Some comments added, we should ensure we are not collecting anything which has potential to be identifiable/sensitive for analytics

src/aap_eda/analytics/analytics_collectors.py
Ladas: I think name and type will be what we will show?

The rest yeah, I agree these could be sensitive and we shouldn't collect it.

hsong-rh: @zkayyali812 I will collect event streams' names and types and skip others.

src/aap_eda/analytics/analytics_collectors.py
hsong-rh: As @Ladas mentioned, the name and type of the event stream are needed.

tests/integration/analytics/test_analytics_collectors.py
hsong-rh: same as above.

src/aap_eda/analytics/analytics_collectors.py
zkayyali812: Perfect, thank you @hsong-rh

src/aap_eda/analytics/analytics_collectors.py
zkayyali812: Sounds great, I will resolve my comment

zkayyali812 (dismissed):

src/aap_eda/analytics/analytics_collectors.py
```diff
@@ -163,8 +176,77 @@ def decision_environments_table(
...
+    ).distinct()
+
+    event_streams = models.EventStream.objects.none()
+    for activation in activations:
+        event_streams |= activation.event_streams.all()
+
+    if not bool(event_streams):
+        return
+
+    event_streams = event_streams.extra(
```
Alex-Izquierdo, line 231: Using `extra` method is not recommended. Ref: https://docs.djangoproject.com/en/4.2/ref/models/querysets/#extra

We should be able to do using annotate as you do in https://github.com/ansible/eda-server/pull/1084/

bzwei (approved):

PR #1073, opened 2024-09-27
feat: add a periodic task to run gather_analytics

bzwei:
Run the task with internval set to AUTOMATION_ANALYTICS_GATHER_INTERVAL. Admin is able to change the interval and auto reschedule without restart

AAP-31635: Create a scheduler task to collect and send analytics data

### DIFF
--- a/src/aap_eda/conf/registry.py
+++ b/src/aap_eda/conf/registry.py
@@ -136,18 +136,20 @@ class SettingsRegistry(object):
         return self._registry[key].type
 
     def db_update_setting(self, key: str, value: Any) -> None:
-        self._validate_key(key, writable=True)
-        Setting.objects.filter(key=key).update(
-            value=self._setting_value(key, value)
-        )
+        self.db_update_settings({key: value})
 
     def db_update_settings(self, settings: dict[str, Any]) -> None:
+        interval = None
         with transaction.atomic():
             for key, value in settings.items():
                 self._validate_key(key, writable=True)
                 Setting.objects.filter(key=key).update(
                     value=self._setting_value(key, value)
                 )
+                if key == "AUTOMATION_ANALYTICS_GATHER_INTERVAL":
+                    interval = int(value)
+        if interval:
+            SettingsRegistry._reschedule_gather_analytics(interval)
 
     def db_get_settings_for_display(self) -> dict[str, Any]:
         keys = self.get_registered_settings()
@@ -187,6 +189,12 @@ class SettingsRegistry(object):
         if self.is_setting_secret(key) and value:
             return ENCRYPTED_STRING
         return value
+
+    @staticmethod
+    def _reschedule_gather_analytics(interval: int):
+        from aap_eda.tasks.analytics import reschedule_gather_analytics
+
+        reschedule_gather_analytics(interval)
 
 
 settings_registry = SettingsRegistry()
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -517,7 +517,12 @@ RQ_QUEUES = get_rq_queues()
 # Otherwise, the default name is used
 RULEBOOK_QUEUE_NAME = settings.get("RULEBOOK_QUEUE_NAME", "activation")
 
-RQ_STARTUP_JOBS = []
+RQ_STARTUP_JOBS = [
+    {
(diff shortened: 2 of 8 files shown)


src/aap_eda/conf/registry.py
```diff
@@ -140,14 +140,21 @@ def db_update_setting(self, key: str, value: Any) -> None:
...
         with transaction.atomic():
             for key, value in settings.items():
                 self._validate_key(key, writable=True)
                 Setting.objects.filter(key=key).update(
                     value=self._setting_value(key, value)
                 )
+                if key == "AUTOMATION_ANALYTICS_GATHER_INTERVAL":
+                    interval = int(value)
+        if interval:
+            SettingsRegistry._reschedule_gather_analytics(interval)
```
jshimkus-rh, line 157: We could save a little duplicate code by re-writing `db_update_setting` as a call to `db_update_settings({key: value})`

jshimkus-rh (dismissed):

jshimkus-rh (approved):

src/aap_eda/tasks/analytics.py
```diff
@@ -0,0 +1,74 @@
...
+    logger.info("Queue EDA analytics")
+    unique_enqueue(queue_name, ANALYTICS_JOB_ID, _gather_analytics)
+
+
+def _gather_analytics() -> None:
+    if not application_settings.INSIGHTS_TRACKING_STATE:
+        logger.info("INSIGHTS_TRACKING_STATE is not enabled")
+        return
+    logger.info("Collecting EDA analytics")
+    call_command("gather_analytics", "--ship")
```
hsong-rh, line 74: The insights-analytics-collector has `schduled` collection_type, it may be the right type for scheduler tasks. https://github.com/RedHatInsights/insights-analytics-collector/blob/main/insights_analytics_collector/collector.py#L31
In parity with the controller's `gather_analytics` system command, we don't accept `scheduled` options neither.

bzwei: The scheduler cannot set the collection_type unless the system command supports it. Is there a plan for adding the type?

hsong-rh: The command is parity with the controllers', you can call `AnalyticsCollector` directly, which allow you to set `content_type`

src/aap_eda/settings/default.py
```diff
@@ -517,7 +517,9 @@ def get_rq_queues() -> dict:
 # Otherwise, the default name is used
 RULEBOOK_QUEUE_NAME = settings.get("RULEBOOK_QUEUE_NAME", "activation")
 
-RQ_STARTUP_JOBS = []
+RQ_STARTUP_JOBS = [
+    {"func": "aap_eda.tasks.analytics.schedule_gather_analytics"},
```
Alex-Izquierdo, line 521: we should define here "job_id" argument. It will prevent potential duplicates or race conditions with restarts of the scheduler same as we do with other scheduled jobs. 
(notice that here is not "id" param, but "job_id" because it uses `scheduler.enqueue_at`

src/aap_eda/tasks/analytics.py
```diff
@@ -0,0 +1,78 @@
...
+        interval=interval,
+        id=ANALYTICS_SCHEDULE_JOB_ID,
+    )
+
+
+def reschedule_gather_analytics(new_interval: int, serializer=None) -> None:
+    try:
+        job = Job.fetch(ANALYTICS_SCHEDULE_JOB_ID, serializer=serializer)
+    except NoSuchJobError:
+        logger.warning(f"Job {ANALYTICS_SCHEDULE_JOB_ID} does not exist")
```
Alex-Izquierdo, line 54: Why are you just skipping the action? I think we should call `schedule_gather_analytics`if the job doesn't exist, otherwise we lose the track of the change triggered by the system registry until next restart of the scheduler

bzwei: This is a normal case, for example, in a local dev env where only api is running, and application_settings.AUTOMATION_ANALYTICS_GATHER_INTERVAL is set to a new value.

Alex-Izquierdo (approved):

zkayyali812 (approved):
/lgtm!!

hsong-rh (approved):
LGTM

PR #1077, opened 2024-10-01
Add support for some more missing settings

AlanCoding:
What I got merged in https://github.com/ansible/eda-server/pull/1059 was necessary, but didn't process everything that we needed. This should have been obvious to me, but the way I was testing didn't make everything immediately break at the time.

I'm trying to use a new method with `settings.exists` here, which I think it way way nicer than catching `NameError`. I haven't tested that it works with env vars (I'm pretty sure it will), but I'll try to post a confirmation of that soon.

### DIFF
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -690,9 +690,21 @@ ANSIBLE_BASE_JWT_KEY = settings.get(
     "ANSIBLE_BASE_JWT_KEY", "https://localhost"
 )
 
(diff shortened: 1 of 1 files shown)


AlanCoding:
```python
if "FOO" in settings:
    FOO = settings.FOO
```

Suggested by @rochacbruno should work, and might be preferable

Alex-Izquierdo:
@AlanCoding I have confirmed that works with envvars.

Alex-Izquierdo (approved):

zkayyali812 (approved):
/lgtm!

PR #1078, opened 2024-10-01
fix: add IsSystemAdmin permission check

bzwei:
AAP-32353: created superuser by gateway cannot access the system setting API

### DIFF
--- a/src/aap_eda/api/permissions.py
+++ b/src/aap_eda/api/permissions.py
@@ -1,0 +1,22 @@
+#  Copyright 2024 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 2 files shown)


Alex-Izquierdo (commented):
I wonder if a system auditor should have read permissions here. Have you check how it works in AWX? Maybe we should use `IsSystemAdminOrAuditor` from https://github.com/ansible/django-ansible-base/blob/5f1ca896a4dc94a4ed7f1da0dc371ddc4217669f/ansible_base/rbac/api/permissions.py#L20

bzwei:
> I wonder if a system auditor should have read permissions here. Have you check how it works in AWX? Maybe we should use `IsSystemAdminOrAuditor` from https://github.com/ansible/django-ansible-base/blob/5f1ca896a4dc94a4ed7f1da0dc371ddc4217669f/ansible_base/rbac/api/permissions.py#L20

This fix was copied from AWX. But our user model does not have auditor attribute so I have to remove it.

Alex-Izquierdo (approved):

zkayyali812 (approved):

PR #1087, opened 2024-10-07
Common Redis connection retry mechanism.

jshimkus-rh:
Derived from previous work adding retries specifically to scheduler and worker this change provides a common re-usable implementation for handling Redis connection issues.

This re-usable implementation replaces those previously used in scheduler and worker and extends the retry mechanism to those "lower-level" functions which have a direct use of django_rq or rq APIs with a Redis dependency. This insulates the users of those functions (e.g., the API container's usage of delete_rulebook_process()) from Redis connection issues.


Alex-Izquierdo:
We should test this in real deployments to make sure we don't introduce regressions.

Alex-Izquierdo (commented):
This PR changes in multiple places the way we import the modules. That is not a feature or improvement but just a code style concern. 

I would agree to change how we import the modules, but it is not a requirement for the goal of this PR. Only creates noise and more line changes that makes more difficult the peer review and can make harder potential conflicts in later backports. 

Such a change should be done in a different PR.

jshimkus-rh:
> This PR changes in multiple places the way we import the modules. That is not a feature or improvement but just a code style concern.

@Alex-Izquierdo Yes, it is not strictly a requirement for the functionality being added but it is more than just a code style concern.

it was utilized as part of the process to identify all the areas where that functionality was needed and to make it unambiguous, as far as reasonably possible, where those dependencies exist within the code base.  As such it makes it easier to identify those dependencies in future changes which helps with code changes and maintenance.

jshimkus-rh:
> We should test this in real deployments to make sure we don't introduce regressions.

@Alex-Izquierdo Done.  In a containerized deployment with redis cluster.

zkayyali812 (approved):
/lgtm! Thank you for putting this together. Interested to see how this code holds up and improves our stance against the armada of platform automated tests as well.

Alex-Izquierdo (approved):

PR #1088, opened 2024-10-07
fix: relying on heartbeat to turn to running status

bzwei:
Before we turn an activation status to running as soon as the rulebook container is running, but the rulebook engine may not be actually ready. This fix turns the status to running only when it receives the first heartbeat from the rulebook engine.

AAP-31225: event-stream connected Rulebook activations report running before they are actually able to receive events


Alex-Izquierdo (commented):
I really think the WS worker should not handle the status of the activation, it breaks the separation of concerns. If I remember correctly we removed in the past logic like one precisely because it can collide with other usecases
Why not just enforce it from the manager itself?

bzwei:
> I really think the WS worker should not handle the status of the activation, it breaks the separation of concerns. If I remember correctly we removed in the past logic like one precisely because it can collide with other usecases Why not just enforce it from the manager itself?

Yes. I am thinking of signaling the worker to change the status. This is a PoC when to change the status.

Alex-Izquierdo (commented):
I'm concerned about this approach. I think it is the same issue that your previous approach but with a layer of indirection. The WS worker still be who decides when the activation is running status and I think that is wrong. Is the manager who should determine it. 

If the problem is that we set the activation as running when the container is running and we want to set it as running when we receive the heartbeat, why not just check the heartbeat in the manager? We can do it inspecting the stats or looking the updated_at field as we already do to check unresponsiveness.

Alex-Izquierdo:
Also, if we change the meaning of "running" from the running pod to the reported process (heartbeat) we must consider the window between both events and apply a timeout to consider the activation failed if the pod started but we don't receive heartbeat.

src/aap_eda/core/enums.py
```diff
@@ -119,6 +119,7 @@ class DefaultCredentialType(DjangoStrEnum):
 # TODO: rename to "RulebookProcessRequest"
 class ActivationRequest(DjangoStrEnum):
     START = "start"
+    RUNNING = "running"
```
Alex-Izquierdo, line 122: As extension of the reasoning of [this comment](https://github.com/ansible/eda-server/pull/1088#pullrequestreview-2363157927). "Running" is a state, not a verb. The fact that your are adding such a state here is a clue that the approach might be mistaken.

bzwei:
> I'm concerned about this approach. I think it is the same issue that your previous approach but with a layer of indirection. The WS worker still be who decides when the activation is running and I think that is wrong. Is the manager who should determine it.
> 
> If the problem is that we set the activation as running when the container is running and we want to set it as running when we receive the heartbeat, why not just check the heartbeat in the manager? We can do it inspecting the stats or looking the updated_at field as we already do to check unresponsiveness.

We could let the manager to check the heartbeat. The status change will have a delay due to the fact that the monitor task is running at interval. My current solution is to let WS worker to notify the manager to change status almost immediately. The other approach close to your proposal is to let WS worker to notify the manager to run a monitor task where the manager could check the heartbeat result.

bzwei:
> Also, if we change the meaning of "running" from the running pod to the reported process (heartbeat) we must consider the window between both events and apply a timeout to consider the activation failed if the pod started but we don't receive heartbeat.

The monitor task already has logic to fail an activation due to not receiving heartbeats. I will investigate whether we need to make small changes for this use.

Alex-Izquierdo:
> > I'm concerned about this approach. I think it is the same issue that your previous approach but with a layer of indirection. The WS worker still be who decides when the activation is running and I think that is wrong. Is the manager who should determine it.
> > If the problem is that we set the activation as running when the container is running and we want to set it as running when we receive the heartbeat, why not just check the heartbeat in the manager? We can do it inspecting the stats or looking the updated_at field as we already do to check unresponsiveness.
> 
> We could let the manager to check the heartbeat. The status change will have a delay due to the fact that the monitor task is running at intervals. My current solution is to let WS worker to notify the manager to change status almost immediately. The other approach close to your proposal is to let WS worker to notify the manager to run a monitor task where the manager could check the heartbeat result.

But we don't need the WS to notify anything, what I'm saying is just wait for the heartbeat in the manager as part of the start operation. There is where we also handle a timeout for the time window between the start of the pod and the reception of the heartbeat.  The WS worker should be agnostic about the livecycle of the pod and "running" is not a verb/request.

Alex-Izquierdo:
> > But we don't need the WS to notify anything, what I'm saying is just wait for the heartbeat in the manager as part of the start operation. There is where we also handle a timeout for the time window between the start of the pod and the reception of the heartbeat. The WS worker should be agnostic about the livecycle of the activation and "running" is not a verb/request.
> 
> My concerning is the timing for user experience. We want to show the user as soon as the activation is in running. We can let the manager to detect the status change but let the WS worker to request an additional monitor task besides the scheduled ones.

The activation can take time also to download the image. Once we want to consider the activation running when the heartbeat happens instead of when the pod is running we assume that it might take more time.

The monitor is an unique periodic task, you can not schedule it. Adding the request to the db doesn't trigger anything, all the requests are processed by the monitor except the delete operation. This was done in that way to solve race conditions. 

Adding such a "running" request in the queue only can make the things worse: between the request is done and is processed by the monitor the pod could die and you would have an activation in "running" status until the next loop where the manager would detect the issue. Or even worse, the activation might be stuck in starting state if for some reason the "running" request never happens. 

The whole point is that the WS worker should not determine the status of the activation, either "running" should not be a request. 

IMO what we have to do is to wait for the heartbeat with a timeout as part of the start operation.

Alex-Izquierdo:
There is another potential race condition, you can change the status to running when it is already in place another operation scheduled previously by the monitor.

bzwei:
@Alex-Izquierdo The manager now detects the condition to change status to running as part of the monitoring process. The same process changes status to failed if the container died. The WS worker only encourages the manager to run the monitoring process when it receives the heartbeat.

src/aap_eda/services/activation/activation_manager.py
```diff
@@ -251,12 +251,7 @@ def _start_activation_instance(self):
             self._error_activation(msg)
             raise exceptions.ActivationStartError(msg) from exc
 
-        # update status
-        self.set_status(ActivationStatus.RUNNING)
-        with transaction.atomic():
-            self.set_latest_instance_status(ActivationStatus.RUNNING)
-            self._set_activation_pod_id(pod_id=container_id)
-        self._reset_failure_count()
+        self._set_activation_pod_id(pod_id=container_id)
```
Alex-Izquierdo, line 254: So, if understand correctly what are you doing is to create the activation and leave it in starting state and wait for next loop of the monitor to change it as running, I disagree with this approach for some reasons:

1. It is not efficient. I think "starting" state should be a transitory state for the start operation and it should end up always in a "running" or "failed | error" state in the same loop. 

2. There is no a explicit predictable timeout. The time we allow for ansible-rulebook to report the first heartbeat is the time between monitor loop executions, which is variable. 

3 RULEBOOK_LIVENESS_TIMEOUT_SECONDS is set to 10 minutes!!! This is pretty excessive as a timeout for a starting activation. Why RULEBOOK_LIVENESS_TIMEOUT_SECONDS and RULEBOOK_LIVENESS_CHECK_SECONDS have so high values is a different story. 

4. We don't handle the lack of heartbeat as a starting error, being handled as unresponsive activation which is misleading for the user because in that case, the activation never started. 

In my opinion: 
It is this point of the code when we should wait for the heartbeat and set the activation and activation_instance as running. We should have a new system setting to define the timeout to receive the first heartbeat with a reasonable value, let's say 15? 30 seconds? 
If we reach the timeout, we can continue with the failed policy with a message indicating the process didn't start in time.

Alex-Izquierdo:
Changing the criteria by which we consider an activation as “running” sets a new requirement on ansible-rulebook logic, since eda-server now would depend on ansible-rulebook to send the heartbeat unconditionally at startup. This is something we need to always keep in mind from now on. @mkanoor @bzwei @hsong-rh @jshimkus-rh

src/aap_eda/wsapi/consumers.py
```diff
@@ -161,6 +163,9 @@ def handle_heartbeat(self, message: HeartbeatMessage) -> None:
                 message.stats["ruleSetName"]
             ] = message.stats
             activation.save(update_fields=["ruleset_stats"])
+
+            if activation.status == ActivationStatus.STARTING:
+                orchestrator.monitor_rulebook_processes()
```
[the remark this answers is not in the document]
bzwei: I understand what you said. But I don't think it will cause any conflicts because it is an unique task. What I should do here is to call `enqueue_monitor_rulebook_processes` instead of `monitor_rulebook_process()`. It does not matter it is called by a scheduler or any other code.

src/aap_eda/services/activation/activation_manager.py
mkanoor: @bzwei While testing this feature have you seen what happens if you don't send matching events to ansible-rulebook. The Heartbeat is controlled in the settings RULEBOOK_LIVENESS_TIMEOUT_SECONDS so if the user decides to make this as 5 minutes we won't get any session stats for 5 minutes.

bzwei: A new commit is pushed with a new constant RULEBOOK_READINESS_TIMEOUT_SECONDS and messages distinguish the timeout is due to Liveness check or Readiness check

Alex-Izquierdo: @mkanoor @bzwei I didn't check it, I assumed that ansible-rulebook will send the heartbeat when it starts regardless of the matching events. Is this true?  If this is not case, we can not use the heartbeat to transition an activation from starting to running.

Alex-Izquierdo, line 254: I still thinking that we should not wait for the next cycle of the monitor to move to a running state and it should be  here where we check the readiness as part of the start verb, but I won't block the PR if the rest of team agreed with this approach.

Alex-Izquierdo: @bzwei maybe is because you are using rulebooks that triggers a rule at the beginning?

src/aap_eda/wsapi/consumers.py
```diff
@@ -161,6 +163,9 @@ def handle_heartbeat(self, message: HeartbeatMessage) -> None:
                 message.stats["ruleSetName"]
             ] = message.stats
             activation.save(update_fields=["ruleset_stats"])
+
+            if activation.status == ActivationStatus.STARTING:
+                orchestrator.enqueue_monitor_rulebook_processes()
```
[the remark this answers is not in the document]
bzwei: no, it get execute immediately, if the worker is free. It does not wait for 5 seconds. In my tests I almost did not see the status starting on UI. It just changes from pending to running. But without this call, I can see sometimes the status is on starting.

src/aap_eda/services/activation/activation_manager.py
```diff
@@ -905,17 +909,21 @@ def monitor(self):
         # Detect unresponsive activation instance
         # TODO: we should decrease the default timeout/livecheck
         # in the future might be configurable per activation
-        if self._is_unresponsive():
+        check_readiness = (
+            self.latest_instance.status == ActivationStatus.STARTING
+        )
+        check_type = "Readiness" if check_readiness else "Liveness"
+        if self._is_unresponsive(check_readiness):
```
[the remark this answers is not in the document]
bzwei: sure I can change. For boolean I agree we should not simply pass it by True or False. Here it is a named variable.

src/aap_eda/services/activation/activation_manager.py
bzwei: As I explained above we don't want 5 seconds to run the monitor. The monitor task runs as soon as the heartbeat is received.

Alex-Izquierdo:
> @bzwei @Alex-Izquierdo The session stats are sent when the heart beat timer expires. It doesn't wait for an event to arrive. My question was more about we would now be tied to the heart beat interval, if the customer sets this to 10 minutes we will see the first session stat come up after 10 minutes.

This needs to be tested more, according to another comment of @bzwei he receives the stats at startup, not after 5 minutes (the current default)

src/aap_eda/services/activation/activation_manager.py
Alex-Izquierdo: > The monitor task runs as soon as the heartbeat is received.

That is not accurate @bzwei   You schedule the monitor task. In your tests it happens fastly because your local env is powerful and mostly idle with 1? 2? 3 concurrent activations? so the workers are idle and everything happens quickly.

In an environment with more workload and/or worse performance, the workers can take much more time to execute both, the monitor or the activation.monitor. 

The point is the fact that you are waiting for the next loop, no matter how fast it happens. It is not only a performance concern but a conceptual one, is the readiness check part of the start flow or the monitor flow?

At the container level, we are not being consistent either. In k8s [we are waiting for it](https://github.com/ansible/eda-server/blob/e05f0359cc859dffb60daa7f371f75b4d3c06270/src/aap_eda/services/activation/engine/kubernetes.py#L496, but we don't do it in podman. 

**As I said, I won't block the PR for this point if the rest of the team agreed with the approach. I don't think it is wrong, Its just I feel it should better to do it as part of the start flow.** ☮️

src/aap_eda/services/activation/activation_manager.py
```diff
@@ -356,18 +351,24 @@ def _is_already_stopped(self) -> bool:
...
+            timeout = settings.RULEBOOK_READINESS_TIMEOUT_SECONDS
+        elif self.db_instance.status in [
             ActivationStatus.RUNNING,
             ActivationStatus.STARTING,
         ]:
-            cutoff_time = timezone.now() - timedelta(
-                seconds=settings.RULEBOOK_LIVENESS_TIMEOUT_SECONDS,
-            )
-            return self.latest_instance.updated_at < cutoff_time
+            previous_time = self.latest_instance.updated_at
```
mkanoor, line 363: @bzwei Can this instance be updated in the DB for any other reason other than the arrival of the session stats? I mean if some other code were to update the instance would we declare that the rulebook is running. Should we check if the session_stats is present and not empty.

Alex-Izquierdo: Right now I think is not the case, but I think it is a very valid concern, I agree to enhance this check.

bzwei: there is no other place that the updated_at is set. I will enhance the code to use current time just in case the session stat does not have the timestamp.

bzwei:
I will fix e2e tests by relaxing some tolerance time window

mkanoor:
@bzwei Currently we don't listen for events from podman or Kubernetes engine to see when a state has changed so we can kick off a monitor task on Activation to check the state of Activation. Our current approach is a scheduled task that runs every [5 seconds](https://github.com/ansible/eda-server/blob/ae8bfc0ffc7393ab672aada24dc096ce94732f99/src/aap_eda/settings/default.py#L528). If we decide to use the events coming from podman/k8s and increase the refresh interval to say 10 minutes we won't be able to get any activity for 10 minutes. Should we instead call monitor request when ws_consumer gets a session stats and ask the activation manager to update the Activations internal state. That approach would work if we decide to switch to using events. Currently by running every 5 seconds we are running too many monitor tasks even if nothing has changed in Activation.

Alex-Izquierdo:
> @bzwei Currently we don't listen for events from podman or Kubernetes engine to see when a state has changed so we can kick off a monitor task on Activation to check the state of Activation. Our current approach is a scheduled task that runs every [5 seconds](https://github.com/ansible/eda-server/blob/ae8bfc0ffc7393ab672aada24dc096ce94732f99/src/aap_eda/settings/default.py#L528). If we decide to use the events coming from podman/k8s and increase the refresh interval to say 10 minutes we won't be able to get any activity for 10 minutes. Should we instead call monitor request when ws_consumer gets a session stats and ask the activation manager to update the Activations internal state. That approach would work if we decide to switch to using events. Currently by running every 5 seconds we are running too many monitor tasks even if nothing has changed in Activation.

@mkanoor that approach is not compatible with our current design. The monitor is a [control loop](https://kubernetes.io/docs/concepts/architecture/controller/) and we should not bind it to the WS worker who has a different responsibility, especially if the heartbeat has so high value. 

The point is if to include the first heartbeat as part of the start flow or leverage to the next loop. @bzwei has decided this second approach and IMO any other major change is out of the scope of the fix.

bzwei:
@mkanoor My current implementation is already what you described.

Alex-Izquierdo (approved):

mkanoor (dismissed):

mkanoor (approved):

Alex-Izquierdo (approved):

PR #1092, opened 2024-10-09
fix: remove event_stream_type from the payload when creating/updating event streams

hsong-rh:
https://issues.redhat.com/browse/AAP-33158


bzwei (approved):

Alex-Izquierdo:
@hsong-rh Have you check if the UI is impacted by this change?

hsong-rh:
> @hsong-rh Have you check if the UI is impacted by this change?

@Alex-Izquierdo I tested in my local, from UI to create different types of event streams. It worked properly for us.

mkanoor (approved):

mkanoor:
Checked with @lgalis we should be ok with this change.

PR #1093, opened 2024-10-09
fix: stop running instance if activation restart failed at validation

bzwei:
Upon user requested restart, if the validatin fails, not only we need to mark the activation status as error, but also to stop its instance that is still running.

AAP-29755: Rulebook Status should be the last running job status.


src/aap_eda/api/views/activation.py
```diff
@@ -415,6 +415,10 @@ def restart(self, request, pk):
 
         valid, error = is_activation_valid(activation)
         if not valid:
+            stop_rulebook_process(
```
Alex-Izquierdo, line 418: According with [what we do later,](https://github.com/ansible/eda-server/blob/327dce4904b1106ebfd283b9af5d8889dabe4523/src/aap_eda/api/views/activation.py#L432) we should check redis prior to call this function. Anyway, I want to confirm first with @jshimkus-rh this logic, because I think it is a mistake. Those requests are enqueued in the DB and we don't require connectivity with redis.

bzwei: I moved `self.redis_is_available()` before the validation. Even though requests are enqueued in the DB, they need redis to be consumed.

Alex-Izquierdo (approved):

src/aap_eda/api/views/activation.py
jshimkus-rh: @bzwei @josephpisciotta 
That's too early.  If you look at the other locations where `self.redis_is_available()` is used it's always *after* performing validation of the request; i.e., first make certain the user has given us a valid request and only then tell the user "sorry, can't do it: no redis." 

We could allow that behavior to differ in this particular scenario.
However, presuming we want to try to maintain it (in the non-faulted case) we'd need to move `stop_rulebook_process` to after the db update.

As long as redis is available that would preserve the existing behavior; e.g.,
```
if not valid:
    update db
    self.redis_is_available()
    stop_rulebook_process(..)
    return Response(...)
```

But if redis were faulted that'd give it precedence over the validation fault.  It might also be something of a surprise to the user to get the redis error but see that the db is updated.

Alternately we could catch the `status.HTTP_409_CONFLICT` from `self.redis_is_available()`, ignore it and return the validation fault.  That'd be, imo, most consistent with previous behavior but obviously `stop_rulebook_process` wouldn't get run and there'd be no indication of that.

Lastly we could simply return the validation error and *not* update the db.  That would be a large behavioral difference.

jshimkus-rh: That makes sense.

jshimkus-rh (approved):

zkayyali812 (approved):

PR #1101, opened 2024-10-11
fix: add EDA_PODMAN_USER_ID env var use in compose files

jshimkus-rh:
This allows the composes to also function on macOS.


Alex-Izquierdo (changes requested):
If we need this for mac, then we need to hold on and investigate further because this line was the cause of the failures when running e2e tests.

jshimkus-rh:
> If we need this for mac, then we need to hold on and investigate further because this line was the cause of the failures when running e2e tests.

We can do that.  The previous change title implied it was the data part which this doesn't put back, only the user parameter.

Alex-Izquierdo:
> > If we need this for mac, then we need to hold on and investigate further because this line was the cause of the failures when running e2e tests.
> 
> We can do that. The previous change title implied it was the data part which this doesn't put back, only the user parameter.

I added the user to mount the volume, previously is was not there. But I don't know if it was tested in mac anyway.

jshimkus-rh:
> > > If we need this for mac, then we need to hold on and investigate further because this line was the cause of the failures when running e2e tests.
> > 
> > 
> > We can do that. The previous change title implied it was the data part which this doesn't put back, only the user parameter.
> 
> I added the user to mount the volume, previously is was not there. But I don't know if it was tested in mac anyway.

I've been using docker-compose-dev.yaml for months on macOS.  Doston is likely also using it as he was having some trouble and I told him what I was doing and he tried it and was happy (so it seemed 😄).  It wasn't until the recent change to address the ci concern that it started to fail on macOS.  I deleted all the images, volumes, etc. and it was still failing.

Alex-Izquierdo (approved):

zkayyali812 (approved):

PR #1102, opened 2024-10-15
feat: [AAP-33453] - Collect the stats data of jobs which launched by the EDA

hsong-rh:
https://issues.redhat.com/browse/AAP-33453

The job stats data is collected and saved in the file `jobs_stats.json`. It looks like:
```
{
  "8018": [
    {
      "job_id": "8018",
      "type": "run_job_template",
      "created_at": "2023-12-14T15:19:02.319506Z",
      "status": "successful",
      "url": "https://controller_1/#/jobs/playbook/8018/details/",
      "install_uuid": "9509fade-acf8-4ff8-b5d1-6e58e01dceb9"
    },
    {
      "job_id": "8018",
      "type": "run_workflow_template",
      "created_at": "2023-12-14T15:19:02.326503Z",
      "status": "successful",
      "url": "https://controller_2/#/jobs/workflow/8018/details/",
      "install_uuid": "6e78176d-87ba-4837-9316-bf84660ca77d"
    }
  ],
  "8020": [
    {
      "job_id": "8020",
      "type": "run_workflow_template",
      "created_at": "2023-12-14T15:19:02.327547Z",
      "status": "successful",
      "url": "https://controller_1/#/jobs/workflow/8020/details/",
      "install_uuid": "9509fade-acf8-4ff8-b5d1-6e58e01dceb9"
    }
  ]
}
```

### DIFF
--- a/src/aap_eda/analytics/analytics_collectors.py
+++ b/src/aap_eda/analytics/analytics_collectors.py
@@ -9,6 +9,10 @@ from django.db import connection
 from django.db.models import Manager, Q
 from insights_analytics_collector import CsvFileSplitter, register
 
+from aap_eda.analytics.utils import (
+    collect_controllers_info,
+    extract_job_details,
+)
 from aap_eda.core import models
 from aap_eda.utils import get_eda_version
 
@@ -38,6 +42,43 @@ def config(**kwargs) -> dict:
         "eda_version": get_eda_version(),
         "eda_deployment_type": settings.DEPLOYMENT_TYPE,
     }
+
+
+@register(
+    "jobs_stats",
+    "1.0",
+    description="Stats data for jobs",
+)
+def jobs_stats(since: datetime, full_path: str, until: datetime, **kwargs):
+    stats = {}
+    audit_actions = _get_audit_action_qs(since, until)
+
+    if not bool(audit_actions):
+        return stats
+
+    controllers = collect_controllers_info()
+    for action in audit_actions.all():
+        job_type, job_id, install_uuid = extract_job_details(
+            action.url, controllers
+        )
+        if not job_type:
+            continue
+
+        data = stats.get(job_id, [])
+        job_stat = {}
+        job_stat["job_id"] = job_id
+        job_stat["type"] = job_type
+        job_stat["created_at"] = action.fired_at.strftime(
+            "%Y-%m-%dT%H:%M:%S.%fZ"
+        )
+        job_stat["status"] = action.status
+        job_stat["url"] = action.url
+        job_stat["install_uuid"] = install_uuid
+        data.append(job_stat)
+
+        stats[job_id] = data
+
+    return stats
 
 
 @register(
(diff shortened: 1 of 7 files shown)


src/aap_eda/analytics/utils.py
```diff
@@ -23,3 +35,72 @@ def datetime_hook(dt: dict) -> dict:
...
+            headers = {"Authorization": f"{auth_value}"}
+            logger.info("Use Basic authentication to ping the controller.")
+
+        try:
+            resp = requests.get(url, headers=headers, verify=verify)
+            resp_json = resp.json()
+            controller_info["install_uuid"] = resp_json["install_uuid"]
+
+            info[host] = controller_info
+        except requests.exceptions.RequestException as e:
```
bzwei, line 75: Are we certain an invalid credential is the only cause for RequestException?

src/aap_eda/analytics/utils.py
```diff
@@ -23,3 +35,72 @@ def datetime_hook(dt: dict) -> dict:
...
+                f"Credential {credential.name} failed in authorization: {e}"
+            )
+
+    return info
+
+
+def extract_job_details(
+    url: str,
+) -> Tuple[Optional[str], Optional[str], Optional[str]]:
+    controllers = collect_controllers_info()
```
bzwei, line 86: For each job (url) you call collect_controllers_info() which does not change. We should only call it once per analytics task.

src/aap_eda/analytics/analytics_collectors.py
```diff
@@ -40,6 +41,40 @@ def config(**kwargs) -> dict:
...
+)
+def jobs_stats(since: datetime, full_path: str, until: datetime, **kwargs):
+    stats = {}
+    audit_actions = _get_audit_action_qs(since, until)
+
+    if not bool(audit_actions):
+        return stats
+
+    for action in audit_actions.all():
+        job_type, job_id, install_uuid = extract_job_details(action.url)
```
bzwei, line 57: Can action.url be empty? If the action is not a controller job, it does not have an url.

hsong-rh: the `action.url` can only be an empty string in its model definition, which is handled properly.

tests/integration/conftest.py
```diff
@@ -933,6 +933,31 @@ def default_aap_credential(
     )
 
 
+@pytest.fixture
+def new_aap_credential(
```
bzwei, line 937: Consider a better name for this fixture. Even `another_aap_credential` might be better than `new_aap_credential`.

Alex-Izquierdo: According to our name pattern for test fixtures, "new" prefix means it is a new randomized object so every test use a different instance of the object. Here is a static object of a specific type. 

Looking at how is used this fixture I think it would be better if you just create a batch of credentials for your test instead of combine `new_aap_credential` with `new_aap_credential` 

The fixture can be placed in a new conftest.py file only for tests/integration/analytics unless you expect to be reused outside that scope.

hsong-rh: New conftest.py file is created

zkayyali812 (approved):

bzwei (approved):

PR #1110, opened 2024-10-18
Revert pre-work for analytics

bzwei:
Those work will be committed to a feature branch


hsong-rh (approved):

hsong-rh (approved):

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.4 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
```
Alex-Izquierdo, line 1: @bzwei can you upgrade poetry and update the lock with 1.8.4?

Alex-Izquierdo (approved):

poetry.lock
bzwei: I upgraded to 1.8.4, and ran `poetry lock --no-update`. It did not generate a new lock file. We can update this file when a real change is detected.

Alex-Izquierdo: well, you have to restore the poetry lock so it is generated from the new version. anyway is not a big deal.

zkayyali812 (approved):
LGTM!
Once we have the feature branch for this created, can we note/link this branch name in the overall feature?

Perhaps this would be good to add in our Slack Canvas as well

PR #1111, opened 2024-10-18
fix: remove organization default from BaseOrgModel

jshimkus-rh:
The organization is a required parameter everywhere it appears in the API. Having a default for it runs counter to this and was only of value internally to EDA for eliminating the need to specify an organization when creating test instances of various models.  This introduced a dependency on the default organization that was not applicable to EDA's normal operation and resulted in the generation of faults if the name of the organization was changed.

This change modifies EDA's unit testing to use the pre-existing pytest fixture 'default_organization' in testing rather than a functionally unnecessary and fragile internal convenience.

This change does not completely eliminate the presumption of the existence of a default organization accessible by a specific name. This presumption, however, is restricted to the initial database population run once when an EDA deployment is started for the first time.


dhaustein:
Note: Quite a lot of the RBAC tests are in ATF. We need to do a pass over this once this is ready. Since we don't have a good way to run ATF over a branch in eda-server, I will do a run on my local machine.

Alex-Izquierdo:
@jshimkus-rh Can you refer to the internal ticket related with this PR please? 

I think what we need to do is to figure out if there is a way to get the default organization dynamically instead of from a name in the static settings, maybe DAB already provides a way. This would be an issue for other components, so I would bet it is already addressed.

jshimkus-rh:
> I think what we need to do is to figure out if there is a way to get the default organization dynamically instead of from a name in the static settings, maybe DAB already provides a way. This would be an issue for other components, so I would bet it is already addressed.

From an analysis of the usage of a default organization by EDA it was seen that the API requires the organization id everywhere it is used.  There is only one reason to actually refer to a default organization and that is the initial database population when an EDA deployment is started for the first time.  This change does not modify that.

The only reason to have a default on BaseOrgModel (which is the origin of the logged issue) is to circumvent the requirement of providing an organization when instantiating a model instance in EDA unit tests.  This change removes that and provides an organization as required.

zkayyali812 (approved):

bzwei (approved):

PR #1115, opened 2024-10-21
Run CI on feature branches

zkayyali812:
Leverages syntax defined here - https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#using-filters

Note that this currently just will run the CI, but will not upload the test results currently as defined in this task - https://github.com/ansible/eda-server/blob/main/.github/workflows/ci.yaml#L118

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -5,10 +5,14 @@ env:
 
 on:
   push:
-    branches: [ 'main' ]
+    branches: 
+      - main
+      - 'feature/**'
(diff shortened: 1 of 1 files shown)


Alex-Izquierdo:
I think we don't need it. The current config already runs the checks for every PR, I added the config in the repo to make the checks mandatory, that's all what we need. If we would have releases branches it could make sense to trigger it on push there, but it is not the case. Neither we are using merge groups, btw.

PR #1117, opened 2024-10-22
fix: make event_stream url a dynamic field

Alex-Izquierdo:
Event stream url was persisted in the url at the creation but it is composed by the environment variable `EDA_EVENT_STREAM_BASE_URL` The url field should be dynamic. 

Jira: https://issues.redhat.com/browse/AAP-33819


hsong-rh (approved):

zkayyali812 (approved):

mkanoor (approved):

jshimkus-rh (approved):

mkanoor:
@Alex-Izquierdo How would we leave an audit trail for this change by being dynamic it will always print the latest URL.
Should we keep the URL field in the DB and keep it to track the original value before it's updated.
When things start to fail because the URL has changed, if there was something in the log that can be traced back to when the user changed it would help in debugging.

Alex-Izquierdo:
> @Alex-Izquierdo How would we leave an audit trail for this change by being dynamic it will always print the latest URL. Should we keep the URL field in the DB and keep it to track the original value before it's updated. When things start to fail because the URL has changed, if there was something in the log that can be traced back to when the user changed it would help in debugging.

@mkanoor It is no different from any other configuration change. What happens when things start to fail because the ANY_SETTING has changed like websocket url or allowed hosts or the api_prefix? In this case the url is not user data, EVENT_STREAM_BASE_URL is not an input of the user. As webserver the application is agnostic about variables like the hostname and that's why it doesn't preserve any other url. 

Let's say we store the url, what value should have if the application is restarted three times with a different value each time? It doesn't make sense. 

I believe other components print in the application log the django settings at startup, that would be the way to track such a change. It is something that we can look as new feature.

PR #1122, opened 2024-10-23
fix: checking out a non-existing branch fails the project sync

bzwei:
Need to distinguish a git checkout failure is due to a non-existing branch/tag/commit, or due to an empty repository. The raw error messages are very similar but we have set the project sync state to completed for the later case in an earlier fix. This fix identifies the first case and set the state to be failed.

AAP-29144: Project sync should report status as "failed" when using non-existent branch

Also refactor the existing code to use a proper error class instead of. an error message to identify the case for an empty repo.

Will add e2e tests for the two cases.

### DIFF
--- a/src/aap_eda/services/project/imports.py
+++ b/src/aap_eda/services/project/imports.py
@@ -26,7 +26,7 @@ from django.core import exceptions
 
 from aap_eda.core import models
 from aap_eda.core.types import StrPath
-from aap_eda.services.project.scm import ScmRepository
+from aap_eda.services.project.scm import ScmEmptyError, ScmRepository
 from aap_eda.services.rulebook import insert_rulebook_related_data
 
 logger = logging.getLogger(__name__)
@@ -61,13 +61,12 @@ def _project_import_wrapper(
         try:
             func(self, project)
             project.import_state = models.Project.ImportState.COMPLETED
-        except Exception as e:
+        except ScmEmptyError as e:
             # if a project is empty, sync status should show completed
-            project.import_state = (
-                models.Project.ImportState.COMPLETED
-                if "Project folder is empty" in e.args[0]
-                else models.Project.ImportState.FAILED
-            )
+            project.import_state = models.Project.ImportState.COMPLETED
+            project.import_error = str(e)
+        except Exception as e:
+            project.import_state = models.Project.ImportState.FAILED
             project.import_error = str(e)
             error = e
         finally:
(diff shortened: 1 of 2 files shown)


hsong-rh:
@bzwei You will add pytest for the change, right?

bzwei:
> @bzwei You will add pytest for the change, right?

yes, new tests will be added to eda_qa

src/aap_eda/services/project/imports.py
```diff
@@ -61,13 +61,12 @@ def wrapper(self: ProjectImportService, project: models.Project):
...
-        except Exception as e:
+        except ScmEmptyError as e:
             # if a project is empty, sync status should show completed
-            project.import_state = (
-                models.Project.ImportState.COMPLETED
-                if "Project folder is empty" in e.args[0]
-                else models.Project.ImportState.FAILED
-            )
+            project.import_state = models.Project.ImportState.COMPLETED
+            project.import_error = str(e)
```
Alex-Izquierdo, line 67: It doesn't make sense to set an "import error" if the state is not error. Here I see a potential inconsistency.

src/aap_eda/services/project/imports.py
```diff
@@ -61,13 +61,12 @@ def wrapper(self: ProjectImportService, project: models.Project):
...
             project.import_state = models.Project.ImportState.COMPLETED
-        except Exception as e:
+        except ScmEmptyError as e:
             # if a project is empty, sync status should show completed
-            project.import_state = (
-                models.Project.ImportState.COMPLETED
-                if "Project folder is empty" in e.args[0]
-                else models.Project.ImportState.FAILED
-            )
+            project.import_state = models.Project.ImportState.COMPLETED
```
Alex-Izquierdo, line 66: should be this failed? in the description you say: 
> This fix identifies the first case and set the state to be failed.

src/aap_eda/services/project/imports.py
bzwei: As discussed in the SU. This field should be renamed in another issue when it is time to change UI together.

src/aap_eda/services/project/imports.py
bzwei: ScmEmptyError is for the second case where the repo is truly empty. We want it to be completed.

Alex-Izquierdo (approved):

zkayyali812 (approved):

PR #1123, opened 2024-10-24
fix: prevent multiple AAP credentials from being attached to activations

hsong-rh:
https://issues.redhat.com/browse/AAP-34025

400 validation error will be raised if multiple AAP credentials are attached when creating activations.


Alex-Izquierdo (dismissed):

src/aap_eda/api/serializers/activation.py
```diff
@@ -411,7 +411,10 @@ class Meta:
         required=False,
         allow_null=True,
         child=serializers.IntegerField(),
-        validators=[validators.check_multiple_credentials],
+        validators=[
+            validators.check_multiple_credentials,
+            validators.check_if_has_multiple_aap_credentials,
```
mkanoor, line 416: @hsong-rh Would a better name for this function be

* check_single_aap_credential

zkayyali812:
PR looks good, but It appears our test suite needs to be updated here as well

src/aap_eda/api/serializers/activation.py
hsong-rh: Changed

mkanoor (approved):

Alex-Izquierdo (approved):

PR #1124, opened 2024-10-24
fix: add event_streams into credentials refs list

hsong-rh:
https://issues.redhat.com/browse/AAP-34452

### DIFF
--- a/src/aap_eda/api/serializers/eda_credential.py
+++ b/src/aap_eda/api/serializers/eda_credential.py
@@ -12,6 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 from django.db.models import Q
+from django.urls import reverse
 from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 
@@ -192,13 +193,16 @@ def get_references(eda_credential: models.EdaCredential) -> list[dict]:
         Q(eda_credential=eda_credential)
         | Q(signature_validation_credential=eda_credential)
     )
+    used_event_streams = models.EventStream.objects.filter(
+        eda_credential=eda_credential
+    )
 
     for activation in used_activations:
         resource = {
             "type": "Activation",
             "id": activation.id,
             "name": activation.name,
-            "url": f"api/eda/v1/activations/{activation.id}/",
+            "uri": reverse("activation-detail", kwargs={"pk": activation.id}),
         }
         resources.append(resource)
 
@@ -207,8 +211,9 @@ def get_references(eda_credential: models.EdaCredential) -> list[dict]:
             "type": "DecisionEnvironment",
             "id": decision_environment.id,
             "name": decision_environment.name,
-            "url": (
-                f"api/eda/v1/decision-environments/{decision_environment.id}/"
+            "uri": reverse(
+                "decisionenvironment-detail",
+                kwargs={"pk": decision_environment.id},
             ),
         }
         resources.append(resource)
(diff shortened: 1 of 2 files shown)


src/aap_eda/api/serializers/eda_credential.py
```diff
@@ -222,4 +225,13 @@ def get_references(eda_credential: models.EdaCredential) -> list[dict]:
         }
         resources.append(resource)
 
+    for event_stream in used_event_streams:
+        resource = {
+            "type": "EventStream",
+            "id": event_stream.id,
+            "name": event_stream.name,
+            "url": f"api/eda/v1/event-streams/{event_stream.id}/",
```
Alex-Izquierdo, line 233: The value should not be hardcoded, paths are dynamic based on the settings and routing configuration. You should use `reverse` to the get the path as for example, I did [here](https://github.com/ansible/eda-server/pull/1117/files#diff-fefcb7e46a2ae907d8ac7af48e6065035f2a6017f0c427ced76f78072da75224R107). 

https://docs.djangoproject.com/en/5.1/ref/urlresolvers/


Also the name "url" might be misleading. This would not be an url strictly speaking because it doesn't contain protocol or domain. I think it would be better to say "url_path" or "uri" or "path"

src/aap_eda/api/serializers/eda_credential.py
```diff
@@ -207,8 +211,9 @@ def get_references(eda_credential: models.EdaCredential) -> list[dict]:
             "type": "DecisionEnvironment",
             "id": decision_environment.id,
             "name": decision_environment.name,
-            "url": (
-                f"api/eda/v1/decision-environments/{decision_environment.id}/"
+            "uri": reverse(
```
Alex-Izquierdo, line 214: is this field used in some any other place? When I suggested a rename because "url" would not be accurate I didn't notice that we are already using the same struct. This would be a change of the interface and we should make sure first it doesn't brake anything else. Take into account too that the previous value was without leading slash api/eda/v1/decision-environments/{decision_environment.id}/ but reverse returns it with a leading slash
/api/eda/v1/decision-environments/[pk]/

Alex-Izquierdo (approved):

src/aap_eda/api/serializers/eda_credential.py
hsong-rh: As far as I know, the `refs` are only used by the UI to display them to end users when they try to delete the credentials. @lgalis can you double-check? thanks

lgalis: @hsong-rh - I could not find a place where this url field is used in the ui.

bzwei (dismissed):

mkanoor (approved):

hsong-rh:
The e2e should be fixed at https://github.com/ansible/eda-qa/pull/438

src/aap_eda/api/serializers/eda_credential.py
```diff
@@ -32,7 +33,7 @@ class EdaCredentialReferenceSerializer(serializers.Serializer):
     name = serializers.CharField(
         required=True, help_text="Name of the related resource"
     )
-    url = serializers.URLField(
+    uri = serializers.URLField(
         required=True, help_text="URL of the related resource"
```
Alex-Izquierdo, line 37:
```suggestion
        required=True, help_text="URI of the related resource"
```

Alex-Izquierdo (dismissed):

Alex-Izquierdo (approved):

PR #1125, opened 2024-10-24
fix: update activation instance's started_at

bzwei:
The started_at used to be the instance created time, which is even before the DE is downloaded. The download time plus the container init time can exceed the default 30 seconds LIVENESS timeout window. This fix resets the started_at time to be at the moment after the container is initiated.

This fix should help to resolve the sporadic eda-qa's e2e test failures.

### DIFF
--- a/src/aap_eda/services/activation/activation_manager.py
+++ b/src/aap_eda/services/activation/activation_manager.py
@@ -77,6 +77,11 @@ class ActivationManager(StatusManager):
 
         self.container_logger_class = container_logger_class
 
+    def _update_started_time(self) -> None:
+        """Update latest instance's started_at to now."""
+        self.latest_instance.started_at = timezone.now()
+        self.latest_instance.save(update_fields=["started_at"])
+
     def _set_activation_pod_id(self, pod_id: tp.Optional[str]) -> None:
         """Set the pod id of the activation instance."""
         # TODO: implement db locking?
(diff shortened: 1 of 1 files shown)


src/aap_eda/services/activation/activation_manager.py
```diff
@@ -77,6 +77,11 @@ def __init__(
 
         self.container_logger_class = container_logger_class
 
+    def _update_started_time(self) -> None:
+        """Update latest instance's started_at to now."""
+        self.latest_instance.started_at = timezone.now()
+        self.latest_instance.save(update_fields=["started_at"])
+
```
hsong-rh, line 84: @bzwei do you also need to update the `updated_at` timestamp? which is also used in checking if is_unresponsive

bzwei: `updated_at` is only set when a heartbeat is received.

Alex-Izquierdo: Should we use a field ad-hoc for the heartbeat? updated_at is in the eda a field related to the DB record and its plausible to be updated on any other change not related to heartbeat. I feel we should track the heartbeat in a field explicitly dedicated for it.

dhaustein (approved):

src/aap_eda/services/activation/activation_manager.py
bzwei: @Alex-Izquierdo sorry I missed your comment before the merge. updated_at is a particular field added for the heartbeat. The one for DB record and auto generated is normally called modified_at, which is not enable in this table.

PR #1126, opened 2024-10-25
fix: check that DE image url is well-formed

jshimkus-rh:
https://issues.redhat.com/browse/AAP-30564

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,7 @@ xxhash = "3.4.*"
 pyjwt = { version = "2.7.*", extras = ["crypto"] }
 ecdsa = "0.18.*"
 distro = "^1.9.0"
+validators = "^0.34.0"
 
 [tool.poetry.group.test.dependencies]
 pytest = "*"
--- a/src/aap_eda/core/validators.py
+++ b/src/aap_eda/core/validators.py
@@ -17,6 +17,7 @@ import re
 import typing as tp
 import urllib
 
+import validators
 import yaml
 from django.conf import settings
 from django.utils.translation import gettext_lazy as _
@@ -59,31 +60,46 @@ def check_if_de_exists(decision_environment_id: int) -> int:
 
 
 def check_if_de_valid(image_url: str, eda_credential_id: int):
-    parsed_url = urllib.parse.urlparse(image_url)
-    base_message = f"Image url {image_url} is malformed; "
-    if parsed_url.scheme:
-        raise serializers.ValidationError(base_message + "scheme not allowed")
-    if parsed_url.netloc:
-        raise serializers.ValidationError(
-            base_message + "network location not allowed"
-        )
-    if parsed_url.params:
-        raise serializers.ValidationError(
-            base_message + "parameters not allowed"
-        )
-    if parsed_url.query:
-        raise serializers.ValidationError(base_message + "query not allowed")
-    if parsed_url.fragment:
-        raise serializers.ValidationError(
-            base_message + "fragment not allowed"
-        )
-
-    # Now that we've passed the above and know there's no netloc check if the
-    # path, rsplit into the <name> and <tag> (if any), matches their respective
-    # regexes as defined here:
+    # The OCI standard format for the image url is a combination of a host
+    # (with optional port) separated from the image path (with optional tag) by
+    # a slash: <host>[:port]/<path>[:tag].
+    #
     # https://github.com/opencontainers/distribution-spec/blob/8376368dd8aadc33bf6c88a8b765df90287bb5c8/spec.md?plain=1#L155 # noqa: E501
+    #
+    # We split the image url on the first slash into the host and path.  The
+    # path is further split into a name and tag on the rightmost colon.
+    #
+    # THe host portion is validated using the validators package and the path
+    # and tag are validated using the OCI regexes for each.
+    base_message = f"Image url {image_url} is malformed; "
 
-    split = parsed_url.path.rstrip().rsplit(":", 1)
+    split = image_url.split("/", 1)
+    host = split[0]
+    path = split[1] if len(split) > 1 else None
+
+    if host == "":
+        raise serializers.ValidationError(
+            _(base_message + "no host name found")
+        )
+
+    # Yes, validators returns (not throws) an exception if the argument doesn't
+    # pass muster (it returns True otherwise).  Consequently we have to check
+    # the class of the return to know what happened and if it's not validators'
+    # validation exception raise whatever the heck it is.
+    validity = validators.hostname(host)
+    if isinstance(validity, Exception):
+        if not isinstance(validity, validators.ValidationError):
+            raise
+        raise serializers.ValidationError(
+            _(base_message + f"invalid host name: '{host}'")
+        )
+
+    if (path is None) or (path == ""):
+        raise serializers.ValidationError(
+            _(base_message + "no image path found")
+        )
+
+    split = path.rsplit(":", 1)
     name = split[0]
     tag = split[1] if (len(split) > 1) else None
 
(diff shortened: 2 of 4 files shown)


jshimkus-rh:
Got agreement from @matburt on using the OCI regexes.

zkayyali812 (approved):

tests/integration/api/test_decision_environment.py
```diff
@@ -97,6 +97,101 @@ def test_create_decision_environment(
...
+            "tag does not match OCI standard",
+        ),
+        (
+            status.HTTP_400_BAD_REQUEST,
+            "registry.com/group/bad^img1",
+            "name does not match OCI standard",
+        ),
+        (
+            status.HTTP_400_BAD_REQUEST,
+            "registry.com/group/img1:bad^tag",
```
mkanoor, line 147: @jshimkus-rh Can this include the port number. registry.com:8443/group1/img1?

src/aap_eda/core/validators.py
```diff
@@ -57,6 +59,49 @@ def check_if_de_exists(decision_environment_id: int) -> int:
 
 
 def check_if_de_valid(image_url: str, eda_credential_id: int):
+    parsed_url = urllib.parse.urlparse(image_url)
+    base_message = f"Image url {image_url} is malformed; "
```
mkanoor, line 63: @jshimkus-rh This pattern of using gettext will help us down the road https://github.com/ansible/eda-server/blob/aaee25008fdd8f5f5bdf8e2c416039615cf004e4/src/aap_eda/core/validators.py#L137

tests/integration/api/test_decision_environment.py
jshimkus-rh: Good point.  I thought the name was the totality but it excludes the host:port.  Gonna have to make some changes.

pyproject.toml
```diff
@@ -57,6 +57,7 @@ xxhash = "3.4.*"
 pyjwt = { version = "2.7.*", extras = ["crypto"] }
 ecdsa = "0.18.*"
 distro = "^1.9.0"
+validators = "^0.34.0"
```
Alex-Izquierdo, line 60: Can you confirm if productization is aware of this new dependency and we are using the desired version constraint?

src/aap_eda/core/validators.py
```diff
@@ -57,6 +59,49 @@ def check_if_de_exists(decision_environment_id: int) -> int:
...
+    if parsed_url.scheme:
```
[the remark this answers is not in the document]
jshimkus-rh: I'm concerned about the inclusion of the '@sha256:...' in this example. It doesn't appear that anything beyond a single occurrence matching [a-zA-Z0-9_][a-zA-Z0-9._-]{0,127} is valid as a tag (the OCI REST interface does allow for queries involving such). 

I'll have to check when I have better access, but if someone can point to a definitive statement on the topic before then that would be great.

pyproject.toml
jshimkus-rh: Good point.  I meant to check but forgot.  Anyhow, I reached out to PDE and added this request: [AAP-35017](https://issues.redhat.com/browse/AAP-35017)

src/aap_eda/core/validators.py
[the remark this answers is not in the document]
jshimkus-rh: A tag with a digest is rejected when creating an image.  And while you can specify multiple tags the result is each tag is applied to a separate copy of the image.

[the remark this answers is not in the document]
mkanoor: @jshimkus-rh The sha256 in the image url is a valid syntax. Its supported by podman-pull as well as podman-py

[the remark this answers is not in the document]
jshimkus-rh: OK.  Any others?  
My plan is then to split the tag off as being bounded by `:<tag>[@<everything-else>]`, validating the tag and passing everything else through.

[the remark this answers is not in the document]
mkanoor: @jshimkus-rh Does the OCI spec talk about the tags in the image url format. We recently found an issue with podman-py where it can't handle port numbers in the image_url. If port numbers are not supposed to be defined in the image_url it might help us with the parsing.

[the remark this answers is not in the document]
jshimkus-rh: Docker describes what constitutes a valid specification here: https://docs.docker.com/reference/cli/docker/image/tag/. This includes a link to the OCI distribution spec whose section (beginning here https://github.com/opencontainers/distribution-spec/blob/8376368dd8aadc33bf6c88a8b765df90287bb5c8/spec.md?plain=1#L147) talks about the same thing though less straightforwardly. 
The OCI spec provides regexes for both the <path/to/image> and <tag> which comport with Docker's specification. It also, rather tangentially, talks about <host>[:port] (another case where Docker is clearer on the topic). Regardless, both agree that the composite specification is <host>[:port]/<path/to/image>[:<tag>]. If podman code isn't handling that then it's not standard compliant.

jshimkus-rh:
@ansible/eda-maintainers 
PDE has added the validators package. Other than possibly needing to update the .toml and .lock files everything is set. Would like to hit the ground running on Nov 11 with getting this in. Review please.

Alex-Izquierdo (dismissed):

zkayyali812 (dismissed):

src/aap_eda/core/validators.py
```diff
@@ -57,19 +60,95 @@ def check_if_de_exists(decision_environment_id: int) -> int:
...
+    # a slash: <host>[:port]/<path>[:tag].
+    #
+    # https://github.com/opencontainers/distribution-spec/blob/8376368dd8aadc33bf6c88a8b765df90287bb5c8/spec.md?plain=1#L155 # noqa: E501
+    #
+    # We split the image url on the first slash into the host and path.  The
+    # path is further split into a name and tag on the rightmost colon.
+    #
+    # THe host portion is validated using the validators package and the path
+    # and tag are validated using the OCI regexes for each.
+    base_message = f"Image url {image_url} is malformed; "
```
mkanoor, line 74: @jshimkus-rh When using gettext  we should be using %s instead of f strings for messages. That data goes to the translators.

src/aap_eda/core/validators.py
```diff
@@ -57,19 +60,95 @@ def check_if_de_exists(decision_environment_id: int) -> int:
...
+
+    split = image_url.split("/", 1)
+    host = split[0]
+    path = split[1] if len(split) > 1 else None
+
+    if host == "":
+        raise serializers.ValidationError(
+            _(base_message + "no host name found")
```
mkanoor, line 82: @jshimkus-rh Can this be changed to
```
   _("Image URL %(image_url)s is malformed, no host name found") % {'image_url': image_url}
```

src/aap_eda/core/validators.py
```diff
@@ -57,19 +60,95 @@ def check_if_de_exists(decision_environment_id: int) -> int:
...
+    # Yes, validators returns (not throws) an exception if the argument doesn't
+    # pass muster (it returns True otherwise).  Consequently we have to check
+    # the class of the return to know what happened and if it's not validators'
+    # validation exception raise whatever the heck it is.
+    validity = validators.hostname(host)
+    if isinstance(validity, Exception):
+        if not isinstance(validity, validators.ValidationError):
+            raise
+        raise serializers.ValidationError(
+            _(base_message + f"invalid host name: '{host}'")
```
mkanoor, line 94: @jshimkus-rh Can this be changed to
```
   _("Image URL %(image_url)s is malformed, invalid host name %(host)s") % {'image_url': image_url, 'host': host}
```

src/aap_eda/core/validators.py
```diff
@@ -57,19 +60,95 @@ def check_if_de_exists(decision_environment_id: int) -> int:
...
+        )
+
+    if (path is None) or (path == ""):
+        raise serializers.ValidationError(
+            _(base_message + "no image path found")
```
mkanoor, line 99: @jshimkus-rh Can this be changed to
```
   _("Image URL %(image_url)s is malformed, no image path found") % {'image_url': image_url}
```

src/aap_eda/core/validators.py
```diff
@@ -57,19 +60,95 @@ def check_if_de_exists(decision_environment_id: int) -> int:
...
+    # is passed without validation.
+    tag = split[1] if (len(split) > 1) else None
+    tag = tag if tag is None else tag.split("@", 1)[0]
+
+    if not re.fullmatch(
+        r"[[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*",  # noqa: E501
+        name,
+    ):
+        raise serializers.ValidationError(
+            _(base_message + f"'{name}' does not match OCI name standard")
```
mkanoor, line 114: @jshimkus-rh Can this be changed to
```
   _("Image URL %(image_url)s is malformed, does not match OCI name standard") % {'image_url': image_url}
```

src/aap_eda/core/validators.py
```diff
@@ -57,19 +60,95 @@ def check_if_de_exists(decision_environment_id: int) -> int:
...
+        )
+
+    if (tag is not None) and (
+        not re.fullmatch(r"[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}", tag)
+    ):
+        raise serializers.ValidationError(
+            _(base_message + f"'{tag}' does not match OCI tag standard")
```
mkanoor, line 121: @jshimkus-rh Can this be changed to
```
   _("Image URL %(image_url)s is malformed, tag %(tag)s does not match OCI tag standard") % {'image_url': image_url, 'tag': tag}
```

src/aap_eda/core/validators.py
```diff
@@ -57,19 +60,95 @@ def check_if_de_exists(decision_environment_id: int) -> int:
...
     credential = get_credential_if_exists(eda_credential_id)
     inputs = yaml.safe_load(credential.inputs.get_secret_value())
-    host = inputs.get("host")
+    credential_host = inputs.get("host")
 
-    if not host:
+    if not credential_host:
         raise serializers.ValidationError(
-            f"Credential {credential.name} needs to have host information"
+            _(f"Credential {credential.name} needs to have host information")
```
mkanoor, line 130: @jshimkus-rh Can this be changed to
```
   _("Credential %(name)s needs to have host information") % {'name': credential.name}
```

src/aap_eda/core/validators.py
```diff
@@ -57,19 +60,95 @@ def check_if_de_exists(decision_environment_id: int) -> int:
...
+        if parsed_credential_host.scheme:
+            parsed_host = ":".join(
+                [parsed_credential_host.scheme, parsed_host]
+            )
+
+    if host != parsed_host:
+        msg = _(
             f"DecisionEnvironment image url: {image_url} does "
-            f"not match with the credential host: {host}"
+            f"not match with the credential host: {credential_host}"
```
mkanoor, line 151: @jshimkus-rh Can this be changed to
```
   _("DecisionEnvironment image url %(image_url)s does not match with the credential host %(host)s") % {'image_url': image_url, 'host': credential_host}
```

src/aap_eda/core/validators.py
jshimkus-rh: @mkanoor updated for get text

mkanoor (approved):

zkayyali812 (approved):
lgtm,
Appreciate all of the detailed test cases added here as well

PR #1129, opened 2024-10-29
fix: continue to sync rulebooks when previous sync failed

bzwei:
It helps to reset properly the import state and error message.

AAP-34744: Sync a previously failed project importing falsely changes the state to completed

### DIFF
--- a/src/aap_eda/services/project/imports.py
+++ b/src/aap_eda/services/project/imports.py
@@ -105,7 +105,14 @@ class ProjectImportService:
     @_project_import_wrapper
     def sync_project(self, project: models.Project) -> None:
         with self._clone_and_process(project) as (repo_dir, git_hash):
-            if project.git_hash == git_hash:
+            # At this point project.import_state and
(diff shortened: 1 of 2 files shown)


src/aap_eda/services/project/imports.py
```diff
@@ -56,8 +56,10 @@ def _project_import_wrapper(
 ):
     @wraps(func)
     def wrapper(self: ProjectImportService, project: models.Project):
-        project.import_state = models.Project.ImportState.RUNNING
-        project.save(update_fields=["import_state"])
+        # The view already changed the state to running before importing
```
Alex-Izquierdo, line 59: Where this happen? The view set the status to pending as far as I can see. And I think that's correct. We should not set the running before the import task begins.

bzwei: sorry. I misread the code.

Alex-Izquierdo (approved):

hsong-rh (approved):

PR #1132, opened 2024-11-01
Add EDA_ANALYTICS Feature Flag

zkayyali812:
Adds a feature flag for the EDA Analytics feature. Ensures flag can be toggled on/off via settings. Can also be toggled off via django management command - https://cfpb.github.io/django-flags/management_commands/

https://issues.redhat.com/browse/AAP-34975


src/aap_eda/api/urls.py
```diff
@@ -109,6 +107,18 @@
...
+    flagged_path(
+        "EDA_ANALYTICS",
+        "settings/system/",
+        views.SettingView.as_view(),
+        name="setting-system",
+        state=True,
+    ),
+]
+
+eda_v1_urls = eda_v1_urls + flagged_urls
```
bzwei, line 120: consider `eda_v1_urls.extend()`

src/aap_eda/core/utils/logging_utils.py
```diff
@@ -25,3 +27,9 @@ def get_organization_name_from_data(data):
             else None
         )
     return org_name
+
+
+def feature_flag_analytics_disabled_logger():
```
bzwei: this function name suggests you're returning a logger, but it actually writes a message to the log.

src/aap_eda/core/utils/logging_utils.py
```diff
@@ -25,3 +27,9 @@ def get_organization_name_from_data(data):
...
+    logger = logging.getLogger("aap_eda.analytics")
+    logger.error("EDA_ANALYTICS is not enabled.")
+    return
```
bzwei: no need `return` statement

src/aap_eda/settings/default.py
```diff
@@ -198,6 +199,15 @@ def _get_boolean(name: str, default=False) -> bool:
     "aap_eda.core",
 ]
 
+FLAGS = {
+    "EDA_ANALYTICS": [
+        {
+            "condition": "boolean",
+            "value": settings.get("EDA_FEATURE_FLAG_EDA_ANALYTICS", False),
```
bzwei, line 206: `settings.get("FEATURE_FLAG_EDA_ANALYTICS", False)`

tests/integration/analytics/test_gather_analytics.py
```diff
@@ -124,6 +126,7 @@ def test_gather_analytics_invalid_settings(
     ],
 )
 @pytest.mark.django_db
+@override_settings(FLAGS={"EDA_ANALYTICS": [("boolean", True)]})
```
bzwei, line 129: Is there a better way to override only flag `EDA_ANALYTICS`, consider FLAGS in the future can expand?

src/aap_eda/settings/default.py
zkayyali812: Great catch :)

tests/integration/analytics/test_gather_analytics.py
zkayyali812: There are a few ways to do this via unit testing effectively that I've found - 
1. override-settings: https://github.com/cfpb/django-flags/blob/main/flags/tests/test_urls.py#L132
   a. Can also do this via this [syntax](https://github.com/ansible/eda-server/pull/1132/files#diff-e3e0c48a81c019581f7b22fe7d99bb2c712d7460877558335201d913daa193d6R37-R39)
2. FlagCreate: https://github.com/cfpb/django-flags/blob/main/flags/tests/test_models.py#L8-L10
3. Via django-admin cli command - https://github.com/ansible/eda-server/pull/1132/files#diff-e3e0c48a81c019581f7b22fe7d99bb2c712d7460877558335201d913daa193d6R37-R39
   a. https://cfpb.github.io/django-flags/management_commands/
   
 Not entirely sure which is the best currently, but open to perspectives here

zkayyali812: After some further thought, I can probably define a generic function to set feature flags, to be used in unit test, which can take in an array of feature flags to set their state automatically via environment variable directly as well

src/aap_eda/analytics/collector.py
```diff
@@ -87,9 +91,13 @@ def _save_last_gather(self) -> None:
...
-    collector = AnalyticsCollector(
-        collector_module=analytics_collectors,
-        collection_type=collection_type,
-        logger=logger,
-    )
-    return collector.gather(since=since, until=until)
+    if not flag_enabled("EDA_ANALYTICS"):
+        feature_flag_analytics_disabled_logger()
+        return
+    else:
```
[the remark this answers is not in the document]
zkayyali812: Fixed

src/aap_eda/api/urls.py
zkayyali812: Fixed

src/aap_eda/core/utils/logging_utils.py
zkayyali812: Function is removed now since it introduced a bug, reworked how this is done

src/aap_eda/core/utils/logging_utils.py
zkayyali812: function now removed

src/aap_eda/settings/default.py
zkayyali812: Fixed

src/aap_eda/analytics/collector.py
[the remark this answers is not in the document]
mkanoor: @zkayyali812 Would it make sense to use a decorator that can be used for any function where the check needs to be performed. 
e.g.
@check_feature_enabled("EDA_ANALYTICS")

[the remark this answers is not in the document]
zkayyali812: Yes, I think this is a great suggestion. I will try to implement this. They have a decorator as part of the package, but this seems to be limited to working with views. Implementing our own decorator here, I think is more sufficient.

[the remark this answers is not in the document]
zkayyali812: After trying to implement this, there is not a great way of doing this without reworking either how imports are done in the unit test, or reworking the class from what I have seen.

Currently the [url](https://github.com/ansible/eda-server/pull/1132/files#diff-2970751c57bcb2f4906a6353ed7d8f16c396995e16c41b35bf6c84a780c5b4cfR111) to access this view is feature flagged

tests/integration/analytics/test_gather_analytics.py
zkayyali812: Tried to implement this, but ran into some hurdles here. Wondering if we have a way to refresh the settings after updating environmental variables?
For unit testing, I expect way may be able to override the flags directly via override settings for the tests that would require this, but open to other perspectives here

zkayyali812:
@bzwei I re added the `settings.get("FEATURE_FLAG_EDA_ANALYTICS", True)` here. 

After further discussion - 
1. The goal of this PR is to show how we can toggle a feature flag at install time only
    1. Toggling the state of the feature flag during runtime can come later, likely leveraging the aap-preferences effort.
2. Once this code path has been in existence for a while, this feature flag will eventually be removed entirely, but the **INSIGHTS_TRACKING_STATE** variable will remain, so both should co-exist for now

tests/integration/analytics/test_gather_analytics.py
Alex-Izquierdo: @zkayyali812 
We can have something like [this](https://github.com/ansible/ansible-rulebook/blob/7c1840b17b6934c804b5cf47fc62489eb1cb2780/tests/e2e/conftest.py#L8), but modiying direclty `os.environ` Anyway this might be tricky and we need to take care of ensuring it doesn't not affect to preloaded entities like dynaconf settings, so I guess it won't work in some cases. 
I think it is fine as it is, we can rely on dynaconf for envvars handling.

src/aap_eda/settings/default.py
```diff
@@ -198,6 +199,16 @@ def _get_boolean(name: str, default=False) -> bool:
     "aap_eda.core",
 ]
 
+FLAGS = {
+    "EDA_ANALYTICS": [
+        {
+            "condition": "boolean",
+            "value": settings.get("FEATURE_FLAG_EDA_ANALYTICS", True),
```
Alex-Izquierdo, line 206: We already have the automatic "EDA" prefix, so the environment variable will be `EDA_FEATURE_FLAG_EDA_ANALYTICS` which seems a bit redundant. 
There is no chance of collision, so I would just use `FEATURE_FLAG_ANALYTICS`

Also, I don't know if we already agreed a name pattern, but I would like to use the "ENABLE" term to be clearer.
I propose `ENABLE_FEATURE` prefix, "ENABLE_FEATURE_[feature name]" with "ENABLE" I feel we state clearly what means the value and that we expect a boolean. 

 If FEATURE_FLAG has been already decided for the whole platform, it's fine then.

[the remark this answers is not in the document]
zkayyali812: Currently, yes, per our last architecture call, but this is only going into our feature branch.
We can quickly switch the default if we'd like as well

zkayyali812: It has not yet been decided for the whole platform, but I think that is a great note.

I can change this variable name to - `ENABLE_FEATURE_EDA_ANALYTICS`.

Since this is intended to be eventually toggled on/off at the platform level, and it will not affect other components analytics, I think it is good to keep the `EDA` part in there currently, even though this may be redundant.

Alex-Izquierdo (dismissed):

bzwei:
@zkayyali812 You should guard https://github.com/ansible/eda-server/blob/feature/analytics/src/aap_eda/settings/default.py#L521 with the feature flag too. If the flag is off, the RQ_STARTUP_JOBS list is empty.

src/aap_eda/settings/default.py
zkayyali812: I ended up renaming this too - `FEATURE_EDA_ANALYTICS`, with the exception of the `EDA_` leading prefix, I figured it may be good to not set a standard on starting this variable with `ENABLE` for now.
I think it would be good to try to ensure all feature flag variables start with `FEATURE_` (ignoring the potential component leading prefix)

zkayyali812:
@bzwei Great catch! I just updated to include that. I cant use the `flag_enabled` function here in the `default.py` file as a note, due to us not being allowed to check the feature flag state until the app registry is ready. To get around this I checked the setting directly here.

bzwei (approved):

Alex-Izquierdo (approved):

PR #1133, opened 2024-11-07
feat: Add a system command to purge log records

hsong-rh:
https://issues.redhat.com/browse/AAP-30684

Add a system command `purge_log_records` to clean up unneeded logs. After log records are purged, one single log record with content like "All log records are purged at 2024-11-06 22:31:16." is added.

**Usage:**

```
usage: aap-eda-manage purge_log_records [-h] --activation-ids ACTIVATION-IDS [--all]

Purge logs from rulebook process

options:
  -h, --help            show this help message and exit
  --activation-ids ACTIVATION-IDS
                        Specify the activation ids which you want to clear their records
                        (e.g., ActivationID1 ActivationID2)
  --all                 Purge all instances's logs from the activations

```


Alex-Izquierdo (commented):
Where can I see the requirements for this command if they exists? 

As it is now I'm not sure how much useful it can be.
I have the following concerns:
- Only supports PK's, should support as well names?
- What is the goal to purge the logs of just the latest instance?
- Should we support a filter by date? e.g.:  purge all the instances before X date. Should it be granular per instance? (to be able to keep the most recent lines of the instance logs)
- Should we purge the logs of a running instance?

hsong-rh:
@Alex-Izquierdo Good questions, I agree that cleanup based on timestamp makes more sense than PKs, even than names. The active instances should be untouched, I will fix it.

mkanoor:
@hsong-rh The log for a running activation can be truncated, we should have an option to delete the log files for the running activation. Maybe also add a log line when we truncate so that the only log line is
" Log truncated on YYYYMMDDTHH:MM:SSZ" 
So if the user sees this message they would know the reason the logs are empty because someone decided to truncate it.
If there is only 1 activation instance running and it has been running for days/months the user might want to truncate the logs for the running Activation. There is nothing in the logs that ansible-rulebook needs also we store the last time we read the logs in [here](https://github.com/ansible/eda-server/blob/16a34da1e03b7b9fcaef5f0b4c58898c4dba9eef/src/aap_eda/core/models/rulebook_process.py#L72)

bzwei:
Here are my concerns:
* Why the cutoff date applies to activation creation time? For an activation that was create a month ago but its latest instance completes a fews seconds ago, its logs will be purged by this command. But it may be desired to keep the logs for debugging.
* Why exclude activation instances that are still running? An instance can be running continuously very long and accumulate large amount of logs. It can be beneficial to purge older logs but keep only recent ones. I would recommend the cutoff date being the log creation time, not activation creation time.
* Why the user must select activation ids or names? I would select all activations if neither is specified.

src/aap_eda/core/management/commands/purge_log_records.py
```diff
@@ -0,0 +1,158 @@
...
+            log_timestamp__lt=int(ts.timestamp()),
+        )
+        if not log_records.exists():
+            return
+
+        log_records.delete()
+        dt = f"{datetime.now().strftime('%Y-%m-%d %H:%M:%S')}"
+
+        return models.RulebookProcessLog(
+            log=f"All log records are purged at {dt}.",
```
bzwei, line 106: consider `f"All log records older than {cutoff_date} are purged at {dt}."`

src/aap_eda/core/management/commands/purge_log_records.py
```diff
@@ -0,0 +1,158 @@
...
+            activation_instance_id=instance.id,
+            log_timestamp=int(datetime.now().timestamp()),
```
bzwei, line 108: Should not log_timestamp be the cutoff time or the log_timstamp of the last deleted log? This way when sorted by log_timestamp we can see the messages in logic order.

src/aap_eda/core/management/commands/purge_log_records.py
```diff
@@ -0,0 +1,160 @@
...
+        input_ids = [int(id) for id in options.get("activation-ids") or []]
+        input_names = options.get("activation-names") or []
+        cutoff_date = options.get("date")
+
+        try:
+            if not bool(input_ids) and not bool(input_names):
+                activation_ids = [
+                    activation.id
+                    for activation in models.Activation.objects.all()
+                ]
```
Alex-Izquierdo, line 143:
```suggestion
                activation_ids = list(models.Activation.objects.values_list("id", flat=True))
```

src/aap_eda/core/management/commands/purge_log_records.py
```diff
@@ -0,0 +1,160 @@
...
+            else:
+                activation_ids = list(
+                    set(
+                        self.get_activation_ids(input_ids)
+                        + self.get_activation_ids(input_names)
+                    )
+                )
+
+            if not bool(activation_ids):
+                logger.info("No log records are purged.")
```
Alex-Izquierdo, line 153:
```suggestion
                logger.info("No records have been found for purging.")
```

src/aap_eda/core/management/commands/purge_log_records.py
```diff
@@ -0,0 +1,160 @@
...
+                    activation.id
+                    for activation in models.Activation.objects.all()
+                ]
+            else:
+                activation_ids = list(
+                    set(
+                        self.get_activation_ids(input_ids)
+                        + self.get_activation_ids(input_names)
+                    )
+                )
```
Alex-Izquierdo, line 150: Do we need this logic and the function `get_activation_ids`?
I believe we can just write a query that returns activations by both, ids or names, something like:
```
models.Activation.objects.filter(
    Q(id__in=ids) | Q(name=name)
).distinct()
```

Also, why do we have to run a query to get the ids and later run another query to get the rulebook_processes?I think believe we can just pass the queryse, then the ORM will run only one query.

bzwei (approved):

src/aap_eda/core/management/commands/purge_log_records.py
```diff
@@ -0,0 +1,139 @@
...
+            log=(
+                f"All log records older than {cutoff_date} are purged at {dt}."
+            ),
+            activation_instance_id=instance.id,
+            log_timestamp=int(ts.timestamp()),
+        )
+
+    @transaction.atomic
+    def handle(self, *args, **options):
+        input_ids = [int(id) for id in options.get("activation-ids") or []]
```
Alex-Izquierdo, line 129: I think you don't want to have a nested list and this is:
```suggestion
        input_ids = [int(id) for id in options.get("activation-ids")] or []
```
Anyway I think there is no need to call `int(id)` since argparse validates and cast the value for you, so it should be enough with 
```suggestion
        input_ids = options.get("activation-ids") or []
```
as you already do with `input_names`

src/aap_eda/core/management/commands/purge_log_records.py
```diff
@@ -0,0 +1,139 @@
...
+        new_instance_logs = []
+        if not bool(ids) and not bool(names):
+            instances = models.RulebookProcess.objects.all()
+        else:
+            instances = models.RulebookProcess.objects.filter(
+                Q(activation__id__in=ids) | Q(activation__name__in=names),
+            )
+
+        if not instances.exists():
+            logger.info("No records has been found for purging.")
```
Alex-Izquierdo, line 81: This should not be a logger:
```suggestion
            self.stdout.write(
                self.style.SUCCESS("No records has been found for purging.")
            )
```

src/aap_eda/core/management/commands/purge_log_records.py
```diff
@@ -0,0 +1,139 @@
...
+                self.clean_instance_logs(instance, cutoff_date)
+            )
+
+        new_instance_logs = [
+            item for item in new_instance_logs if item is not None
+        ]
+        if len(new_instance_logs) > 0:
+            models.RulebookProcessLog.objects.bulk_create(new_instance_logs)
+
+            logger.info(f"Log records older than {cutoff_date} are purged.")
```
Alex-Izquierdo, line 95: This should not be a logger:
```suggestion
            self.stdout.write(
                self.style.SUCCESS("Log records older than %s are purged." % cutoff_date)
            )
```

Alex-Izquierdo (approved):

bzwei (approved):

PR #1136, opened 2024-11-14
fix: retrieve project using web browser

jshimkus-rh:
Utilize project instance as input to project retrieve read serializer.
https://issues.redhat.com/browse/AAP-29995
---
The current EDA code when retrieving a project passes a dictionary as the `instance` of the instantiated read serializer.  In the context of a web browser request (`rest_framework. renderers.BrowsableAPIRenderer`) this results in the issues as described in https://issues.redhat.com/browse/AAP-29995.

One of these issues is the logging of
```
ansible_base.rbac.api.permissions INFO Object-permission check called with non-object <class 'rest_framework.utils.serializer_helpers.ReturnDict'> 
```
which occurs when constructing a `PATCH` section of the page.

The other is
```
 File "/usr/lib/python3.11/site-packages/aap_eda/api/serializers/project.py", line 31, in get_proxy
    if not obj.proxy:
           ^^^^^^^^^
AttributeError: 'ReturnDict' object has no attribute 'proxy'
```
which results from `proxy` being declared as a `SerializerFieldMethod` and the read serializer using a mixin to provide the `get_proxy` functionality which expects its input to be a project instance.  

To address both issues this change modifies the project retrieve and project read serializer code to utilize a project instance as input to the read serializer.

### DIFF
--- a/src/aap_eda/api/views/project.py
+++ b/src/aap_eda/api/views/project.py
@@ -171,38 +171,19 @@ class ProjectViewSet(
         },
     )
     def retrieve(self, request, pk):
-        project = super().retrieve(request, pk)
-        project.data["eda_credential"] = (
-            models.EdaCredential.objects.get(
-                pk=project.data["eda_credential_id"]
-            )
-            if project.data["eda_credential_id"]
-            else None
-        )
-        project.data["signature_validation_credential"] = (
-            models.EdaCredential.objects.get(
-                pk=project.data["signature_validation_credential_id"]
-            )
-            if project.data["signature_validation_credential_id"]
-            else None
-        )
-        project.data["organization"] = (
-            models.Organization.objects.get(pk=project.data["organization_id"])
-            if project.data["organization_id"]
-            else None
-        )
+        project = self.get_object()
 
         logger.info(
             logging_utils.generate_simple_audit_log(
                 "Read",
                 resource_name,
-                project.data["name"],
-                project.data["id"],
-                project.data["organization"].name,
+                project.name,
+                project.id,
+                project.organization.name,
             )
         )
 
-        return Response(serializers.ProjectReadSerializer(project.data).data)
+        return Response(serializers.ProjectReadSerializer(project).data)
 
     @extend_schema(
         description="Partial update of a project",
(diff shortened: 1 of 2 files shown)


zkayyali812 (commented):
This looks good to me.
Wondering if it is possible to add a test here to effectively catch this? 
Or if, since according to the issue, that this only happens from the browser, can this not effectively be unit-tested?

jshimkus-rh:
> This looks good to me. Wondering if it is possible to add a test here to effectively catch this? Or if, since according to the issue, that this only happens from the browser, can this not effectively be unit-tested?

As far as I know we don't have support for test browser-based access.
Out of curiosity I tried it with lynx and it failed as expected and worked with this change.  So possibly if we are inclined to take on such an effort, which wouldn't be a knock it out in 5 minutes sort of thing.

src/aap_eda/api/views/project.py
```diff
@@ -171,38 +171,19 @@ def create(self, request):
         },
     )
     def retrieve(self, request, pk):
-        project = super().retrieve(request, pk)
```
Alex-Izquierdo, line 174: mixins.RetrieveModelMixin is no longer necessary in ProjectViewSet after this change, right?

Alex-Izquierdo (approved):

zkayyali812 (approved):
lgtm!

If more bugs come up like this, perhaps it may be worth adding lynx functionality to unit tests to ensure we dont regress, but I am not aware of any others

jshimkus-rh:
> lgtm!
> 
> If more bugs come up like this, perhaps it may be worth adding lynx functionality to unit tests to ensure we dont regress, but I am not aware of any others

Decision Environment has the same `dictionary as instance` approach, but the nature of what is being serialized doesn't result in an exception.  It does have the `non-object class` log messages though:
```
2024-11-15 10:16:58 2024-11-15 15:16:58,418 ansible_base.rbac.api.permissions INFO     Object-permission check called with non-object <class 'rest_framework.utils.serializer_helpers.ReturnDict'> for PATCH
2024-11-15 10:16:58 2024-11-15 15:16:58,422 ansible_base.rbac.api.permissions INFO     Object-permission check called with non-object <class 'rest_framework.utils.serializer_helpers.ReturnDict'> for DELETE
2024-11-15 10:16:58 2024-11-15 15:16:58,422 ansible_base.rbac.api.permissions INFO     Object-permission check called with non-object <class 'rest_framework.utils.serializer_helpers.ReturnDict'> for OPTIONS 
```

PR #1137, opened 2024-11-14
fix: disallow updating credential type on credential patch endpoint

Dostonbek1:
This PR disallows changing Credential Type when updating an Eda Credential. 

## Testing

1. Create a new Eda Credential with type Container Registry
2. Try patching the credential with a payload of just the `credential_type_id` corresponding to the RHAAP Credential Type
3. Assert Credential Type has not been changed

**NOTE**: The UI already disables updating Credential Type; this issue exists only on the API. 

JIRA: [AAP-34968](https://issues.redhat.com/browse/AAP-34968)

### DIFF
--- a/src/aap_eda/api/serializers/eda_credential.py
+++ b/src/aap_eda/api/serializers/eda_credential.py
@@ -115,23 +115,12 @@ class EdaCredentialCreateSerializer(serializers.ModelSerializer):
     inputs = serializers.JSONField()
 
     def validate(self, data):
-        credential_type_id = data.get("credential_type_id")
-        if credential_type_id:
-            credential_type = models.CredentialType.objects.get(
-                id=credential_type_id
-            )
-        else:
-            # for update
-            credential_type = self.instance.credential_type
+        credential_type = models.CredentialType.objects.get(
+            id=data.get("credential_type_id")
+        )
 
         inputs = data.get("inputs", {})
-
-        # allow emtpy inputs during updating
-        if self.partial and not bool(inputs):
-            return data
-
         errors = validate_inputs(credential_type.inputs, inputs)
-
         if bool(errors):
             raise serializers.ValidationError(errors)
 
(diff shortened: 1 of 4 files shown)


src/aap_eda/api/serializers/eda_credential.py
```diff
@@ -115,23 +115,45 @@ class EdaCredentialCreateSerializer(serializers.ModelSerializer):
...
+            "organization_id",
+        ]
+
+
+class EdaCredentialUpdateSerializer(serializers.ModelSerializer):
+    organization_id = serializers.IntegerField(
+        required=True,
+        allow_null=False,
+        validators=[validators.check_if_organization_exists],
+    )
```
jshimkus-rh, line 145: For an update wouldn't this be optional?

hsong-rh: we shouldn't allow updating the organization either, right?

Dostonbek1: That's a good point but it still somehow works when we don't pass `organization_id` field in patch payload, I can't tell exactly how..

Dostonbek1:
```
..when using a PATCH request, fields that are required on the model or serializer are not necessarily required in the payload. This behavior is due to the partial nature of the PATCH method, which is designed to update only the fields provided in the request rather than requiring all fields.
```
Regardless, I agree that it can be confusing to set it as required. I think I could even take out the custom definition of this field since it is already in the model.

jshimkus-rh: @hsong-rh  
> we shouldn't allow updating the organization either, right?

Moving a credential from one org to another is different than changing the kind of credential it is.  Though it does raise the question of what impact it might have on RBAC?  If people who are in the old org lose the access they had at least they aren't getting access they shouldn't, but is that (and only that)  what would happen?  Might they get access to an org to which they don't belong?  Possibly dependent on what constellation of accesses are checked; if both being in the org and having the credential then that would appear fine, they'd be denied. If simply having the credential for an org that could be an issue.

jshimkus-rh: @matburt Any insights you can share or point us to someone who can?  I think there was someone from one of the staff meetings who was going to think about some of these issues but I don't remember who and no one reached out to me. 🤷‍♂️

jshimkus-rh: > @matburt Any insights you can share or point us to someone who can?  I think there was someone from one of the staff meetings who was going to think about some of these issues but I don't remember who and no one reached out to me. 🤷‍♂️ 

Actually, I think that was about the Default organization and not credential-related.

src/aap_eda/api/serializers/eda_credential.py
```diff
@@ -115,23 +115,45 @@ class EdaCredentialCreateSerializer(serializers.ModelSerializer):
...
-        if credential_type_id:
-            credential_type = models.CredentialType.objects.get(
-                id=credential_type_id
-            )
-        else:
-            # for update
-            credential_type = self.instance.credential_type
+        credential_type = models.CredentialType.objects.get(
+            id=data.get("credential_type_id")
+        )
```
Alex-Izquierdo, line 120: Is the field validator executed before this function? I wonder if this raises DoesNotExist if the id does not exist.

Dostonbek1: I'm positive that field validators are executed first before the `validate` function and since `credential_type_id` is required, we can assume we'll always have a value for it.

src/aap_eda/api/serializers/eda_credential.py
Dostonbek1: @jshimkus-rh @hsong-rh Controller allows updating the organization for a credential; I'd say it is probably better to align with it.

zkayyali812 (approved):

PR #1140, opened 2024-11-18
fix: add validation of host input for registry credential

jshimkus-rh:
https://issues.redhat.com/browse/AAP-34969


zkayyali812 (approved):

Alex-Izquierdo (approved):

src/aap_eda/api/serializers/eda_credential.py
```diff
@@ -120,7 +120,11 @@ def validate(self, data):
         )
 
         inputs = data.get("inputs", {})
-        errors = validate_inputs(credential_type.inputs, inputs)
+        errors = validate_inputs(
+            credential_type,
```
zkayyali812, line 124: Should we be doing type checking here and in other cases of this function, like we are doing in the utils function [here](https://github.com/ansible/eda-server/pull/1140/files#diff-06294198ff5faf34e2f2dd8135ddd4e7853e3ccb815fb8bfc2b4e14d7570625cR91)?

jshimkus-rh: @zkayyali812 Not following you.  Please be more specific.

zkayyali812: Will reach out in slack shortly

zkayyali812 (approved):

PR #1143, opened 2024-11-19
fix: retrieve decision environment using web browser

jshimkus-rh:
https://issues.redhat.com/browse/AAP-36030

The current EDA code when retrieving a decision environment passes a dictionary as the instance of the instantiated read serializer. In the context of a web browser request (rest_framework. renderers.BrowsableAPIRenderer) this results in logging of the following...

```
ansible_base.rbac.api.permissions INFO Object-permission check called with non-object <class 'rest_framework.utils.serializer_helpers.ReturnDict'> 
```

To address this issue this change modifies the decision environment retrieve and decision environment read serializer code to utilize a decision environment instance as input to the read serializer.

### DIFF
--- a/src/aap_eda/api/serializers/decision_environment.py
+++ b/src/aap_eda/api/serializers/decision_environment.py
@@ -97,28 +97,22 @@ class DecisionEnvironmentReadSerializer(serializers.ModelSerializer):
     def to_representation(self, decision_environment):
         eda_credential = (
             EdaCredentialRefSerializer(
-                decision_environment["eda_credential"]
+                decision_environment.eda_credential
             ).data
-            if decision_environment["eda_credential"]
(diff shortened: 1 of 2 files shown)


jshimkus-rh:
> @jshimkus-rh @Dostonbek1 The changes to get the related objects was done as part of RBAC PR #710. I think if we directly fetch the objects and not check if the organization and related objects are accessible are we bypassing RBAC. Can we test if RBAC would still work after these changes.

Trying to retrieve as a non-privileged user does not succeed.

Dostonbek1 (approved):
Approving as per internal discussion.

mkanoor (approved):

Alex-Izquierdo (approved):

PR #1145, opened 2024-11-20
feat: collect event streams used by running activations

hsong-rh:
https://issues.redhat.com/browse/AAP-36177

Collect those event streams that are currently used by activations

### DIFF
--- a/src/aap_eda/analytics/analytics_collectors.py
+++ b/src/aap_eda/analytics/analytics_collectors.py
@@ -479,6 +479,57 @@ def event_streams_by_activation_table(
 
 
 @register(
+    "event_streams_by_running_activations_table",
+    "1.0",
+    format="csv",
+    description="Data on event_streams used by each running activation",
+)
+def event_streams_by_running_activations_table(
+    since: datetime, full_path: str, until: datetime, **kwargs
+):
+    running_states = [
+        ActivationStatus.RUNNING,
+        ActivationStatus.STARTING,
+        ActivationStatus.PENDING,
+    ]
+    activations = (
+        models.Activation.objects.filter(
+            Q(created_at__gt=since, created_at__lte=until)
+            | Q(modified_at__gt=since, modified_at__lte=until)
+        )
+        .filter(status__in=running_states)
+        .distinct()
+    )
+
+    event_streams = models.EventStream.objects.none()
+    for activation in activations:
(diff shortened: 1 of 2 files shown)


tests/integration/analytics/test_analytics_collectors.py
```diff
@@ -575,31 +624,73 @@ def test_event_streams_table_by_activation_collector(
...
+@pytest.mark.django_db
+def test_event_streams_table_by_running_activations_collector(
+    activations_with_different_status: list[models.Activation],
+    default_event_streams: list[models.EventStream],
+):
+    activations_with_different_status[0].event_streams.add(
+        default_event_streams[0]
+    )
+    activations_with_different_status[1].event_streams.add(
+        default_event_streams[1]
```
bzwei, line 641: Can you add a case that an event stream is used by multiple activations? Can it be deduplicated?

src/aap_eda/analytics/analytics_collectors.py
```diff
@@ -478,6 +478,57 @@ def event_streams_by_activation_table(
...
+    event_streams = models.EventStream.objects.none()
+    for activation in activations:
+        event_streams |= activation.event_streams.all()
+
+    if not bool(event_streams):
+        return
+
+    event_streams = event_streams.annotate(
+        event_stream_id=F("id"),
+        activation_id=F("activations__id"),
```
Alex-Izquierdo, line 513: This query will duplicate the event stream record per activation related if there are multiple activations using the same event stream. Is this the desired outcome or maybe we want only one record per event stream with a list of the activation ids consuming it? 

I don't have context about this data will be consumed, I'm not saying it is wrong as it is, I'm just making sure.

tests/integration/analytics/test_analytics_collectors.py
hsong-rh: Yes, even streams used by different activations are treated as other records in the collection table.

src/aap_eda/analytics/analytics_collectors.py
hsong-rh: Yeah, the analytics team can easily use the flat list with other data to combine the final result.

bzwei (approved):

dhaustein (approved):

PR #1147, opened 2024-11-21
feat: prevent ansible, eda keys to be used in extra vars

mkanoor:
Prevent users from using reserved attribute names when creating extra vars which are
  1. eda
  2. ansible

  * eda is used when we create file name for injectors e.g "{{ eda.filename.certfile }}"
  * ansible is used by ansible rulebook to send events to extra vars Controller e.g ansible.eda.events

We check the extra_vars in Activation as well as extra_vars in Credential Type injectors for reserved names

### DIFF
--- a/src/aap_eda/core/utils/credentials.py
+++ b/src/aap_eda/core/utils/credentials.py
@@ -32,6 +32,8 @@ PROTECTED_PASSPHRASE_ERROR = (
     "The key is passphrase protected, please provide passphrase."
 )
 
+RESERVED_EXTRA_VAR_KEYS = {"ansible", "eda"}
+
 
 class InjectorMissingKeyException(Exception):
     pass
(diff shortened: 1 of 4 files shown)


hsong-rh (dismissed):

src/aap_eda/core/utils/credentials.py
```diff
@@ -446,3 +455,18 @@ def _validate_file_template_key(key: str, key_names: list[str]) -> None:
...
                 )
+
+
+def check_reserved_keys_in_extra_vars(data: dict[str, any]) -> None:
+    for key in data.keys():
+        if key in RESERVED_EXTRA_VAR_KEYS:
+            raise ValidationError(
+                _(
+                    "Extra vars key %(key)s cannot be one of these "
+                    "reserved keys %(reserved)s"
```
jshimkus-rh, line 466: I find it preferable to visually offset both offending key and reserved keys to make them stand out as the salient portions of the message.  E.g.,
`"Extra vars key '%(key)s' cannot be one of these "
                     "reserved keys '%(reserved)s'"`

src/aap_eda/core/utils/credentials.py
```diff
@@ -446,3 +455,18 @@ def _validate_file_template_key(key: str, key_names: list[str]) -> None:
...
+                )
+                % {
+                    "key": key,
+                    "reserved": ",".join(sorted((RESERVED_EXTRA_VAR_KEYS))),
```
jshimkus-rh, line 470: Not certain of applicability to internationalization but `", ".join` (note the space after ',') is in keeping with English.

jshimkus-rh (approved):

PR #1149, opened 2024-11-22
feat: address a corner case when the file contents is a dict

mkanoor:
Users can put all kinds of text data for file contents, if they put in a python dictionary as a string, jinja substitution converts it into a dict, we need to convert it back to a string.


mkanoor:
@Dostonbek1 This is the fix to the issue that you had observed when you added a dict as file contents

Alex-Izquierdo (approved):

hsong-rh (approved):

Dostonbek1 (approved):

PR #1150, opened 2024-11-22
feat: enable custom podman socket timeout

Alex-Izquierdo:
Adds PODMAN_SOCKET_TIMEOUT to the system settings to allow custom timeout for podman socket.
Includes a little refactor of the `get_podman_client` and extends tests.

### DIFF
--- a/src/aap_eda/services/activation/engine/podman.py
+++ b/src/aap_eda/services/activation/engine/podman.py
@@ -37,25 +37,28 @@ from .common import (
 LOGGER = logging.getLogger(__name__)
 
 
+def _get_podman_socket_url() -> str:
+    if settings.PODMAN_SOCKET_URL:
+        return settings.PODMAN_SOCKET_URL
+    if os.getuid() == 0:
+        return "unix:///run/podman/podman.sock"
+    xdg_runtime_dir = os.getenv("XDG_RUNTIME_DIR", f"/run/user/{os.getuid()}")
+    return f"unix://{xdg_runtime_dir}/podman/podman.sock"
+
+
 def get_podman_client() -> PodmanClient:
     """Podman client factory."""
+    params = {}
+    if settings.PODMAN_SOCKET_TIMEOUT:
+        params["timeout"] = settings.PODMAN_SOCKET_TIMEOUT
+    podman_url = _get_podman_socket_url()
+    params["base_url"] = podman_url
+    LOGGER.info(f"Using podman socket: {podman_url}")
     try:
-        podman_url = settings.PODMAN_SOCKET_URL
-        if podman_url:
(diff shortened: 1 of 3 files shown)


src/aap_eda/services/activation/engine/podman.py
```diff
@@ -37,25 +37,28 @@
...
+        return "unix:///run/podman/podman.sock"
+    xdg_runtime_dir = os.getenv("XDG_RUNTIME_DIR", f"/run/user/{os.getuid()}")
+    return f"unix://{xdg_runtime_dir}/podman/podman.sock"
+
+
 def get_podman_client() -> PodmanClient:
     """Podman client factory."""
+    params = {}
+    if settings.PODMAN_SOCKET_TIMEOUT:
+        params["timeout"] = settings.PODMAN_SOCKET_TIMEOUT
```
jshimkus-rh, line 53: As `settings.PODMAN_SOCKET_TIMEOUT` is always set and `None` is consistent with `PodmanClient` instantiation there's no need for a conditional on specifying `timeout`.

Alex-Izquierdo: I don't rely enough in podman client, so I want to exclude it explicitly, just in case the behavior changes in the future.

jshimkus-rh: Ok.  That raises the question of the difference, if any, between None and zero (0)?  Whether in practice it makes a difference the code makes a distinction at initialization:

https://github.com/containers/podman-py/blob/2fbd32fe6139763d0a75aed61761f10d7fad7d10/podman/api/client.py#L142

jshimkus-rh: OK.  Zero is bad.  If I mod the code to pass it along we get the following:
```
2024-12-13 16:15:07,901 aap_eda.services.activation.engine.podman INFO   Using podman socket: unix://socket_url
E        ValueError: Attempted to set connect timeout to 0, but the timeout cannot be set to a value less than or equal to 0.
.venv/lib/python3.11/site-packages/urllib3/util/timeout.py:154: ValueError
```
I handling that in settings.py rather than podman.py.
e.g.,
```
if PODMAN_SOCKET_TIMEOUT == "0":
    PODMAN_SOCKET_TIMEOUT = None
```

that way the internals of EDA (should other parts ever care) can assume the appropriateness of the value rather having to special case it.

Alex-Izquierdo: Ok, @jshimkus-rh changed as per suggestion, please review again. Thanks

jshimkus-rh (approved):

jshimkus-rh (approved):

Dostonbek1 (approved):

PR #1151, opened 2024-11-22
feat: handle binary files

mkanoor:
For Kerberos we need to support a keytab file which is a binary file and its used for service accounts during authentication with Kerberos.

This uses the format: binary_base64 in the field definition to let the UI know that the expected file is supposed to be binary and needs to be sent up as base64
You can create a credential with the following schema

Inputs configuration

```
fields:
  - id: keytab
    type: string
    label: Kerberos Keytab file
    format: binary_base64
    secret: true
    help_text: Please select a Kerberos Keytab file
    multiline: true
```

Injector Configuration

```
file:
  template.keytab_file: '{{ keytab }}'
```

No changes are needed to ansible-rulebook it already handles binary files.


src/aap_eda/core/utils/strings.py
```diff
@@ -41,6 +41,21 @@ def _render_string_or_return_value(value: Any, context: Dict) -> Any:
     return value
 
 
+def extract_variables(template_string):
```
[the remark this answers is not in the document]
mkanoor: Fixed

src/aap_eda/wsapi/consumers.py
```diff
@@ -498,20 +497,32 @@ def _get_url(self, message: ActionMessage, inputs: dict) -> str:
...
+        file_messages = []
         for eda_credential in activation.eda_credentials.all():
             inputs = yaml.safe_load(eda_credential.inputs.get_secret_value())
             injectors = eda_credential.credential_type.injectors
+            binary_fields = []
+            for field in eda_credential.credential_type.inputs.get(
+                "fields", []
+            ):
+                if field.get("format") in BINARY_FORMATS:
+                    binary_fields.append(field["id"])
```
hsong-rh, line 511: Is `id` a required field? it may be better to use `field.get("id")` to avoid potential KeyError

src/aap_eda/wsapi/consumers.py
```diff
@@ -575,3 +586,26 @@ def get_vault_password_and_id(
             vault_inputs = yaml.safe_load(vault_inputs.get_secret_value())
             return vault_inputs["vault_password"], vault_inputs["vault_id"]
         return None, None
+
+    @staticmethod
+    def get_file_content_message(
+        template: str, binary_fields: list[str], value: str, inputs: dict
+    ) -> FileContentMessage:
+        binary_file = False
```
hsong-rh, line 594: `binary_file = any(attr in binary_fields for attr in extract_variables(value))` may be better

src/aap_eda/wsapi/consumers.py
mkanoor: @hsong-rh In a Credential Schema a id is a required field. We can have an empty schema but not a field with a missing id.

hsong-rh (approved):

bzwei (approved):

PR #1152, opened 2024-11-22
Demonstrate DAB version generation problem

AlanCoding:
I mean to show here how it generates a commit hash _from today_ but gives it the `2024.4.25` version.

Also see how it adds feature flags as a dependency, which is super duper recent stuff from @john-westcott-iv, and I create this to show how something is broken on a more basic level of packaging for DAB (in addition to the release job, https://issues.redhat.com/browse/AAP-36569).

I also want the DAB folks to review the other dependency changes, because they seem to raise some red flags.

Produced with

```
poetry lock --no-update
```

If it's a problem with my computer, then fine, but the more I looked at this, I don't think it is.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -621,7 +621,7 @@ bcrypt = ["bcrypt"]
 
 [[package]]
 name = "django-ansible-base"
-version = "2024.11.8.0.dev7+gfe4ee752"
+version = "2024.4.25"
 description = "A Django app used by ansible services"
 optional = false
 python-versions = ">=3.9"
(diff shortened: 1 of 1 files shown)


poetry.lock
```diff
@@ -631,31 +631,28 @@ develop = false
 [package.dependencies]
 channels = {version = "*", optional = true, markers = "extra == \"channel-auth\""}
 cryptography = "*"
-Django = ">=4.2.16,<4.3.0"
+Django = ">=4.2.5,<4.3.0"
```
AlanCoding, line 634: This appears to not be picking up (reverting) https://github.com/ansible/django-ansible-base/pull/605 from Oct 22.

AlanCoding:
Related https://github.com/ansible/django-ansible-base/pull/657

poetry.lock
```diff
@@ -631,31 +631,28 @@ develop = false
...
 redis-client = ["django-redis", "redis"]
-resource-registry = ["asgiref", "pyjwt", "requests", "urllib3"]
 testing = ["cryptography", "pytest", "pytest-django"]
 
 [package.source]
 type = "git"
 url = "https://github.com/ansible/django-ansible-base.git"
 reference = "devel"
-resolved_reference = "3dc1dca437ee594211ec33bfb4c3fa1c6d39d11a"
+resolved_reference = "7c71771fbd2ca54dcb400ab12fe0f7a4ee655eed"
```
AlanCoding, line 655: This is a recent commit from 4 days ago, so that part seemed to work right.

https://github.com/ansible/django-ansible-base/commit/7c71771fbd2ca54dcb400ab12fe0f7a4ee655eed

poetry.lock
```diff
@@ -631,31 +631,28 @@ develop = false
...
-all = ["asgiref", "channels", "cryptography", "django-auth-ldap", "django-oauth-toolkit (<2.4.0)", "django-redis", "drf-spectacular", "pyjwt", "pyjwt", "pyrad", "pytest", "pytest-django", "python-ldap", "python3-saml", "redis", "requests", "requests", "social-auth-app-django (==5.4.1)", "tabulate", "tacacs_plus", "urllib3", "xmlsec (==1.3.13)"]
+all = ["channels", "cryptography", "django-auth-ldap", "django-redis", "drf-spectacular", "pyjwt", "pytest", "pytest-django", "python-ldap", "python3-saml", "redis", "requests", "social-auth-app-django", "tabulate", "tacacs_plus", "xmlsec (==1.3.13)"]
 api-documentation = ["drf-spectacular"]
-authentication = ["django-auth-ldap", "pyrad", "python-ldap", "python3-saml", "social-auth-app-django (==5.4.1)", "tabulate", "tacacs_plus", "xmlsec (==1.3.13)"]
+authentication = ["django-auth-ldap", "python-ldap", "python3-saml", "social-auth-app-django", "tabulate", "tacacs_plus", "xmlsec (==1.3.13)"]
 channel-auth = ["channels"]
 jwt-consumer = ["pyjwt", "requests"]
-oauth2-provider = ["django-oauth-toolkit (<2.4.0)"]
 redis-client = ["django-redis", "redis"]
-resource-registry = ["asgiref", "pyjwt", "requests", "urllib3"]
```
AlanCoding, line 651: This change is not at all correct. These requirements were added ~1 month ago, and for some reason it is trying to take them away.

poetry.lock
```diff
@@ -621,7 +621,7 @@ bcrypt = ["bcrypt"]
 
 [[package]]
 name = "django-ansible-base"
-version = "2024.11.8.0.dev7+gfe4ee752"
+version = "2024.4.25"
```
AlanCoding, line 624: version is obviously wrong.

PR #1154, opened 2024-11-26
Pick up help_text changes from DAB

AlanCoding:
The PR https://github.com/ansible/django-ansible-base/pull/638 modified `help_text` on fields of the common models, and eda-server was using those abstract classes as bases for some of their models.

This causes checks to fail on other PRs, so I expect this needs to get in.


AlanCoding:
Ok, so it seems we need to actually update DAB to hit this. Right now I have https://github.com/ansible/django-ansible-base/pull/657 open so we can potentially do this with a tag. I can try to do it with the current pyproject.toml, but I don't think it'll work.

Dostonbek1 (approved):
I was able to run this locally and confirm the help texts for name fields of organization and team models. 
Organization:
image

Team:
image

Not sure how to manually check for the other fields though but I'd assume they should be there considering it works for name fields.

PR #1156, opened 2024-12-03
feat: AAP-36635 - change payload_content_type to aap-event-driven-ansible

hsong-rh:
https://issues.redhat.com/browse/AAP-36635

Update to use new content type `aap-event-driven-ansible` to upload data


jshimkus-rh:
Might be time to consider breaking up parts of the compose into different files and using `include:`.  Though a quick search looks like maybe podman (or podman-compose) may not support it at present.

jshimkus-rh (approved):

hsong-rh:
> Might be time to consider breaking up parts of the compose into different files and using `include:`. Though a quick search looks like maybe podman (or podman-compose) may not support it at present.
> 
Yeah, we can plan it in the future

bzwei (approved):

Dostonbek1 (approved):

PR #1155, opened 2024-12-02
fix: rework project sync on empty repo

bzwei:
The project import status will be completed when the repo is empty or does not contain the required folder. The status will be failed if the repo is uninitiazlied. This error comes from the git checkout operation.

AAP-35777: Rework project sync on empty git repo

PR #1158, opened 2024-12-04
feat: [AAP-35878] add copy endpoint for EDA Credentials

Dostonbek1:
This PR adds a new endpoint `/eda-credentials/<id>/copy` to create a new copy of the given EDA Credential.

image

JIRA: [AAP-35878](https://issues.redhat.com/browse/AAP-35878)

### DIFF
--- a/src/aap_eda/api/views/eda_credential.py
+++ b/src/aap_eda/api/views/eda_credential.py
@@ -25,6 +25,7 @@ from drf_spectacular.utils import (
     extend_schema,
 )
 from rest_framework import mixins, status, viewsets
+from rest_framework.decorators import action
 from rest_framework.filters import BaseFilterBackend
 from rest_framework.response import Response
 
@@ -32,6 +33,7 @@ from aap_eda.api import exceptions, filters, serializers
 from aap_eda.api.serializers.eda_credential import get_references
 from aap_eda.core import models
 from aap_eda.core.utils.credentials import (
+    build_copy_post_data,
     inputs_to_store,
     inputs_to_store_dict,
 )
@@ -277,3 +279,43 @@ class EdaCredentialViewSet(
             )
         self.perform_destroy(eda_credential)
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+    @extend_schema(
+        description="Copy an EDA credential.",
+        request=None,
+        responses={
+            status.HTTP_201_CREATED: OpenApiResponse(
+                serializers.EdaCredentialSerializer,
+                description="Return the copied EDA credential.",
+            ),
+            status.HTTP_404_NOT_FOUND: OpenApiResponse(
+                None, description="EDA credential not found."
+            ),
+        },
+    )
+    @action(methods=["post"], detail=True)
+    def copy(self, request, pk):
+        eda_credential = self.get_object()
+        post_data = build_copy_post_data(eda_credential)
+
+        serializer = serializers.EdaCredentialCreateSerializer(data=post_data)
+        serializer.is_valid(raise_exception=True)
+        serializer.validated_data["inputs"] = inputs_to_store(
+            serializer.validated_data["inputs"]
+        )
+        with transaction.atomic():
+            response = serializer.create(serializer.validated_data)
+            check_related_permissions(
+                request.user,
+                serializer.Meta.model,
+                {},
+                model_to_dict(response),
+            )
+            RoleDefinition.objects.give_creator_permissions(
+                request.user, response
+            )
+
+        return Response(
+            serializers.EdaCredentialSerializer(response).data,
+            status=status.HTTP_201_CREATED,
+        )
--- a/src/aap_eda/api/views/eda_credential.py
+++ b/src/aap_eda/api/views/eda_credential.py
@@ -134,30 +134,7 @@ class EdaCredentialViewSet(
         },
     )
     def create(self, request):
-        serializer = serializers.EdaCredentialCreateSerializer(
-            data=request.data
-        )
-
-        serializer.is_valid(raise_exception=True)
-        serializer.validated_data["inputs"] = inputs_to_store(
-            serializer.validated_data["inputs"]
-        )
-        with transaction.atomic():
-            response = serializer.create(serializer.validated_data)
-            check_related_permissions(
-                request.user,
-                serializer.Meta.model,
-                {},
-                model_to_dict(response),
-            )
-            RoleDefinition.objects.give_creator_permissions(
-                request.user, response
-            )
-
-        return Response(
-            serializers.EdaCredentialSerializer(response).data,
-            status=status.HTTP_201_CREATED,
-        )
+        return self._create_eda_credential(request, request.data)
 
     @extend_schema(
         description="List all EDA credentials",
@@ -299,8 +276,11 @@ class EdaCredentialViewSet(
     def copy(self, request, pk):
         eda_credential = self.get_object()
         post_data = build_copy_post_data(eda_credential)
+        return self._create_eda_credential(request, post_data)
 
-        serializer = serializers.EdaCredentialCreateSerializer(data=post_data)
+    def _create_eda_credential(self, request, data):
+        """Create a new credential object given payload data."""
+        serializer = serializers.EdaCredentialCreateSerializer(data=data)
         serializer.is_valid(raise_exception=True)
         serializer.validated_data["inputs"] = inputs_to_store(
             serializer.validated_data["inputs"]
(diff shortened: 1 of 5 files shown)


src/aap_eda/api/views/eda_credential.py
```diff
@@ -277,3 +279,46 @@ def destroy(self, request, *args, **kwargs):
...
+
+    @extend_schema(
+        description="Copy an EDA credential.",
+        request=None,
+        responses={
+            status.HTTP_201_CREATED: OpenApiResponse(
+                serializers.EdaCredentialSerializer,
+                description="Return the copied EDA credential.",
+            ),
+            status.HTTP_403_FORBIDDEN: OpenApiResponse(
```
jshimkus-rh, line 291: I assume a specific test for this wasn't added based on RBAC enforcement for `self.get_object()` and the serializer create? 
Given our usage of `self.get_object()` in other locations where we don't list a 403 as a possible response are we being inconsistent?

The above was my original comment but this popped into my head:
Can we have a situation where a user could get an object but not be able to create a new one?

zkayyali812:
Im curious why the e2e tests are failing here? 
This is likely unrelated to the PR, but I thought this was fixed in the changes related to this PR - https://github.com/ansible/eda-server/pull/1155

src/aap_eda/api/views/eda_credential.py
Dostonbek1: I'm sure we are being inconsistent in other places where we don't list 403 as one of possible responses. But this is mainly for API schema documentation, thus it has little to no effect on the functionality. 

As for the second use case you mentioned, it is possible and can easily happen when a user has object-level role on one credential but has no org-level credential role because this copy endpoint requires `add_edacredential` permission to succeed.

jshimkus-rh: > I'm sure we are being inconsistent in other places where we don't list 403 as one of possible responses. But this is mainly for API schema documentation, thus it has little to no effect on the functionality.

Wasn't concerned about functionality but the consistency of what we're doing and presenting.  We should be consistent.  The reasonable (to me) way of doing that would be to only include a 403 response in the schema if it is a response EDA itself can create and not for transitive responses.  

Providing consistency would be a separate effort from this change, though you could remove the occurrence here.  Regardless, I'm still going to approve this change. 😄

jshimkus-rh (approved):

src/aap_eda/api/views/eda_credential.py
Dostonbek1: Removed the 403 response as per your feedback to stay consistent.

Dostonbek1:
@AlanCoding I'd like your input too here for RBAC action, is it consistent with the controller to require users `add` permission for copy action or do you guys have custom permission `copy` for this use case? It makes sense to me to require `add` permission since copying a credential is basically creating a new object.

AlanCoding:
I think this is a case where we might want to add `rbac_action='add'` to the `@action` (IIRC) decorator.

What you have is a good approach, and is how I would have done it. I actually think it will work completely, security-wise. But the rbac_action will connect it for general view-specific, non-object-related, considerations, which probably only matters for OPTIONS.

tests/integration/api/test_eda_credential.py
```diff
@@ -1002,3 +1002,29 @@ def test_custom_credential_type_empty_inputs(
...
+        f"{api_url_v1}/eda-credentials/{default_registry_credential.id}/copy/"
+    )
+    assert response.status_code == status.HTTP_201_CREATED
+    new_credential = response.data
+    assert new_credential["id"] != default_registry_credential.id
+    assert (
+        new_credential["credential_type"]["id"]
+        == default_registry_credential.credential_type.id
+    )
+    assert default_registry_credential.name in new_credential["name"]
```
hsong-rh, line 1022: can you add a check to ensure the encrypted field is decryptable and its value is the same as the original credential?

src/aap_eda/core/utils/credentials.py
```diff
@@ -470,3 +478,30 @@ def check_reserved_keys_in_extra_vars(data: dict[str, any]) -> None:
                     "reserved": ", ".join(sorted((RESERVED_EXTRA_VAR_KEYS))),
                 }
             )
+
+
+def build_copy_post_data(eda_credential: "models.EdaCredential") -> dict:
```
hsong-rh, line 483: Why do you use string, instead of class object as the type hint? Seems like the latter is more recommended

src/aap_eda/core/utils/credentials.py
```diff
@@ -14,15 +14,23 @@
...
 import yaml
 from django.core.exceptions import ValidationError
+from django.forms import model_to_dict
 from django.utils.translation import gettext_lazy as _
 from jinja2.nativetypes import NativeTemplate
 
 from aap_eda.api.constants import EDA_SERVER_VAULT_LABEL
+from aap_eda.core.utils.crypto.base import SecretValue
+
+if typing.TYPE_CHECKING:
```
hsong-rh, line 31: is this avoiding circular importing?

src/aap_eda/core/utils/credentials.py
Dostonbek1: This is a smart way of avoiding circular import error (which happens when typing is used without quotation marks) that @jshimkus-rh came up with, IIRC.

src/aap_eda/core/utils/credentials.py
Dostonbek1: AFAIK, yes.

src/aap_eda/core/utils/credentials.py
jshimkus-rh: Credit where credit is due. 🫡 @Alex-Izquierdo is the one who told me about it.

jshimkus-rh: I used it to avoid rearranging code for the simple inclusion of a new parameter to an existing method.  At some point we may want to take that on, but at least it's still confined to the one file.

jshimkus-rh (dismissed):

jshimkus-rh (approved):

hsong-rh (changes requested):
@Dostonbek1 can you double check your copy logic is in parity with AAP controller? In the future we will migrate the credential service

AlanCoding (approved):

hsong-rh (dismissed):

zkayyali812 (approved):
/lgtm
After further thought our e2e tests are likely failing here because we need to update `bugfix/injectors` branch with latest main.
Recommend we do this after this PR gets in

src/aap_eda/api/views/eda_credential.py
```diff
@@ -277,3 +281,43 @@ def destroy(self, request, *args, **kwargs):
...
+                None, description="EDA credential not found."
+            ),
+        },
+    )
+    @action(methods=["post"], detail=True, rbac_action=Action.READ)
+    def copy(self, request, pk):
+        eda_credential = self.get_object()
+        post_data = build_copy_post_data(eda_credential)
+
+        serializer = serializers.EdaCredentialCreateSerializer(data=post_data)
```
bzwei, line 303: after the post_data is prepared, the rest is the same as create. `copy` and `create` should share the same logic.

src/aap_eda/core/enums.py
```diff
@@ -56,9 +56,9 @@ class ResourceType(DjangoStrEnum):
 
 
 class Action(DjangoStrEnum):
-    CREATE = "create"
-    READ = "read"
-    UPDATE = "update"
+    CREATE = "add"
```
bzwei, line 59: I am surprised that you can change the value here. Are `create` and `add` both predefined or it can be any word? Same for the rest of changes.

Dostonbek1: Left side of the assignment are the conventional CRUD action name, but we have to change the values to adapt to the DAB RBAC naming convention. Also, note that these enums are only used for RBAC on API level and has no effect on db, etc.

bzwei: for curiosity, how does it work before the change? unless DAB supports both words.

bzwei (approved):

src/aap_eda/core/enums.py
Dostonbek1: @bzwei We were not using any of the CRUD before except for READ, but I'm a bit confused why it was working for READ; my guess is that in all the use cases we had `detail=False` which somehow ignores or doesn't make use of `rbac_action` but setting `detail=True` and giving rbac action that doesn't match the ones in DAB result in error like `RuntimeError: The permission read_edacredential is not valid for model edacredential` so I had to make these changes above.

jshimkus-rh (dismissed):

hsong-rh (approved):

jshimkus-rh (approved):

jshimkus-rh (approved):

bzwei (approved):

PR #1160, opened 2024-12-04
fix: [AAP-36058] correct the default log_timestamp of RulebookProcessLog records

hsong-rh:
https://issues.redhat.com/browse/AAP-36058

Correct the default value of log_timestamp with the right timestamp.

### DIFF
--- a/src/aap_eda/services/activation/db_log_handler.py
+++ b/src/aap_eda/services/activation/db_log_handler.py
@@ -51,6 +51,10 @@ class DBLogger(LogHandler):
         ):
             self.flush()
 
+        # set default value of log_timestamp
+        if log_timestamp == 0:
+            log_timestamp = int(datetime.now().timestamp())
+
         if not isinstance(lines, list):
             lines = [lines]
 
--- a/tests/integration/services/activation/engine/test_podman.py
+++ b/tests/integration/services/activation/engine/test_podman.py
@@ -174,6 +174,8 @@ def test_engine_start(
 
     engine.client.containers.run.assert_called_once()
     assert models.RulebookProcessLog.objects.count() == 4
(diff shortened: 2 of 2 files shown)


bzwei:
@hsong-rh from the example in https://issues.redhat.com/browse/AAP-36058 it shows the messages already have timestamp embedded, can't we extract them and set them in the right column?

hsong-rh:
> @hsong-rh from the example in https://issues.redhat.com/browse/AAP-36058 it shows the messages already have timestamp embedded, can't we extract them and set them in the right column?

Yes, the sample log records have the log_timestamp column, but they were set with the wrong default value of 0. This PR fixes the issue, so all newly generated log records will have the right log_timestamp value. We will leave those existing records untouched because we missed their created timestamps.

bzwei:
> > @hsong-rh from the example in https://issues.redhat.com/browse/AAP-36058 it shows the messages already have timestamp embedded, can't we extract them and set them in the right column?
> 
> Yes, the sample log records have the log_timestamp column, but they were set with the wrong default value of 0. This PR fixes the issue, so all newly generated log records will have the right log_timestamp value. We will leave those existing records untouched because we missed their created timestamps.

I am not requesting to amend existing records. Based on the sample logs if there is already a timestamp in the message body, should we extract it and set it to the right column instead of generating a timestamp based on now()?

Dostonbek1 (approved):

hsong-rh:
> I am not requesting to amend existing records. Based on the sample logs if there is already a timestamp in the message body, should we extract it and set it to the right column instead of generating a timestamp based on now()?

No, the `log_timestamp` is the only timestamp column in RulebookProcessLog records

PR #1168, opened 2024-12-16
feat: [AAP-22893] add filtering by activation instance name to audit rules list view

Dostonbek1:
- pin DAB to [2024.12.13](https://github.com/ansible/django-ansible-base/releases/tag/2024.12.13) release tag
- change to using `rest-filters` app from DAB as a default filter backend for filtering and ordering
- add unit test for filtering audit rules by activation instance name

NOTE: Using `rest-filters` changes the keyword `ordering` for sorting to `order` and `order_by`. Also, this change will transfer the freedom to the UI to build the query param with the correct lookup expression, e.g. `/audit-rules/?name__contains=Test`.

JIRA: [AAP-22893](https://issues.redhat.com/browse/AAP-22893)

### DIFF
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -194,6 +194,7 @@ INSTALLED_APPS = [
     "ansible_base.rbac",
     "ansible_base.resource_registry",
     "ansible_base.jwt_consumer",
+    "ansible_base.rest_filters",
     # Local apps
     "aap_eda.api",
     "aap_eda.core",
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -563,6 +563,7 @@ SPECTACULAR_SETTINGS = {
     "PREPROCESSING_HOOKS": [
         "aap_eda.api.openapi.preprocess_filter_api_routes"
     ],
+    "GENERIC_ADDITIONAL_PROPERTIES": "bool",
 }
 
 # ---------------------------------------------------------
--- a/tests/integration/api/test_rulebook.py
+++ b/tests/integration/api/test_rulebook.py
@@ -302,7 +302,7 @@ def test_list_audit_rules_ordering(
     ordering_field,
 ):
     response = admin_client.get(
-        f"{api_url_v1}/audit-rules/?ordering={ordering_field}"
+        f"{api_url_v1}/audit-rules/?order={ordering_field}"
     )
     assert response.status_code == status.HTTP_200_OK
     audit_rules = response.data["results"]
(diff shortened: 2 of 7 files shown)


tests/integration/api/test_rulebook.py
```diff
@@ -426,7 +454,7 @@ def test_list_actions_from_audit_rule_ordering(
     ordering_field,
 ):
     response = admin_client.get(
-        f"{api_url_v1}/audit-rules/{audit_rule_2.id}/actions/?ordering="
+        f"{api_url_v1}/audit-rules/{audit_rule_2.id}/actions/?order="
```
mkanoor, line 457: @Dostonbek1 Will this query difference have to be coordinated with UI @lgalis? Is the EDA UI using ordering?

tests/integration/api/test_rulebook.py
```diff
@@ -352,6 +352,34 @@ def test_list_audit_rules_ordering_activation_name(
...
 
+@pytest.mark.django_db
+def test_list_audit_rules_filter_activation_instance_name(
+    audit_rule_1: models.AuditRule,
+    audit_rule_2: models.AuditRule,
+    admin_client: APIClient,
+):
+    filter_name = audit_rule_1.activation_instance.name
+    response = admin_client.get(
+        f"{api_url_v1}/audit-rules/?activation_instance__name={filter_name}"
```
mkanoor, line 363: @Dostonbek1 Since we have removed the Audit Action and Audit Event filter should we have a test case for that. Also doesn't the DAB rest filter support AND and OR combinations. should we have a test based on that. To pick a specific rule name under an Activation Instance. 
```
f"{api_url_v1}/audit-rules/?activation_instance__name={filter_name} AND rule_name=R2" something like this.

Comment (Dostonbek1) [reply tests/integration/api/test_rulebook.py:363]: @mkanoor We are removing our custom filtering from Audit Action and Audit Event but the way we are implementing DAB filters is that it automatically applies to all views that do not have filter class specified, so Audit Action and Audit Event are using DAB filters right now with these changes. And we do have existing tests for these; check tests starting on line 418 on this file.

**Review by Alex-Izquierdo — APPROVED** [review]

File: src/aap_eda/settings/default.py [hunk 3]
```diff
@@ -562,6 +563,7 @@ def get_rq_queues() -> dict:
     "PREPROCESSING_HOOKS": [
         "aap_eda.api.openapi.preprocess_filter_api_routes"
     ],
+    "GENERIC_ADDITIONAL_PROPERTIES": "bool",
```
mkanoor, line 566: @Dostonbek1 Since this is going to change the generated  should we we change the SPECTACULAR_SETTINGS.VERSION to be the same as the EDA SERVER Version or will that break the generated OpenAPI client.

zkayyali812 (approved):

src/aap_eda/settings/default.py
mkanoor: We can do this in a separate PR and set the version to be 
```
VERSION: get_eda_version(),
```

mkanoor (approved):

PR #1170, opened 2024-12-18
fix: use package version for app version in openapi

Alex-Izquierdo:
app version in openapi spec is hardcoded and outdated
This should be taken dynamically from the package version using the existing "get_eda_version" helper. 
Jira: https://issues.redhat.com/browse/AAP-37834

### DIFF
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -99,8 +99,8 @@ import dynaconf
 from django.core.exceptions import ImproperlyConfigured
 from split_settings.tools import include
 
+from aap_eda import utils
 from aap_eda.core.enums import RulebookProcessLogLevel
-from aap_eda.utils import str_to_bool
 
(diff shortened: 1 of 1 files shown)


src/aap_eda/settings/default.py
```diff
@@ -139,7 +139,7 @@ def _get_secret_key() -> str:
 def _get_boolean(name: str, default=False) -> bool:
     value = settings.get(name, default)
     if isinstance(value, str):
-        value = str_to_bool(value)
+        value = utils.str_to_bool(value)
```
mkanoor, line 142: @Alex-Izquierdo Is this related to this PR?

Alex-Izquierdo: I just import utils package for readability and usability, instead of importing functions one by one.

Dostonbek1 (approved):

PR #1172, opened 2025-01-02
fix: [AAP-38148] import utils module properly to solve AttributeError in get_default_organization

hsong-rh:
Fix the AttributeError raised in `models.utils.get_default_organization()`

```
     models.utils.get_default_organization()

    AttributeError: module 'aap_eda.core.models' has no attribute 'utils'
```
https://issues.redhat.com/browse/AAP-38148

### DIFF
--- a/src/aap_eda/core/management/commands/create_initial_data.py
+++ b/src/aap_eda/core/management/commands/create_initial_data.py
@@ -24,6 +24,7 @@ from django.db import transaction
 from django.db.models import Q
 
 from aap_eda.core import enums, models
+from aap_eda.core.models.utils import get_default_organization
 from aap_eda.core.tasking import enable_redis_prefix
 from aap_eda.core.utils.credentials import inputs_to_store
 
@@ -1072,7 +1073,7 @@ class Command(BaseCommand):
         if not credentials:
             return
 
-        default_organization = models.utils.get_default_organization()
+        default_organization = get_default_organization()
 
         cred_type = models.CredentialType.objects.filter(
             name=enums.DefaultCredentialType.REGISTRY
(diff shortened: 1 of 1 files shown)


kaiokmo (approved):

src/aap_eda/core/management/commands/create_initial_data.py
```diff
@@ -24,6 +24,7 @@
 from django.db.models import Q
 
 from aap_eda.core import enums, models
+from aap_eda.core.models import utils
```
mkanoor, line 27: @hsong-rh Would it be better to directly import the fuction instead of importing utils since that is such a generic name.
```
   from aap_eda.core.models.utils import get_default_organization
```

mkanoor (approved):

src/aap_eda/core/management/commands/create_initial_data.py
jshimkus-rh: @hsong-rh @mkanoor 
Not that my opinion matters (at least, anymore 😄) but... at one point I put out a post in the eda channel about avoiding "free-standing" (or whatever I called it) references.  The idea being to provide context for the origin of a reference making life somewhat easier when trying to hunt down all the usages of contents from a particular package (motivated by the effort to hunt down all redis-related usages when adding the connection retry decorator).  No one objected, but no one jumped up and down and said "Yes!" either.

Do with this what you will.

Alex-Izquierdo (approved):

src/aap_eda/core/management/commands/create_initial_data.py
Alex-Izquierdo: I would prefer to import the module instead of the function, especially when we talk about "utils", it gives more context and makes easier to track usages in the future as joe pointed out.

PR #1173, opened 2025-01-03
[dispatcher POC] Update DAB, use advisory lock from it

AlanCoding:


AlanCoding:
Never mind, was already done

PR #1174, opened 2025-01-06
feat: allow to update activations via the patch API

bzwei:
Activations must be in disabled mode. Most fields are editable.

AAP-37295: Edit an existing activation


bzwei:
e2e test failure is fixed by https://github.com/ansible/eda-qa/pull/447

src/aap_eda/api/views/activation.py
```diff
@@ -125,6 +132,53 @@ def create(self, request):
...
+                description="Invalid data to update activation."
+            ),
+            status.HTTP_409_CONFLICT: OpenApiResponse(
+                description="The activation is not allowed to be updated."
+            ),
+        },
+    )
+    def partial_update(self, request, pk):
+        activation = self.get_object()
+        if activation.is_enabled:
```
Alex-Izquierdo, line 149: Disabling is an async operation, do you think we should also check the status of the activation?

tests/integration/api/test_activation.py
```diff
@@ -1028,3 +1028,45 @@ def test_create_activation_with_skip_audit_events(
     data = response.data
     activation = models.Activation.objects.filter(id=data["id"]).first()
     assert activation.skip_audit_events
+
+
+@pytest.mark.django_db
+def test_update_activation(
```
zkayyali812, line 1034: Can we also add a unit test for return code 400 (invalid body)?

src/aap_eda/api/serializers/activation.py
```diff
@@ -473,13 +471,164 @@ def create(self, validated_data):
...
+    extra_var = serializers.CharField(
+        required=False,
+        allow_null=True,
+        allow_blank=True,
+        validators=[validators.is_extra_var_dict],
+    )
+    decision_environment_id = serializers.IntegerField(
+        validators=[validators.check_if_de_exists]
+    )
+    awx_token_id = serializers.IntegerField(
```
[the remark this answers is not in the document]
mkanoor: @Alex-Izquierdo @zkayyali812 We have a separate JIRA for removing it. Its still being used in eda-qa and atf repos we would have to remove those tests first. https://issues.redhat.com/browse/AAP-33565

tests/integration/api/test_activation_with_event_stream.py
```diff
@@ -156,20 +156,49 @@
...
+        inputs=inputs_to_store(
+            {"username": f"{name}-user", "password": "dummy-password"}
+        ),
+        organization=organization,
+    )
+
+
 def create_activation_related_data(
-    event_stream_names, with_project=True, rulesets=TEST_RULESETS
+    event_stream_names, with_project=True, rulesets=TEST_RULESETS, appendix=""
```
mkanoor, line 189: @bzwei suffix might be a better name than appendix.

bzwei:
> @bzwei Since we might be updating and restarting an Activation should we log a line here to indicate that the Activation has been updated, since if the user picks up different rulebook during an update the old Rule Audit wont make any sense. Having a log line that indicates that the Activaiton has been updated since creation will give some context to the old Rule Audits. The old Rule Audits point to an Activation Instance log and this message will tell us that the user has updated this Activation.
> 
> https://github.com/ansible/eda-server/blob/69b06ef686e3d800c88f754c4570b6fe5db5bb05/src/aap_eda/services/activation/activation_manager.py#L200

@mkanoor Here I have the log https://github.com/ansible/eda-server/pull/1174/files#diff-3aa1d8dc47ee0af34b544249cc19758b5919dbddf7057e84dc032f49f6235f33R176. Is it sufficient?

mkanoor:
@bzwei That is an audit log that doesn't come out to the activation_instance log. That audit log was added by @zkayyali812 for audit trail. We need a one liner in the rulebook process log that is visible in the UI that shows it is an updated Activation. When you look at the the "Rule Audit" in out UI and click on the Activation it opens up our Rulebook process log thats where we would like to see it has been updated.

bzwei:
> @bzwei That is an audit log that doesn't come out to the activation_instance log. That audit log was added by @zkayyali812 for audit trail. We need a one liner in the rulebook process log that is visible in the UI that shows it is an updated Activation. When you look at the the "Rule Audit" in out UI and click on the Activation it opens up our Rulebook process log thats where we would like to see it has been updated.

We do not know whether the activation has been updated unless we introduce a new field to its model. This new line in the rulebook process log will be printed in every restart.

zkayyali812 (approved):

bzwei:
The log will be added in another PR when introducing an updated_at timestamp to the activation mode.
The test failure needs to be fixed at the eda-qa repo.
This PR is good to go. Please review again.

src/aap_eda/api/serializers/activation.py
```diff
@@ -473,13 +471,168 @@ def create(self, validated_data):
...
+            "k8s_service_name",
+            "source_mappings",
+            "skip_audit_events",
+        ]
+
+    organization_id = serializers.IntegerField(
+        required=True,
+        allow_null=False,
+        validators=[validators.check_if_organization_exists],
+    )
```
hsong-rh, line 507: Please add customized error messages for `organization_id, rulebook_id` and `decision_environment_id`, similar as other serializers

bzwei: None of these fields are actually required for update.

src/aap_eda/api/views/activation.py
```diff
@@ -125,6 +132,59 @@ def create(self, request):
...
+        serializer.prepare_update(activation)
+
+        old_data = model_to_dict(activation)
+        with transaction.atomic():
+            serializer.update(activation, serializer.validated_data)
+            check_related_permissions(
+                request.user,
+                serializer.Meta.model,
+                old_data,
+                model_to_dict(activation),
```
Alex-Izquierdo, line 173: Are you not using the same data for the old_data for the new_data?

src/aap_eda/api/serializers/activation.py
```diff
@@ -473,13 +471,168 @@ def create(self, validated_data):
...
+    k8s_service_name = serializers.CharField(
+        required=False,
+        allow_null=True,
+        allow_blank=True,
+        validators=[validators.check_if_rfc_1035_compliant],
+    )
+
+    def validate(self, data):
+        _validate_credentials_and_token_and_rulebook(data=data, creating=True)
```
Alex-Izquierdo, line 542: I think the validation is a bit mess. The token is fetched twice, in the field validator and here in _validate_awx_token. We would not need the field validator. 
Similar for the rulebook, it is fetched twice. This issue affects as well to the create serializer. I think we should just remove `check_if_rulebook_exists` and do everything in `_validate_rulebook`

src/aap_eda/api/views/activation.py
```diff
@@ -125,6 +132,59 @@ def create(self, request):
...
+                description="Invalid data to update activation."
+            ),
+            status.HTTP_409_CONFLICT: OpenApiResponse(
+                description="The activation is not allowed to be updated."
+            ),
+        },
+    )
+    def partial_update(self, request, pk):
+        activation = self.get_object()
+        if activation.is_enabled or activation.status in [
```
Alex-Izquierdo, line 149: We should need here too WORKERS_OFFLINE status.

src/aap_eda/api/serializers/activation.py
hsong-rh: Are they editable? if so, we then need to validate them

src/aap_eda/api/views/activation.py
bzwei: the line `serializer.update(activation, serializer.validated_data)` changes activation before and after.

src/aap_eda/api/serializers/activation.py
bzwei: These fields are not required but cannot be null. I added customized error_messages for the null case.

bzwei:
@Alex-Izquierdo I rolled back my changes based on your comments on removing validators for atx token and rulebook_id. Your reasoning for combining the validations in one place to avoid loading the same entry twice makes sense in execution. But the field validator has its value helping the UI to indicate which field has invalid input.

hsong-rh (approved):

zkayyali812 (approved):

Alex-Izquierdo (commented):
Should we have a test to check that an user without permissions can not edit the activation?

bzwei:
> Should we have a test to check that an user without permissions can not edit the activation?

It is automatically covered in test_crud_permissions.py

PR #1176, opened 2025-01-07
feat: support PG Notify for event streams using credentials

mkanoor:
Create a Postgres credential that can store certificates and keys or userid/password at server startup and attach it to Activations that use EventStream.

This allows us to test mTLS for Postgres, a sample pg_hba file is attached to this PR.

To test this you need to create certificates and keys for
 * Postgres Server called (server.crt and server.key)
 * EDA Server called (client.crt and client.key)
 * You also need to create the CA certificate (ca.crt)

These files have to be present in
  tools/docker/postgres_ssl_config/certs

The docker-compose file tools/docker/docker-compose-mac-pg-mtls.yml can be used to test mTLS

The docker-compose file tools/docker/docker-compose-mac.yml can be used to test userid/password auth

https://issues.redhat.com/browse/AAP-38315

### DIFF
--- a/src/aap_eda/api/serializers/activation.py
+++ b/src/aap_eda/api/serializers/activation.py
@@ -66,6 +66,16 @@ REQUIRED_KEYS = [
     "rulebook_hash",
 ]
 
+PG_NOTIFY_DSN = (
+    "host={{postgres_db_host}} port={{postgres_db_port}} "
+    "dbname={{postgres_db_name}} user={{postgres_db_user}} "
+    "password={{postgres_db_password}} sslmode={{postgres_sslmode}} "
+    "sslcert={{eda.filename.postgres_sslcert|default(None)}} "
+    "sslkey={{eda.filename.postgres_sslkey|default(None)}} "
+    "sslpassword={{postgres_sslpassword|default(None)}} "
+    "sslrootcert={{eda.filename.postgres_sslrootcert|default(None)}}"
+)
+
 
 @dataclass
 class VaultData:
@@ -73,17 +83,8 @@ class VaultData:
     password_used: bool = False
 
 
-def _update_event_stream_source(
-    validated_data: dict, vault_data: VaultData
-) -> str:
+def _update_event_stream_source(validated_data: dict) -> str:
     try:
-        vault_data.password_used = True
-        encrypted_dsn = encrypt_string(
-            password=vault_data.password,
-            plaintext=settings.PG_NOTIFY_DSN,
-            vault_id=EDA_SERVER_VAULT_LABEL,
-        )
-
         source_mappings = yaml.safe_load(validated_data["source_mappings"])
         sources_info = {}
         for source_map in source_mappings:
@@ -92,7 +93,7 @@ def _update_event_stream_source(
 
             sources_info[obj.name] = {
                 "ansible.eda.pg_listener": {
-                    "dsn": encrypted_dsn,
+                    "dsn": PG_NOTIFY_DSN,
                     "channels": [obj.channel_name],
                 },
             }
@@ -334,7 +335,7 @@ class ActivationListSerializer(serializers.ModelSerializer):
         )
         eda_credentials = [
             EdaCredentialSerializer(credential).data
-            for credential in activation.eda_credentials.all()
+            for credential in activation.eda_credentials.filter(managed=False)
         ]
         extra_var = (
             replace_vault_data(activation.extra_var)
@@ -476,8 +477,14 @@ class ActivationCreateSerializer(serializers.ModelSerializer):
 
         if validated_data.get("source_mappings", []):
             validated_data["rulebook_rulesets"] = _update_event_stream_source(
-                validated_data, vault_data
+                validated_data
             )
+            eda_credentials = validated_data.get("eda_credentials", [])
+            postgres_cred = models.EdaCredential.objects.filter(
+                name=settings.DEFAULT_SYSTEM_PG_NOTIFY_CREDENTIAL_NAME
+            ).first()
+            eda_credentials.append(postgres_cred.id)
+            validated_data["eda_credentials"] = eda_credentials
 
         vault = _get_vault_credential_type()
 
@@ -651,7 +658,7 @@ class ActivationReadSerializer(serializers.ModelSerializer):
         )
         eda_credentials = [
             EdaCredentialSerializer(credential).data
-            for credential in activation.eda_credentials.all()
+            for credential in activation.eda_credentials.filter(managed=False)
         ]
         extra_var = (
             replace_vault_data(activation.extra_var)
--- a/src/aap_eda/core/management/commands/create_initial_data.py
+++ b/src/aap_eda/core/management/commands/create_initial_data.py
@@ -13,10 +13,12 @@
 #  limitations under the License.
 import hashlib
 import logging
+import os
 from urllib.parse import urlparse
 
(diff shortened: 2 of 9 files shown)


mkanoor:
The api-e2e-tests work but the api-e2e-multinode-tests fails, seems like a flaky test unrelated to changes in this PR.

src/aap_eda/settings/default.py
```diff
@@ -806,7 +797,11 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
     f"port={DATABASES['default']['PORT']} "
     f"dbname={DATABASES['default']['NAME']} "
     f"user={DATABASES['default']['USER']} "
-    f"password={DATABASES['default']['PASSWORD']}"
+    f"password={DATABASES['default']['PASSWORD']} "
+    f"sslmode={settings.get('PGSSLMODE', default='prefer')} "
```
Alex-Izquierdo, line 801: I'm not sure if this is the right approach to handle the settings. We should be honoring `DATABASES[alias][options]`
So we could be able to use or override any existing alias defined. This is important also to support a different database for pg_notify. And that would be in line with AWX too.
Even if we use a global unique setting for `PGSSLCERT`and `PGSSLKEY` and so on, those variables should have a prefix or suffix to clearly indicate where are used and avoid ambiguities.

src/aap_eda/api/serializers/activation.py
```diff
@@ -334,7 +335,7 @@ def to_representation(self, activation):
         )
         eda_credentials = [
             EdaCredentialSerializer(credential).data
-            for credential in activation.eda_credentials.all()
+            for credential in activation.eda_credentials.filter(managed=False)
```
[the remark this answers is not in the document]
mkanoor: @Alex-Izquierdo Credentials can be managed by the system (managed=True) or managed by the user (managed=False) when we list credentials we only want the users credentials and not the ones that we use internally. The Postgres credential is a managed credential that we dont want the user to view and delete.

src/aap_eda/core/management/commands/create_initial_data.py
```diff
@@ -1167,6 +1269,54 @@ def _copy_scm_credentials(self):
             "Control eda-credentials"
         )
 
+    def _update_postgres_credentials(self):
+        cred_type = models.CredentialType.objects.filter(
+            name=enums.DefaultCredentialType.POSTGRES
+        ).first()
```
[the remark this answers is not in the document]
mkanoor: @Alex-Izquierdo The filter function returns a query set we need to execute the query set to get the results so you could use all, first, list(qs).

zkayyali812 (approved):
Approving, @mkanoor can you also add some documentation that explains how this can be tested?

hsong-rh (dismissed):

src/aap_eda/core/management/commands/create_initial_data.py
```diff
@@ -1167,6 +1269,54 @@ def _copy_scm_credentials(self):
...
+        inputs = {
+            "postgres_db_host": settings.ACTIVATION_DB_HOST,
+            "postgres_db_port": settings.DATABASES["default"]["PORT"],
+            "postgres_db_name": settings.DATABASES["default"]["NAME"],
+            "postgres_db_user": settings.DATABASES["default"]["USER"],
+            "postgres_db_password": settings.DATABASES["default"]["PASSWORD"],
+            "postgres_sslmode": settings.PGSSLMODE,
```
Alex-Izquierdo, line 1282: This should be taken from DATABASES: https://github.com/ansible/eda-server/blob/d7fd30ca98545520faff1df2f95161508c310610/src/aap_eda/settings/default.py#L260

Otherwise it could lead to confusion or not honor what it is provided in the settings.yaml (potentially breaking the RPM installer)
_DEFAULT_PG_NOTIFY_DSN_SERVER has the same issue. 

We should change at least this and later we can treat separately the other things that were discussing.

src/aap_eda/settings/default.py
```diff
@@ -781,15 +781,6 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
...
 
-_DEFAULT_PG_NOTIFY_DSN = (
-    f"host={ACTIVATION_DB_HOST} "
-    f"port={DATABASES['default']['PORT']} "
-    f"dbname={DATABASES['default']['NAME']} "
-    f"user={DATABASES['default']['USER']} "
-    f"password={DATABASES['default']['PASSWORD']}"
-)
-
-PG_NOTIFY_DSN = settings.get("PG_NOTIFY_DSN", _DEFAULT_PG_NOTIFY_DSN)
```
Alex-Izquierdo, line 792: Right now the user is able to override this setting through EDA_PG_NOTIFY_DSN. I believe it is very unlikely that someone has overrided this, but, it's still a breaking change if we just remove it. 

So once there is a decision about if want to continue supporting it, isn't? 
Should be honor it when we create the eda_credential?

src/aap_eda/settings/default.py
```diff
@@ -823,3 +818,9 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
 MAX_PG_NOTIFY_MESSAGE_SIZE = int(
     settings.get("MAX_PG_NOTIFY_MESSAGE_SIZE", 6144)
 )
+
+DEFAULT_SYSTEM_PG_NOTIFY_CREDENTIAL_NAME = "_DEFAULT_EDA_PG_NOTIFY_CREDS"
+PGSSLMODE = settings.get("PGSSLMODE", default="prefer")
```
Alex-Izquierdo, line 823: As I point out in my other comment, we should not need to re-define this setting since `PGSSLMODE` is read for `databases` in ` _get_databases_settings()`

mkanoor: @Alex-Izquierdo Done

src/aap_eda/settings/default.py
```diff
@@ -806,8 +797,13 @@ def get_rulebook_process_log_level() -> RulebookProcessLogLevel:
     f"port={DATABASES['default']['PORT']} "
     f"dbname={DATABASES['default']['NAME']} "
     f"user={DATABASES['default']['USER']} "
-    f"password={DATABASES['default']['PASSWORD']}"
+    f"password={DATABASES['default']['PASSWORD']} "
+    f"sslmode={DATABASES['default']['OPTIONS']['sslmode']} "
+    f"sslcert={DATABASES['default']['OPTIONS']['sslcert']} "
+    f"sslkey={DATABASES['default']['OPTIONS']['sslkey']} "
```
Alex-Izquierdo, line 803: The optional params must be dynamic or we must make sure first they already exists, either in a different function or in `_get_databases_settings`
because DATABASES can be arbitrarily provided through the settings.yaml, otherwise we would raise KeyError here.

Alex-Izquierdo (changes requested):

Alex-Izquierdo (approved):
I have not tested it because it would require some time as I have to replicate the env for linux systems (only mac provided) but so far, LGTM. 
We need to revisit how we handle the database and pgnotify config as we were discussing.

src/aap_eda/settings/default.py
mkanoor: @Alex-Izquierdo This is not exposed via any of our installers so we can pretty safely remove it. We only have a single database support currently when we have a separate Database for EventStream we will have to revisit this.

hsong-rh (approved):

PR #1177, opened 2025-01-09
feat: [AAP-34609] - Add created_by and modified_by fields

hsong-rh:
https://issues.redhat.com/browse/AAP-34609

The `created_by` and `modified_by` are added to `Credential, Activation, EventStream, Project` and `DecisionEnvironment`. 

The two fields look like:

```
"created_by": { "id": 1, "username": "admin", "first_name": "", "last_name": "" }
"modified_by": { "id": 1, "username": "admin", "first_name": "", "last_name": "" }
```


src/aap_eda/core/models/base.py
```diff
@@ -40,3 +44,43 @@ def summary_fields(self):
             "id": self.id,
             "name": self.name,
         }
+
+
+class PrimordialModel(models.Model):
```
Alex-Izquierdo, line 49: Where comes from this "primordial" naming? is it not a bit generic/meaningless name?
Should we not use the CreatableModel and ModifiableModel that DAB provides? 
https://github.com/ansible/django-ansible-base/blob/1cf551b7834453dcfd09de1a18b44f9538f93852/ansible_base/lib/abstract_models/common.py#L113

https://github.com/ansible/django-ansible-base/blob/1cf551b7834453dcfd09de1a18b44f9538f93852/ansible_base/lib/abstract_models/common.py#L49

hsong-rh: @Alex-Izquierdo it's from controller: https://github.com/ansible/awx/blob/e605883592aa996ac65eceab3abc3106ab212e14/awx/main/models/base.py#L286

Alex-Izquierdo: yeah, but our reference should be DAB, am I wrong?

hsong-rh: I think the controller's way is more suitable for us than DAB's. Using CreatableModel and ModifiableModel means we almost need to touch every model to extract created and modified fields.

Alex-Izquierdo (approved):

zkayyali812 (approved):

PR #1182, opened 2025-01-13
fix: [AAP-38471] - add default values to the user inputs of credentials

hsong-rh:
https://issues.redhat.com/browse/AAP-38471

### DIFF
--- a/src/aap_eda/core/utils/credentials.py
+++ b/src/aap_eda/core/utils/credentials.py
@@ -33,6 +33,7 @@ from aap_eda.core.utils.crypto.base import SecretValue
 
 if typing.TYPE_CHECKING:
     from aap_eda.core import models
+
 from aap_eda.core.utils.awx import validate_ssh_private_key
 
 ENCRYPTED_STRING = "$encrypted$"
@@ -563,3 +564,18 @@ def _add_file_template_keys(context: dict, files: dict):
             context["eda"]["filename"][parts[1]] = ""
         else:
             context["eda"] = {"filename": {parts[1]: ""}}
+
+
+def add_default_values_to_user_inputs(schema: dict, inputs: dict) -> dict:
+    for field in schema.get("fields", []):
+        key = field.get("id")
+        field_type = field.get("type", "string")
+        default_value = field.get("default")
+
+        if key not in inputs:
+            if field_type == "string":
+                inputs[key] = default_value or ""
+            if field_type == "boolean":
+                inputs[key] = default_value or False
+
+    return inputs
--- a/src/aap_eda/wsapi/consumers.py
+++ b/src/aap_eda/wsapi/consumers.py
@@ -19,9 +19,13 @@ from aap_eda.core.enums import DefaultCredentialType
 from aap_eda.core.exceptions import (
     DuplicateEnvKeyError,
     DuplicateFileTemplateKeyError,
+    InvalidEnvKeyError,
 )
 from aap_eda.core.models.activation import ActivationStatus
-from aap_eda.core.utils.credentials import get_secret_fields
+from aap_eda.core.utils.credentials import (
+    add_default_values_to_user_inputs,
+    get_secret_fields,
+)
 from aap_eda.core.utils.strings import extract_variables, substitute_variables
 from aap_eda.tasks import orchestrator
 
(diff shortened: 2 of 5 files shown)


src/aap_eda/core/utils/credentials.py
```diff
@@ -563,3 +564,17 @@ def _add_file_template_keys(context: dict, files: dict):
             context["eda"]["filename"][parts[1]] = ""
         else:
             context["eda"] = {"filename": {parts[1]: ""}}
+
+
+def add_default_values_to_user_inputs(schema: dict, inputs: dict) -> dict:
+    for field in schema.get("fields", []):
+        key = field.get("id", "")
+        field_type = field.get("type", "")
```
mkanoor, line 572: @hsong-rh The default field type is string if not specified

src/aap_eda/wsapi/consumers.py
```diff
@@ -531,39 +537,40 @@ def get_file_contents_from_credentials(
...
+                user_inputs = yaml.safe_load(
+                    eda_credential.inputs.get_secret_value()
                 )
+                if "env" not in injectors:
+                    continue
 
-        if not env_vars:
-            return None
-        return yaml.dump(env_vars)
+                add_default_values_to_user_inputs(schema_inputs, user_inputs)
```
Alex-Izquierdo, line 556: Adding a default value unconditionally would not break any logic that depends on the existence of the variable? For example, it is not the same if you inject a variable for a token with an empty value rather than not injecting it at all. Since the input has not a "default" value in the schema, I think it is assumed that if is not defined, it should not exist, unless it is defined as a required field, in which case it should have a value. 

I feel for the builtin credential types we should handle the absence of not required fields and not pass any undefined field that is not required for builtin or custom types.

src/aap_eda/core/utils/credentials.py
```diff
@@ -563,3 +564,17 @@ def _add_file_template_keys(context: dict, files: dict):
             context["eda"]["filename"][parts[1]] = ""
         else:
             context["eda"] = {"filename": {parts[1]: ""}}
+
+
+def add_default_values_to_user_inputs(schema: dict, inputs: dict) -> dict:
+    for field in schema.get("fields", []):
+        key = field.get("id", "")
```
Alex-Izquierdo, line 571: is not the ID mandatory?

src/aap_eda/wsapi/consumers.py
mkanoor: @Alex-Izquierdo The problem is not that the problem is in the injectors be it file or env the user can access one of these keys for which the user has not provided a value. There is a single credential type called RH AAP which has userid/password, userid/token just token or nothing just the host URL only the host field is required. The other fields might be there or missing but the injectors are built to access all these variables. We can't change the injectors so we have to provide defaults.

hsong-rh: @mkanoor I added a check on the result of the `substitute_variables` function and removed undefined fields if it returns. The absence of not-required fields now is skipped.

mkanoor: @hsong-rh @Alex-Izquierdo I tried this by checking on AAP and we have to send all the env vars that are listed in the injectors down. Its possible for customers when they write their own credential types to provide defaults on a field by field basis which we should honor.

mkanoor (approved):

Alex-Izquierdo (commented):
I don't want to be a pain, but I really think all the static functions should have unit tests, especially since they are pure functions and therefore easy to test. In this case, we have involved three of them without any tests: add_default_values_to_user_inputs, substitute_envs, and get_env_vars_from_credentials.

hsong-rh:
> I don't want to be a pain, but I really think all the static functions should have unit tests, especially since they are pure functions and therefore easy to test. In this case, we have involved three of them without any tests: add_default_values_to_user_inputs, substitute_envs, and get_env_vars_from_credentials.

@Alex-Izquierdo I agree with you about test coverage. In this PR the test for `add_default_values_to_user_inputs` has already been included. The consumer test is a little bit different, the newly added test case `test_handle_workers_with_env_vars` has covered both `substitute_envs` and `get_env_vars_from_credentials` functions.

Alex-Izquierdo (approved):

PR #1183, opened 2025-01-13
feat: reattach event streams when rulebook changes

bzwei:
Perform source swapping at activation editing when event source mapping is provided, no matter a new rulebook is selected or not. Clear existing event stream maaping if a new rulebook is selected but no event source mapping is provided.

AAP-37298: Reattach event streams when rulebook has changed

### DIFF
--- a/src/aap_eda/api/serializers/activation.py
+++ b/src/aap_eda/api/serializers/activation.py
@@ -585,9 +585,20 @@ class ActivationUpdateSerializer(serializers.ModelSerializer):
         else:
             vault_data = VaultData()
         if self.validated_data.get("source_mappings", []):
+            if not rulebook_id:
+                # load the original ruleset
+                self.validated_data[
+                    "rulebook_rulesets"
+                ] = activation.rulebook.rulesets
+
+            # update with new event streams
             self.validated_data[
                 "rulebook_rulesets"
             ] = _update_event_stream_source(self.validated_data, vault_data)
+        elif rulebook_id:
+            # user selects a rulebook but no mapping, clear existing mappings
+            self.validated_data["source_mappings"] = ""
+            self.validated_data["event_streams"] = []
 
         if self.validated_data.get("eda_credentials"):
             _update_extra_vars_from_eda_credentials(
@@ -611,8 +622,8 @@ class ActivationUpdateSerializer(serializers.ModelSerializer):
         self, instance: models.Activation, validated_data: dict
     ) -> None:
         update_fields = []
-        eda_credentials = []
-        event_streams = []
+        eda_credentials = None
+        event_streams = None
         for key, value in validated_data.items():
             if key == "eda_credentials":
                 eda_credentials = value
@@ -624,10 +635,16 @@ class ActivationUpdateSerializer(serializers.ModelSerializer):
             update_fields.append(key)
 
         instance.save(update_fields=update_fields)
-        if eda_credentials:
-            instance.eda_credentials.set(eda_credentials)
-        if event_streams:
-            instance.event_streams.set(event_streams)
+        if eda_credentials is not None:
+            if eda_credentials:
+                instance.eda_credentials.set(eda_credentials)
+            else:
+                instance.eda_credentials.clear()
+        if event_streams is not None:
+            if event_streams:
+                instance.event_streams.set(event_streams)
+            else:
+                instance.event_streams.clear()
 
     def to_representation(self, activation):
         extra_var = (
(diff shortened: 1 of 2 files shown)


src/aap_eda/api/serializers/activation.py
```diff
@@ -591,8 +602,8 @@ def update(
         self, instance: models.Activation, validated_data: dict
     ) -> None:
         update_fields = []
-        eda_credentials = []
-        event_streams = []
+        eda_credentials = None
```
mkanoor, line 605: @bzwei You can have more than 1 EventStreams and more than 1 Credential that goes with it.

mkanoor:
@bzwei Is the use case for this that the user had an Event Stream attached to a credential and then they want to remove the EventStream during the edit phase and clear the Source Mapping. Or is the usecase that the user picks a new Event Stream(s) and wants to use the new set of Event Streams. Or is it both.

src/aap_eda/api/serializers/activation.py
bzwei: validated_data["event_streams"] and validated_data["eda_credentials"] are already collections. eda_credentials here will be either None or a collection.

bzwei:
> @bzwei Is the use case for this that the user had an Event Stream attached to a credential and then they want to remove the EventStream during the edit phase and clear the Source Mapping. Or is the usecase that the user picks a new Event Stream(s) and wants to use the new set of Event Streams. Or is it both.

Yes we cover both, the user can either clear a source mapping (thus the event_streams attribute), or use new mappings.

src/aap_eda/api/serializers/activation.py
```diff
@@ -565,9 +565,20 @@ def prepare_update(self, activation: models.Activation):
...
+                    "rulebook_rulesets"
+                ] = activation.rulebook.rulesets
+
+            # update with new event streams
             self.validated_data[
                 "rulebook_rulesets"
             ] = _update_event_stream_source(self.validated_data, vault_data)
+        elif rulebook_id:
+            # user selects a rulebook but no mapping, clear existing mappings
+            self.validated_data["source_mappings"] = ""
```
hsong-rh, line 580: Default `source_mappings` should be an empty list, right? we have a validation check for it

src/aap_eda/api/serializers/activation.py
```diff
@@ -565,9 +565,20 @@ def prepare_update(self, activation: models.Activation):
...
+                # load the original ruleset
+                self.validated_data[
+                    "rulebook_rulesets"
+                ] = activation.rulebook.rulesets
+
+            # update with new event streams
             self.validated_data[
                 "rulebook_rulesets"
             ] = _update_event_stream_source(self.validated_data, vault_data)
+        elif rulebook_id:
```
hsong-rh, line 578: can we combine this check with https://github.com/ansible/eda-server/pull/1183/files#diff-b75687856dd6a2e0a5e5a1d5d322ed63ca815ecd7fbfbb6aadb25009b4d0e89bL552?

src/aap_eda/api/serializers/activation.py
```diff
@@ -604,10 +615,16 @@ def update(
             update_fields.append(key)
 
         instance.save(update_fields=update_fields)
-        if eda_credentials:
-            instance.eda_credentials.set(eda_credentials)
-        if event_streams:
-            instance.event_streams.set(event_streams)
+        if eda_credentials is not None:
```
hsong-rh, line 618: what's the difference between these 2 lines?
```
if eda_credentials is not None:
    if eda_credentials:
```
Same question for the check on `event_streams`

src/aap_eda/api/serializers/activation.py
```diff
@@ -473,13 +471,188 @@ def create(self, validated_data):
...
+        self, instance: models.Activation, validated_data: dict
+    ) -> None:
+        update_fields = []
+        eda_credentials = None
+        event_streams = None
+        for key, value in validated_data.items():
+            if key == "eda_credentials":
+                eda_credentials = value
+                continue
+            elif key == "event_streams":
```
mkanoor, line 611: @bzwei The event_streams are built from the source mappings since we have to support multiple sources in a rulebook, the rulebook could have multiple sources and the source mappings tells us which which source to replace with which event stream, you could have sources that have to kept intact, so we can't be directly accepting the event_streams from the caller. The event_streams should get cleared based on the presence of source_mappings and regenerated.

src/aap_eda/api/serializers/activation.py
bzwei: It is a yaml string. The validation is running against the parsed dict from the string.

src/aap_eda/api/serializers/activation.py
bzwei: There are three states. None means do nothing. Empty means clear existing value. A true list means replace with new value.

src/aap_eda/api/serializers/activation.py
bzwei: The event source mappings for updating need to be completed. event_streams here are rebuilt from the source mappings given by user.

hsong-rh (approved):

zkayyali812 (approved):

PR #1185, opened 2025-01-14
fix: [AAP-38546] - Skip anonymous user in updating created_by/modified_by fields

hsong-rh:
https://issues.redhat.com/browse/AAP-38546

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -21,6 +21,7 @@ from unittest.mock import MagicMock, create_autospec
 import pytest
 from ansible_base.rbac.models import DABPermission, RoleDefinition
 from django.conf import settings
+from django.contrib.auth.models import AnonymousUser
 from django.contrib.contenttypes.models import ContentType
 from django.test import override_settings
 from rest_framework.test import APIClient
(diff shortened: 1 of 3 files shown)


tests/integration/conftest.py
```diff
@@ -138,6 +144,15 @@ def admin_client(base_client: APIClient, admin_user: models.User) -> APIClient:
     return base_client
 
 
+@pytest.fixture
+def anonymous_client(
+    base_client: APIClient, anonymous_user: models.User
+) -> APIClient:
+    """Return a pre-configured instance of an APIClient with admin user."""
```
mkanoor, line 151: @hsong-rh This is an anonymous user can we fix the comment.

hsong-rh:
> @hsong-rh Have you check how works `get_current_user` when connected to gateway? DAB already has an util to get the user that maybe we should use.

@Alex-Izquierdo I followed the way the controller does: using Django crum middleware package to get the current user. Let me check how DAB does this.

hsong-rh:
@Alex-Izquierdo DAB uses the same logic to get the current user. We all use the crum package.

Alex-Izquierdo (approved):

mkanoor (approved):

PR #1188, opened 2025-01-15
[AAP-37476]: Validate DE Urls

zkayyali812:
This PR adds validation to DE image URLs. 
Upon checking this issue out, we only had validation on registry credentials, and no validation on DE's. 

This PR ensures we have the same validation on both registry credential URLs and Decision environments.
This PR also updates the validation to ensure that if a DE URL with digest is specified, that this is a valid specification.

### DIFF
--- a/src/aap_eda/api/serializers/decision_environment.py
+++ b/src/aap_eda/api/serializers/decision_environment.py
@@ -74,10 +74,7 @@ class DecisionEnvironmentCreateSerializer(serializers.ModelSerializer):
     def validate(self, data):
         eda_credential_id = data.get("eda_credential_id")
         image_url = data.get("image_url") or self.instance.image_url
-        if eda_credential_id:
-            validators.check_if_de_valid(image_url, eda_credential_id)
-        else:
-            validators.check_if_de_valid(image_url)
+        validators.check_if_de_valid(image_url, eda_credential_id)
 
         return data
 
--- a/src/aap_eda/api/serializers/decision_environment.py
+++ b/src/aap_eda/api/serializers/decision_environment.py
@@ -12,8 +12,6 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from urllib.parse import urlparse
-
 from rest_framework import serializers
 
 from aap_eda.api.serializers.eda_credential import EdaCredentialRefSerializer
--- a/src/aap_eda/api/serializers/decision_environment.py
+++ b/src/aap_eda/api/serializers/decision_environment.py
@@ -12,6 +12,8 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+from urllib.parse import urlparse
+
 from rest_framework import serializers
 
 from aap_eda.api.serializers.eda_credential import EdaCredentialRefSerializer
@@ -73,9 +75,11 @@ class DecisionEnvironmentCreateSerializer(serializers.ModelSerializer):
 
     def validate(self, data):
         eda_credential_id = data.get("eda_credential_id")
+        image_url = data.get("image_url") or self.instance.image_url
         if eda_credential_id:
-            image_url = data.get("image_url") or self.instance.image_url
             validators.check_if_de_valid(image_url, eda_credential_id)
+        else:
+            validators.check_if_de_valid(image_url, -1)
 
         return data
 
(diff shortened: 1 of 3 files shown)


src/aap_eda/core/validators.py
```diff
@@ -132,38 +138,40 @@ def check_if_de_valid(image_url: str, eda_credential_id: int):
             % {"image_url": image_url, "tag": tag}
         )
 
-    credential = get_credential_if_exists(eda_credential_id)
-    inputs = yaml.safe_load(credential.inputs.get_secret_value())
-    credential_host = inputs.get("host")
+    if eda_credential_id != -1:
```
Alex-Izquierdo, line 141: Why this `-1`? Seems to me a bit odd. What I think we should do it's just make `eda_credential_id` optional

tests/integration/api/test_decision_environment.py
```diff
@@ -327,6 +327,92 @@ def test_create_decision_environment_with_empty_credential(
...
+    admin_client: APIClient,
+    preseed_credential_types,
+):
+    data_in = {
+        "name": "de1",
+        "description": "desc here",
+        "image_url": image_url,
+        "organization_id": default_organization.id,
+    }
+    response = admin_client.post(
```
Alex-Izquierdo, line 405: same logic should be tested for PATCH, isn`t?

src/aap_eda/core/validators.py
zkayyali812: fixed! Great point

tests/integration/api/test_decision_environment.py
zkayyali812: Yes, good point, I added another test to validate on patch as well

src/aap_eda/core/validators.py
```diff
@@ -62,7 +62,7 @@ def check_if_de_exists(decision_environment_id: int) -> int:
     return decision_environment_id
 
 
-def check_if_de_valid(image_url: str, eda_credential_id: int):
+def check_if_de_valid(image_url: str, eda_credential_id: int = None):
```
Alex-Izquierdo, line 65: The type checking is not correct because `None` is not an int. You can use [typing.Optional](https://docs.python.org/3/library/typing.html#typing.Optional)
```suggestion
def check_if_de_valid(image_url: str, eda_credential_id: tp.Optional[int] = None):
```

src/aap_eda/api/serializers/decision_environment.py
```diff
@@ -73,9 +73,11 @@ class DecisionEnvironmentCreateSerializer(serializers.ModelSerializer):
 
     def validate(self, data):
         eda_credential_id = data.get("eda_credential_id")
+        image_url = data.get("image_url") or self.instance.image_url
         if eda_credential_id:
-            image_url = data.get("image_url") or self.instance.image_url
             validators.check_if_de_valid(image_url, eda_credential_id)
+        else:
```
Alex-Izquierdo, line 79: Minor: Since you are making the credential id optional in `check_if_de_valid` there is no need for the conditional here, you can simplify it just calling `validators.check_if_de_valid(image_url, eda_credential_id)` always.

tests/integration/api/test_decision_environment.py
```diff
@@ -327,6 +327,189 @@ def test_create_decision_environment_with_empty_credential(
         assert status_message in str(errors)
 
 
+@pytest.mark.parametrize(
```
Alex-Izquierdo, line 330: Suggestion: 
It seems that the test parameters are basically the same across both tests. The only difference is the status code, in POST is 201, and in PATCH 200. What do you think if the parameter has only a generic variable instead of the status code, for example a boolean called "expected_sucess" and determine inside the test what is the expected return code for a successful or failed request. In that way you can reuse the same set of parameters across both tests.

zkayyali812: Good point, I just added this

src/aap_eda/api/serializers/decision_environment.py
zkayyali812: added

Alex-Izquierdo (approved):
Looking [at the spec](https://github.com/opencontainers/image-spec/blob/main/descriptor.md#registered-algorithms) I'm not sure if we should support as well sha512. Anycase LGTM

zkayyali812:
@Alex-Izquierdo great note! I just pushed a commit to ensure this case is covered as well. Better sooner than later

Alex-Izquierdo (approved):

ttuffin (approved):

PR #1191, opened 2025-01-16
feat: new API to copy an activation

bzwei:
AAP-37297

### DIFF
--- a/src/aap_eda/api/serializers/activation.py
+++ b/src/aap_eda/api/serializers/activation.py
@@ -498,6 +498,54 @@ class ActivationCreateSerializer(serializers.ModelSerializer):
         return super().create(validated_data)
 
 
+class ActivationCopySerializer(serializers.ModelSerializer):
+    name = serializers.CharField(
+        required=True, validators=[validators.check_if_activation_name_used]
+    )
+
+    class Meta:
+        model = models.Activation
+        fields = ["name"]
+
+    def copy(self) -> dict:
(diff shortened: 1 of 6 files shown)


src/aap_eda/api/serializers/activation.py
```diff
@@ -498,6 +499,54 @@ def create(self, validated_data):
...
+            "restart_policy": activation.restart_policy,
+            "awx_token_id": activation.awx_token,
+            "log_level": activation.log_level,
+            "eda_credentials": activation.eda_credentials.all(),
+            "k8s_service_name": activation.k8s_service_name,
+            "source_mappings": activation.source_mappings,
+            "event_streams": activation.event_streams.all(),
+            "skip_audit_events": activation.skip_audit_events,
+            "rulebook_name": activation.rulebook.name,
+            "rulebook_rulesets": activation.rulebook_rulesets,
```
mkanoor, line 531: @bzwei Like we had discussed this data could be different from the running  Activation and if event streams are involved this will fail since it wont be using the event streams. A good test would be to copy an Activation with Event Streams and start both of them we should be seeing both the Activations getting data from the shared Event Stream. If the rulebook has not been updated we would have created the links for the EventStream from the newly minted Activation and the EventStreams would work. If the project sync has changed the rulebook the behavior will be different.

zkayyali812 (approved):

mkanoor (approved):

PR #1196, opened 2025-01-23
[AAP-39136] - Register DAB Feature Flags endpoint

zkayyali812:
- Adds feature-flag API from the django-ansible-base.
- Currently only LIST endpoint is added

API is accessible at - `/api/eda/v1/feature_flags_state/`.

To test, deploy EDA and ensure API is accessible at above endpoint. One can modify the FLAGS variable to see the state reflected at the API endpoint at installtime.

Also satisfies - https://issues.redhat.com/browse/AAP-39595

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,13 +48,14 @@ cryptography = ">=42,<43"
 kubernetes = "26.1.*"
 podman = "5.2.*"
 rq-scheduler = "^0.10"
-django-ansible-base = { git = "https://github.com/ansible/django-ansible-base.git", tag = "2025.1.3", extras = [
+django-ansible-base = { git = "https://github.com/zkayyali812/django-ansible-base.git", branch = "zk/feature-flag/api", extras = [
     "channel-auth",
     "rbac",
     "redis-client",
     "resource-registry",
     "jwt-consumer",
     "rest-filters",
+    "feature-flags",
 ] }
 jinja2 = ">=3.1.3,<3.2"
 django-split-settings = "^1.2.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ cryptography = ">=42,<43"
 kubernetes = "26.1.*"
 podman = "5.2.*"
 rq-scheduler = "^0.10"
-django-ansible-base = { git = "https://github.com/zkayyali812/django-ansible-base.git", branch = "zk/feature-flag/api", extras = [
+django-ansible-base = { git = "https://github.com/ansible/django-ansible-base.git", branch = "devel", extras = [
     "channel-auth",
     "rbac",
     "redis-client",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ cryptography = ">=42,<43"
 kubernetes = "26.1.*"
 podman = "5.2.*"
 rq-scheduler = "^0.10"
-django-ansible-base = { git = "https://github.com/ansible/django-ansible-base.git", branch = "devel", extras = [
+django-ansible-base = { git = "https://github.com/ansible/django-ansible-base.git", tag = "2025.1.31", extras = [
     "channel-auth",
     "rbac",
     "redis-client",
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -202,10 +202,10 @@ INSTALLED_APPS = [
 ]
 
 
-def toggle_feature_flags(_flags, _settings):
-    for feature in _flags.keys():
-        if feature in _settings:
-            _flags[feature][0]["value"] = _settings.get(feature)
+def toggle_feature_flags(flags, settings):
+    for feature in flags.keys():
+        if feature in settings:
+            flags[feature][0]["value"] = settings.get(feature)
 
 
 # Defines feature flags, and their conditions.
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -48,13 +48,14 @@ cryptography = ">=42,<43"
 kubernetes = "26.1.*"
 podman = "5.2.*"
 rq-scheduler = "^0.10"
-django-ansible-base = { git = "https://github.com/ansible/django-ansible-base.git", tag = "2025.1.3", extras = [
+django-ansible-base = { git = "https://github.com/ansible/django-ansible-base.git", branch = "devel", extras = [
```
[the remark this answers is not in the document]
zkayyali812: We will need to move to devel here to consume the update. 
Once an upstream tag is cut, I think we can/should revert back to a tag

src/aap_eda/settings/default.py
```diff
@@ -204,6 +205,10 @@ def _get_boolean(name: str, default=False) -> bool:
 # See https://cfpb.github.io/django-flags/
 FLAGS = {}
 
+for feature in list(FLAGS.keys()):
```
hsong-rh, line 208: We don't need to call `list` here, right?

hsong-rh:
@zkayyali812 the coverage report said there are two missings. Could you improve it?

pyproject.toml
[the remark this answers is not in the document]
Alex-Izquierdo: @zkayyali812 We decided to pin the version for stability and better tracking the dependency. Here the solution would be requesting a new upstream release for DAB, no coming back to a branch.

src/aap_eda/settings/default.py
zkayyali812: You're right, I can remove list here, will update

zkayyali812:
Hey @hsong-rh ,
I tried to cover these lines for a bit, but haven't been able to figure out a way to sufficiently cover them in our unit testing, but I am able to confirm they work as expected in running deployments.

I was thinking that this may be sufficient for now, and ideally cover them them in our test-suite, once ATF is updated to consume this API. however if you know how I can cover these better, I would love to learn.

src/aap_eda/settings/default.py
Alex-Izquierdo, line 208: I think it would be better to encapsulate the logic in a function, so the logic to fetch the flag can be tested easily.

src/aap_eda/settings/default.py
```diff
@@ -204,6 +205,10 @@ def _get_boolean(name: str, default=False) -> bool:
...
+    if feature in settings:
+        FLAGS[feature][0]["value"] = settings.get(feature)
```
Alex-Izquierdo, line 210: Are we not going to support multiple conditions? https://cfpb.github.io/django-flags/usage/#defining-flags

zkayyali812: In this first iteration, we will only support booleans

pyproject.toml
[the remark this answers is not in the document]
zkayyali812: I asked if an upstream tag can be cut here, but hoping not to block this PR on it if possible

src/aap_eda/settings/default.py
zkayyali812: Good suggestion, I updated

zkayyali812:
@hsong-rh / @Alex-Izquierdo I think I addressed the code coverage here. Not sure if I can retrigger the codecov automation to update its comment above.

pyproject.toml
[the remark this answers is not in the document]
Alex-Izquierdo: Really IMO devel branch is a stepback that we should not do, specially when we are going to introduce new features that relies on DAB, we need a constraint to know what is the code that we need.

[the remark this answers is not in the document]
zkayyali812: Updated! I did not know we could self-service cut these upstream tags.

hsong-rh (dismissed):

src/aap_eda/settings/default.py
```diff
@@ -195,14 +195,23 @@ def _get_boolean(name: str, default=False) -> bool:
...
     "ansible_base.jwt_consumer",
     "ansible_base.rest_filters",
+    "ansible_base.feature_flags",
     # Local apps
     "aap_eda.api",
     "aap_eda.core",
 ]
 
+
+def toggle_feature_flags(_flags, _settings):
```
Alex-Izquierdo, line 205: Why not just assign the value of `FLAGS` from this function constructor, following the same pattern that we already have for databases or redis? 
In this way we have a function that is not pure and modifies the global state which is not a good practice and source of potential issues.

zkayyali812: After syncing on Slack, we will leave this as is now, but can potentially iterate on this later

Alex-Izquierdo, line 205: Just a tickle request, can you just remove the underscore for `_flags` and `_settings_` variables? the underscore is used to indicate a private context which I think has no meaning here.

zkayyali812: @Alex-Izquierdo Done 👍

Alex-Izquierdo: thanks and sorry :)

Alex-Izquierdo (approved):

hsong-rh (approved):

Dostonbek1 (approved):

PR #1199, opened 2025-01-24
feat: add edited_at field to activation

bzwei:
Expose this field in API for activations. Also write the field to rulebook process log if the activation has been edited.

implements AAP-38382


src/aap_eda/services/activation/activation_manager.py
```diff
@@ -207,6 +207,13 @@ def _start_activation_instance(self):
 
         self.db_instance.refresh_from_db()
         log_handler = self.container_logger_class(self.latest_instance.id)
+        if self.db_instance.edited_at:
+            edited_at = self.db_instance.edited_at.strftime(
+                "%Y-%m-%dT%H:%M:%S.%fZ"
+            )
+            log_handler.write(
+                f"The activation was edited at {edited_at}", flush=True
```
mkanoor, line 215: @bzwei @hsong-rh How would this be handled in the UI, should the UI display a separate field which is TZ aware when displaying the Edited At. The logs are good for record keeping, so we should keep this.

mkanoor: @bzwei Can you paste a screen shot of the Activation log in this PR?

mkanoor: @bzwei To be consistent can be use the created_at if the edited_at is missing that way the log will remain the same.

bzwei: @mkanoor The edited_at timestamp is not the message logging time. It is part of the log message, just like other timestamps appear in the output. Screenshot is posted.
I purposely not log the edited time if not present. When it is missing, it is an indication that the activation is original, never updated. Or perhaps I can write a message here "The activation has never been edited"

mkanoor: @bzwei Thats ok leave it the way it is for now. Did you forget to the add the screen shot?

src/aap_eda/core/migrations/0057_activation_edited_at.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 4.2.16 on 2025-01-23 19:06
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("core", "0056_alter_activation_options"),
```
hsong-rh, line 8: where is this 0056 migration file? do you miss it?

bzwei: 0056 migration exists in feature/edit-activation branch

mkanoor (approved):

hsong-rh (approved):

PR #1200, opened 2025-01-28
fix: prevent race conditions when dispatching start requests - AAP-36642

Alex-Izquierdo:
The (eda) dispatcher dispatches the start request (chooses the node) and enqueue the manager at the end. The request is processed again in the next loop of the monitor if it takes time and the request can be assigned to another node, then the first time the monitor is executed for the activation, causes the following error:
```
2024-11-22 21:33:49,363 Container args ['ansible-rulebook', '--worker', '--websocket-ssl-verify', 'yes', '--websocket-url', 'wss://f34-h10-000-r640.rdu2.scalelab.redhat.com:443/api/eda/ws/ansible-rulebook', '--websocket-access-token', '******', '--websocket-refresh-token', '******', '--websocket-token-url', 'https://f34-h10-000-r640.rdu2.scalelab.redhat.com:443/api/eda/v1/auth/token/refresh/', '--id', '1388', '--heartbeat', '300']
2024-11-22 21:33:49,464 Container 52b40cfb987c96cc8577cfc0e742b86ffca32441d29dcca10252181675329d03 is running.
2024-11-22 21:34:06,671 Missing container for running activation. Restart policy is applied. 
```

This issue can be reproduced easily by adding a `time.sleep(6)` in any point of the start flow, to force the monitor to be executed again. 

This PR fixes the problem by ensuring at the beginning of the dispatcher that no request is processed if there is already a RQ JOB in a not finished state. This has been observed and addressed in a similar manner (but with advisory locks) in the PoC of ansible-dispatcher. 

Also fixes a race condition when there is a window where the activation instance is in STARTING state but it has not activation_instance created yet. 

Jira: https://issues.redhat.com/browse/AAP-36642

### DIFF
--- a/src/aap_eda/core/tasking/__init__.py
+++ b/src/aap_eda/core/tasking/__init__.py
@@ -598,18 +598,26 @@ def unique_enqueue(queue_name: str, job_id: str, *args, **kwargs) -> Job:
     Detects if a job with the same id is already enqueued and if it is
     it will return it instead of enqueuing a new one.
     """
-    for name in settings.RQ_QUEUES:
-        job = job_from_queue(name, job_id)
-        if job:
-            logger.info(
-                f"Skip enqueing job: {job_id} because it is already enqueued"
-            )
-            return job
+    job = get_pending_job(job_id)
+    if job:
+        logger.info(
+            f"Skip enqueing job: {job_id} because it is already enqueued"
+        )
+        return job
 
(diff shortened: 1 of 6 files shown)


Alex-Izquierdo:
I updated the original description of the PR because I think this problem has been here from the first implementation of the eda-dispatcher.

ttuffin (approved):

PR #1201, opened 2025-01-30
EDA_ANALYTICS: Rename feature-flag

zkayyali812:
Renames feature flag to be inline with naming convention FEATURE_(FEATURE_NAME)_ENABLED

### DIFF
--- a/src/aap_eda/core/management/commands/gather_analytics.py
+++ b/src/aap_eda/core/management/commands/gather_analytics.py
@@ -71,8 +71,8 @@ class Command(BaseCommand):
         opt_since = options.get("since")
         opt_until = options.get("until")
 
-        if not flag_enabled("EDA_ANALYTICS"):
-            self.logger.error("EDA_ANALYTICS is disabled.")
+        if not flag_enabled("FEATURE_EDA_ANALYTICS_ENABLED"):
+            self.logger.error("FEATURE_EDA_ANALYTICS_ENABLED is set to False.")
             return
 
         since = parser.parse(opt_since) if opt_since else None
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -210,7 +210,6 @@ def toggle_feature_flags(flags, settings):
 
 # Defines feature flags, and their conditions.
 # See https://cfpb.github.io/django-flags/
-FLAGS = {}
 FLAGS = {
     "FEATURE_EDA_ANALYTICS_ENABLED": [
         {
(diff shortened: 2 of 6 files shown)


src/aap_eda/core/management/commands/gather_analytics.py
```diff
@@ -71,8 +71,8 @@ def handle(self, *args, **options):
         opt_since = options.get("since")
         opt_until = options.get("until")
 
-        if not flag_enabled("EDA_ANALYTICS"):
-            self.logger.error("EDA_ANALYTICS is disabled.")
+        if not flag_enabled("FEATURE_EDA_ANALYTICS_ENABLED"):
+            self.logger.error("FEATURE_EDA_ANALYTICS_ENABLED is disabled.")
```
Dostonbek1, line 75: Should this say `FEATURE_EDA_ANALYTICS_ENABLED is false`? Having enabled in the name and using enable/disable as a verb seems a bit weird.

zkayyali812: We can if this makes more sense to folks, In my opinion though `FEATURE_EDA_ANALYTICS_ENABLED is false` seems stranger to me though

Alex-Izquierdo: I would agree with @Dostonbek1 Features are disabled or enabled, flags are True or False. 

IHMO, the message looks better as `Feature eda analytics is disabled.` or `FEATURE_EDA_ANALYTICS_ENABLED is set to False.`

zkayyali812: Sounds good, I will change it to `FEATURE_EDA_ANALYTICS_ENABLED is set to False.`

src/aap_eda/settings/default.py
```diff
@@ -546,7 +545,7 @@ def get_rq_queues() -> dict:
 RULEBOOK_QUEUE_NAME = settings.get("RULEBOOK_QUEUE_NAME", "activation")
 
 RQ_STARTUP_JOBS = []
-if settings.get("FEATURE_EDA_ANALYTICS", True):
+if settings.get("FEATURE_EDA_ANALYTICS_ENABLED", True):
```
Alex-Izquierdo, line 548: Is not this conditional a potential bug? If `FEATURE_EDA_ANALYTICS_ENABLED` does not exist in settings we would be using "True" as default and scheduling the job when we should not do it.  

I think we should use `FLAGS` since it is the source of truth, and we are overriding it from settings in [toggle_feature_flags](https://github.com/ansible/eda-server/blob/e997de71c60a11dd01e1b1fe99e61877431e3c84/src/aap_eda/settings/default.py#L222)

The logic of this conditional is also prone error, because we are overriding `RQ_STARTUP_JOBS`, it should be:
```
RQ_STARTUP_JOBS = []
if settings.get("FEATURE_EDA_ANALYTICS_ENABLED", True):
    RQ_STARTUP_JOBS.append(
        {
            "func": "aap_eda.tasks.analytics.schedule_gather_analytics",
            "job_id": "start_analytics_scheduler",
        })
```

zkayyali812: Hey @Alex-Izquierdo ,
Updated to include this. Let me know what you think

Alex-Izquierdo (approved):

src/aap_eda/settings/default.py
```diff
@@ -212,11 +212,10 @@ def toggle_feature_flags(flags, settings):
 # See https://cfpb.github.io/django-flags/
 FLAGS = {}
```
bzwei, line 213: remove this line. It was due to merging conflict

zkayyali812: excellent catch

zkayyali812: Fixed.

bzwei (approved):

Dostonbek1 (approved):

PR #1195, opened 2025-01-23
feat: [AAP-38978] - improve rulebook activations logs messages

hsong-rh:
https://issues.redhat.com/browse/AAP-38978

PR #1204, opened 2025-02-04
fix: [AAP-38379] - Rename and fix migration dependencies

kaiokmo:
This is needed due to migration `0056_rulebookprocesslog_log_created_at_and_more` already existing in the main branch.


Alex-Izquierdo:
Here we have another issue. Why https://github.com/ansible/eda-server/pull/1195 did not fail the tests

Alex-Izquierdo:
Ok, the thing is that this branch should be rebased prior to any later PR @bzwei

kaiokmo:
abandoning in favor of #1203

PR #1205, opened 2025-02-04
fix: tests on activation after merge conflicts

bzwei:
AAP-38379. 
insert pg notify credential when update an activation similar to crate an activation.

### DIFF
--- a/src/aap_eda/api/serializers/activation.py
+++ b/src/aap_eda/api/serializers/activation.py
@@ -655,7 +655,10 @@ class ActivationUpdateSerializer(serializers.ModelSerializer):
             # update with new event streams
             self.validated_data[
                 "rulebook_rulesets"
-            ] = _update_event_stream_source(self.validated_data, vault_data)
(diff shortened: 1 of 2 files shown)


src/aap_eda/api/serializers/activation.py
```diff
@@ -679,6 +682,18 @@ def prepare_update(self, activation: models.Activation):
                 ),
             )
 
+    def _insert_pg_notify_credential(self, activation: models.Activation):
+        if "eda_credentials" in self.validated_data:
+            eda_credentials = self.validated_data.get("eda_credentials")
+        else:
+            eda_credentials = [cred.id for cred in activation.eda_credentials]
+        postgres_cred = models.EdaCredential.objects.filter(
```
mkanoor, line 690: @bzwei This might be better
```
postgres_cred = models.EdaCredential.objects.get(name=DEFAULT_SYSTEM_PG_NOTIFY_CREDENTIAL_NAME)
```

mkanoor (approved):

PR #1206, opened 2025-02-05
feat: read analytics settings from gateway 

bzwei:
also remove settings API that is no longer needed

AAP-35922
AAP-37453


bzwei:
@hsong-rh as discussed, the two CI test failure seems not directly related. They should be fixed in another PR.

hsong-rh:
> @hsong-rh as discussed, the two CI test failure seems not directly related. They should be fixed in another PR.

Ok, let's fix it later

hsong-rh (approved):

PR #1208, opened 2025-02-06
fix: eda_credentials field cannot be null

bzwei:
fix AAP-39630.
eda_credentials cannot be null when create or update an activation. empty list [] is allowed.


bzwei:
The Duplication on New Code warning can be ignored. It complains about the test which can hardly be deduped.

Alex-Izquierdo (approved):

PR #1209, opened 2025-02-10
test: add more tests to analytics settings and scheduling

bzwei:
AAP-39884 Test and improve code coverage for analytics collection feature

### DIFF
--- a/src/aap_eda/conf/registry.py
+++ b/src/aap_eda/conf/registry.py
@@ -242,14 +242,22 @@ class SettingsRegistry(object):
         url = f"{settings.RESOURCE_SERVER['URL']}/{sync_slug}"
         token = resource_server.get_service_token()
         logger.info(f"Getting remote settings from {url}")
-        res = requests.get(
-            url,
-            headers={"X-ANSIBLE-SERVICE-AUTH": token},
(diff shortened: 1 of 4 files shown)


src/aap_eda/conf/registry.py
```diff
@@ -242,14 +242,22 @@ def _resync_remote_settings(self, sync_group: str) -> None:
...
-            headers={"X-ANSIBLE-SERVICE-AUTH": token},
-            verify=settings.RESOURCE_SERVER["VALIDATE_HTTPS"],
-        )
+        try:
+            res = requests.get(
+                url,
+                headers={"X-ANSIBLE-SERVICE-AUTH": token},
+                verify=settings.RESOURCE_SERVER["VALIDATE_HTTPS"],
+            )
+        except Exception as e:
```
Alex-Izquierdo, line 251: Is there a reason to catch a generic exception?

Alex-Izquierdo: to catch any exception from requests I think you can just catch `RequestException` which is parent of the rest. 
https://github.com/psf/requests/blob/main/src/requests/exceptions.py#L12

Alex-Izquierdo (approved):

PR #1212, opened 2025-02-13
[AAP-39892] Optimize analytics collection to create tar files

hsong-rh:
https://issues.redhat.com/browse/AAP-39892

This PR covers:
1. remove `payload` from audit event collection table,
2. reduce SQL parameter injection and memory overflow risks

### DIFF
--- a/src/aap_eda/analytics/analytics_collectors.py
+++ b/src/aap_eda/analytics/analytics_collectors.py
@@ -1,6 +1,5 @@
 #  Copyright 2024 Red Hat, Inc.
-#
-#  Licensed under the Apache License, Version 2.0 (the "License");
+# Licensed under the Apache License, Version 2.0 (the "License");
 #  you may not use this file except in compliance with the License.
 #  You may obtain a copy of the License at
 #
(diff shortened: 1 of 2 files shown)


src/aap_eda/analytics/analytics_collectors.py
```diff
@@ -866,12 +747,66 @@ def _get_audit_action_qs(since: datetime, until: datetime):
...
+                        break
+                    byte_data = bytes(data)
+                    # Process data in chunks of buffer_size
+                    for i in range(0, len(byte_data), buffer_size):
+                        chunk = byte_data[i : i + buffer_size]
+                        file.write(chunk.decode())
+
+            cursor.execute(f"DROP VIEW IF EXISTS {view_name}")
+        return file.file_list()
+    except DatabaseError as e:
```
Alex-Izquierdo, line 804: Are you sure we want to pass on these errors?

hsong-rh: @Alex-Izquierdo This function is repeatedly used during file dumping by periodic or manual tasks. If an error occurs, IMO we should log it and let the task continue.

Alex-Izquierdo (dismissed):

Alex-Izquierdo (approved):

ttuffin (approved):

PR #1215, opened 2025-02-19
chore: mount volumes to docker-compose-mac.yml for hot reloading

Dostonbek1:
This helps improve quality of life for developers to be able to hot reload changes in podman-compose environment, without having to restart the containers.

### DIFF
--- a/tools/docker/docker-compose-mac.yml
+++ b/tools/docker/docker-compose-mac.yml
@@ -107,6 +107,7 @@ services:
       retries: 10
     volumes:
       - ${EDA_HOST_PODMAN_SOCKET_URL:-/run/user/501/podman/podman.sock}:/run/podman/podman.sock:z
+      - '../../:/app/src:z'
 
   eda-ws:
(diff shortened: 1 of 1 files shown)


tools/docker/docker-compose-mac.yml
```diff
@@ -107,6 +107,7 @@ services:
       retries: 10
     volumes:
       - ${EDA_HOST_PODMAN_SOCKET_URL:-/run/user/501/podman/podman.sock}:/run/podman/podman.sock:z
+      - '../../:/app/src:z'
```
mkanoor, line 110: @Dostonbek1 What would the ../.. be relative to if we are in the root directory of the project? Should we have a SRC_DIR var defined in the docker-compose file

Alex-Izquierdo: @mkanoor it is relative to the location of the compose file. It is also in line with the rest of the compose files. No need any variable.

Alex-Izquierdo (approved):
LGTM 
If you make this compose to run podman as a service as we do in the rest of compose files, that would be great too.

zkayyali812 (approved):

PR #1216, opened 2025-02-20
AAP-40526 feat: add logging at startup

Alex-Izquierdo:
Log relevant versions and settings at worker startup. 
- Add a logger wrapper helper to log messages unconditionally. 
- Defines a list of safe/relevant settings to be logged. 
- Working and tested in all workers (ws, http, rq)

Also fixed a minor bug when calling the logger in scheduler command. 

Jira: https://issues.redhat.com/browse/AAP-40526


src/aap_eda/logging/startup.py
```diff
@@ -0,0 +1,94 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import logging
+import platform
+
+from django.conf import settings
+
+from aap_eda.logging import UnconditionalLogger
```
mkanoor, line 20: @Alex-Izquierdo Do we need another logger to be defined, can't we just explicitly set the log level in this function. for example
```
from django.apps import AppConfig
import logging
from aap_eda.utils import get_eda_version
logger = logging.getLogger(__name__)


class CoreConfig(AppConfig):
    default_auto_field = "django.db.models.BigAutoField"
    name = "aap_eda.core"

    def ready(self):
        logger.setLevel(logging.INFO)
        logger.info(f"EDA Version: {get_eda_version()}")
        # make sure we apply DAB decorations in case they are not yet imported
        from aap_eda.api.views import dab_decorate  # noqa: F401

```

Alex-Izquierdo: Hi @mkanoor I'm not adding a new logger, it is just a simple wrapper that enforces the message get printed regardless of the log level or logger. Here we have a simple situation: we want to print something at startup regardless of the logger and regardless of the log level. I don't think we should be using a level that doesn't match the app config and we should not be taking care of changing the level for an existing logger at runtime. 

I would agree that creating a new logger with a new custom level is probably excessive for what we need to do, so to me simplest solution is this.

Alex-Izquierdo: I have removed the STARTUP_LOGGING_ENABLED setting as we discussed in the chat. Please review again.

mkanoor (approved):

tests/unit/test_logging.py
```diff
@@ -0,0 +1,35 @@
...
+    assert "Starting eda-server" in caplog.text
+    assert get_eda_version() in caplog.text
+    assert "Python version" in caplog.text
+    assert platform.python_version() in caplog.text
+    for setting in SETTINGS_LIST_FOR_LOGGING:
+        assert setting in caplog.text
+        assert str(getattr(settings, setting)) in caplog.text
+
+    # some assertions to check that sensitive data is not logged
+    assert "some_random_not_so_secret_key" not in caplog.text
```
zkayyali812, line 26: Minor nit - 
I think you could also use
```
assert settings.SECRET_KEY not in caplog.text
```
Instead of passing in this string directly

zkayyali812 (approved):

tests/unit/test_logging.py
Alex-Izquierdo: Agree, nice suggestion. I will be honest. I'm scared to lose the approvals, it is hard to get them xD

zkayyali812: Next iteration perhaps :)

ttuffin (approved):

tests/unit/test_logging.py
Alex-Izquierdo: Nope, I just was crying. I don't like to leave known issues, specially when they are so trivial. :)
It is updated, please review again.

zkayyali812 (approved):

PR #1223, opened 2025-02-25
feat: use DAB to load dynaconf settings

bzwei:
# [AAP-37673] EDA: Use DAB:Dynaconf to load Django static settings

## Description

It is the organization's effect to load settings using the same method. 

EDA used to define all settings and expose to Django settings in `aap_eda/settings/default.py`. We continue to use this file to load and expose settings through DAB:Dynaconf. We move individual settings from this file to other files:
* all settings that can be overwritten are moved to `aap_eda/settings/defaults.py`
* all internal used settings that are not exposed for overwriting are moved to `aap_eda/settings/core.py`
* all computed or adjusted settings and logic are moved to `aap_eda/settings/post_load.py`
* redis specific settings preparations are moved to `aap_eda/settings/redis.py`

Closes AAP-37673
Jira: [AAP-37673](https://jira.aap.com/browse/AAP-37673) 

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Test update
- [ ] Refactoring (no functional changes)
- [ ] Development environment change
- [ ] Configuration change
- [ ] CI change

## Self-Review Checklist

- [ ] I have performed a self-review of my code
- [ ] I have added relevant comments to complex code sections
- [ ] I have updated documentation where needed
- [ ] I have considered the security impact of these changes
- [ ] I have considered performance implications
- [ ] I have thought about error handling and edge cases
- [ ] I have tested the changes in my local environment
- [ ] I have run the linters and test suite locally
- [ ] I have tested the changes on integration environment

### Steps to Test

1.
2.
3.

### Required Actions

- [ ] Requires documentation updates
  
- [ ] Requires downstream repository changes
  
- [ ] Requires infrastructure/deployment changes
  
- [ ] Requires coordination with other teams
  
- [ ] Blocked by PR/MR: #XXX | <https://github.com/example/repo/pull/XXX>


src/aap_eda/settings/default.py
```diff
@@ -1,833 +1,38 @@
...
 
-ANSIBLE_RULEBOOK_LOG_LEVEL = get_rulebook_process_log_level()
-ANSIBLE_RULEBOOK_FLUSH_AFTER = settings.get(
-    "ANSIBLE_RULEBOOK_FLUSH_AFTER", 100
+DYNACONF = factory(
+    __name__,
+    "EDA",
+    # Options passed directly to dynaconf
+    environments=("production", "development", "testing"),
+    settings_files=["defaults.py"],
```
zkayyali812, line 23: I recommend deleting the `environments` line and having all settings files specified in `settings_files` 
```suggestion
    settings_files=["constants.py", "defaults.py", "development_defaults.py"],
```

Taskfile.dist.yaml
```diff
@@ -1,7 +1,8 @@
 version: "3"
 
 env:
-  DJANGO_SETTINGS_MODULE: "aap_eda.settings.development"
+  EDA_MODE: development
```
zkayyali812, line 4: I dont think we need this. We never had an `EDA_MODE` before, no need to add it now

bzwei: `EDA_MODE: development` is a feature introduced in DAB for selecting `aap_eda.settings.development_defaults` on top of defaults. It needs to be in the list of `environments=("production", "development", "testing"),`

src/aap_eda/settings/default.py
bzwei: The `environments` is needed for selecting a setting file. "development_defaults.py" cannot be placed in the settings_files because it is not always loaded, depending on whether `EDA_MODE` selects it or not.

`constants.py` is to be loaded after the ENVs are loaded.

src/aap_eda/settings/post_load.py
```diff
@@ -0,0 +1,412 @@
...
+    settings.SECRET_KEY = _get_secret_key(settings)
+    settings.DEBUG = get_boolean(settings, "DEBUG")
+    settings.ALLOWED_HOSTS = _get_list_from_str(settings, "ALLOWED_HOSTS")
+    # A list or a comma separated string of allowed origins for CSRF protection
+    # in the form of [scheme://]host[:port]. Supports wildcards.
+    # More info: https://docs.djangoproject.com/en/4.2/ref/settings/#csrf-trusted-origins  # noqa: E501
+    settings.CSRF_TRUSTED_ORIGINS = _get_list_from_str(
+        settings, "CSRF_TRUSTED_ORIGINS"
+    )
+    toggle_feature_flags(settings)
```
rochacbruno, line 309: Toggle feature flags is called later at the end of this function, is calling it twice needed?

bzwei: good catch. Will remove this line

poetry.lock
```diff
@@ -649,7 +649,7 @@ bcrypt = ["bcrypt"]
 
 [[package]]
 name = "django-ansible-base"
-version = "2025.1.31"
+version = "2025.1.3"
```
Alex-Izquierdo, line 652: This is an older ref of DAB. I believe this is wrong or maybe a ref development that now in main it should point to a new release.

pyproject.toml
```diff
@@ -48,7 +48,7 @@ cryptography = ">=42,<43"
 kubernetes = "26.1.*"
 podman = "5.2.*"
 rq-scheduler = "^0.10"
-django-ansible-base = { git = "https://github.com/ansible/django-ansible-base.git", tag = "2025.1.31", extras = [
+django-ansible-base = { git = "https://github.com/rochacbruno/django-ansible-base.git", branch = "dynaconf_settings", extras = [
```
Alex-Izquierdo, line 51: Wrong repo url

tests/integration/api/test_feature_flags.py
```diff
@@ -45,9 +44,17 @@ def test_feature_flags_override_flags(admin_client):
             {"condition": "boolean", "value": False},
         ],
     },
+    FEATURE_SOME_PLATFORM_FLAG_ENABLED=True,
```
Alex-Izquierdo, line 47: It does not seem to be a real flag, isn't?

tools/docker/docker-compose-dev.yaml
Alex-Izquierdo, line 1: What about the other compose files and minikube?

tests/integration/api/test_feature_flags.py
bzwei: no, just for testing

pyproject.toml
Alex-Izquierdo: Ok, as I saw the PR as ready I thought you wanted to merge.

tools/docker/docker-compose-dev.yaml
Alex-Izquierdo: and the EDA_MODE?

src/aap_eda/settings/default.py
```diff
@@ -11,823 +11,39 @@
...
-DEFAULT_RULEBOOK_QUEUE_TIMEOUT = settings.get(
-    "DEFAULT_RULEBOOK_QUEUE_TIMEOUT", 120
+from ansible_base.lib import dynamic_config
+from ansible_base.lib.dynamic_config import (
+    export,
+    factory,
+    load_envvars,
+    load_standard_settings_files,
 )
+from split_settings.tools import include
```
rochacbruno, line 23: We have to remove the usage of split_settings and replace with dynaconf equivalent, that is bacause when Dynaconf loads it is tracked on the inspection history.

src/aap_eda/settings/default.py
```diff
@@ -11,823 +11,39 @@
...
-ANSIBLE_RULEBOOK_LOG_LEVEL = get_rulebook_process_log_level()
-ANSIBLE_RULEBOOK_FLUSH_AFTER = settings.get(
-    "ANSIBLE_RULEBOOK_FLUSH_AFTER", 100
+DYNACONF = factory(
+    __name__,
+    "EDA",
+    # Options passed directly to dynaconf
+    environments=("production", "development", "testing"),
+    settings_files=["defaults.py"],
 )
```
rochacbruno, line 35: You can pass `add_dab_settings=False` to skip adding DAB defaults, then later after loading the `constants` you call `load_dab_settings` 

https://github.com/ansible/django-ansible-base/pull/689/files#diff-d7dafe47cdd096243886611525e1f000eca6786be0908ee3c16bbcd08c7bf7a9R25

src/aap_eda/settings/default.py
```diff
@@ -11,823 +11,39 @@
...
+DYNACONF.load_file(EDA_SETTINGS_FILE)
+load_envvars(DYNACONF)  # load envvars prefixed with EDA_
+DYNACONF.load_file("constants.py")  # load internal non-overwritable settings
+post_loading(DYNACONF)
+export(__name__, DYNACONF)  # export back to django.conf.settings
 
 dab_settings = os.path.join(
     os.path.dirname(dynamic_config.__file__), "dynamic_settings.py"
 )
 include(dab_settings)
```
rochacbruno, line 49: You can now replace this with:

```suggestion
load_dab_settings(DYNACONF)

export(__name__, DYNACONF)  # export back to django.conf.settings
```

`load_dab_settings` is now implemented as a separate function to allow you to call it later.

Ensure `export` is the last thing you call in the django settings module

src/aap_eda/settings/post_load.py
```diff
@@ -0,0 +1,409 @@
...
+                "id": "resync_shared_resources",
+            }
+        )
+        settings.RQ_PERIODIC_JOBS = jobs
+
+
+def post_loading(settings: Dynaconf):
+    settings.SECRET_KEY = _get_secret_key(settings)
+    settings.DEBUG = get_boolean(settings, "DEBUG")
+    settings.ALLOWED_HOSTS = _get_list_from_str(settings, "ALLOWED_HOSTS")
```
rochacbruno, line 297: It is ok to mutate the `settings` object directly here like `settings.FOO = "BAR"`

However, those changes will not be tracked when running `dynaconf inspect` commands.

I would suggest using dictionaries to aggregate settings to be mutated.

example:

```py

def post_loading(settings: Dynaconf):
    data = {}
    data["FOO"] = "BAR"
    data["THING"] = value

    # then finally
    settings.update(data, loader_identifier="post_loading")

```

The `loader_identifier` is just a unique string to be shown when running `dynaconf inspect -k FOO` so when troubleshooting we know exactly where those values came from.

src/aap_eda/settings/defaults.py
rochacbruno, line 1: Will not this cause confusion due to similar file names? [constants, default, defaults]

What is the reason for having `constants.py` are those meant to never be overridden?
if so, how would DAB change values on settings defined there such as `TEMPLATES` and `INSTALLED_APPS` ?

src/aap_eda/settings/default.py
```diff
@@ -11,823 +11,39 @@
...
-# ---------------------------------------------------------
-# DJANGO ANSIBLE BASE SETTINGS
-# ---------------------------------------------------------
-from ansible_base.lib import dynamic_config  # noqa: E402
+load_standard_settings_files(
+    DYNACONF
+)  # /etc/ansible-automation-platform/*.yaml
+DYNACONF.load_file(EDA_SETTINGS_FILE)
+load_envvars(DYNACONF)  # load envvars prefixed with EDA_
+DYNACONF.load_file("constants.py")  # load internal non-overwritable settings
```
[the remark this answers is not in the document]
Alex-Izquierdo: > * Values from `/etc/eda/settings.yaml` via `DYNACONF.load_file`
> * Values from `/etc/ansible-automation-platform/*.yaml` via `load_standard_settings_files`

@rochacbruno 
Why a global setting would take precedence over the application settings? That's is not intuitive. If it is the pattern in all the AAP ok, but I feel the precedence should be always from global scope to specific scope, same as $HOME/user/.config overrides /etc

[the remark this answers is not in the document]
Alex-Izquierdo: just a clarification: `/etc/eda/settings.yaml` is not complementary. It is the upstream path equivalent to `/etc/ansible-automation-platform/eda/{settings,flags,.secrets}.yaml`. This last one replaces `/etc/eda/settings.yaml` in downstream builds @rochacbruno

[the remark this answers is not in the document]
rochacbruno: @Alex-Izquierdo thanks for clarification, so `/etc/eda/settings.yaml` is upstream only, that makes sense, and then loading it first makes sense.

One thing to notice is that on dynaconf the latest file loaded, overrides previous settings.

So what I was thinking is:

> Order of precedence, the latest wins, the latest overrides the previous.


Start with the most specific, End with the more global.

1. Defaults.py + constants.py (merge those 2?) [required]
    1. This is the initial set of values  
2. development_defaults.py or production_defaults.py  [conditionally]
    1. This overrrides existing variables loaded on step 1
3. DAB defaults coming from ansible_base.lib.dynamic_settings  [required]
    1. This overrrides existing variables loaded on steps 1-2
4. /etc/eda/settings.yaml [optional]
    1. This overrrides existing variables loaded on steps 1-3
5. /etc/ansible-automation-platform/*.yaml [optional]
    1. This overrrides existing variables loaded on steps 1-4
6. any `export EDA_*` [optional]
    1. This overrrides existing variables loaded on steps 1-5
7. Post loading modifications [required]
    1. This is the latest step, nothing overrides from here.


As you can see, `envvars` always wins, always overrrides existing settings, that is following the unix-philosofy, then we have the deployment customization on /etc, then we have defaults from DAB and defaults from the application.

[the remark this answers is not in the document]
Alex-Izquierdo: I understand the meaning of "constants" is precisely that they are hardcoded values of the application part of the core logic. That's why I believe @bzwei has called them "constants". I may missing something, why would you want to override these constants (like for example the middlewares) and potentially breaking the application?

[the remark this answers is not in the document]
bzwei: @rochacbruno I have no preference the order of `/etc/eda/settings.yaml` and `/etc/ansible-automation-platform/*.yaml`. So I will adjust as suggested. `constants` are the ones used internally not overridable by user through yaml or envvars, therefore it should be placed after step 4. It is ok for DAB to modify so called constants. Naming is hard. I thought about calling it `read-only`, `constants`, or `internal`. Any other suggestion?

src/aap_eda/settings/defaults.py
bzwei: `constants` are the ones used internally not overridable by user through yaml or envvars, it is ok for DAB to modify so called constants. The existence is to make minimal changes to the existing code base.

src/aap_eda/settings/default.py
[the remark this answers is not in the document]
Alex-Izquierdo: I see. I guess they can be considered constants in the scope of EDA and post-processed by DAB. But since are finally modified, called them "constants" seems wrong. Maybe `core` can be a more appropriate name? @rochacbruno @bzwei 

We still have variable in constants.py that we don't expect to be overridden, right? How we differentiate them? 

Should we define `ANSIBLE_BASE_OVERRIDABLE_SETTINGS` to make explicit what we expect to be overridden?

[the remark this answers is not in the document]
bzwei: @Alex-Izquierdo these are variables reference by django.conf.settings in code everywhere but not exposed to users through dynaconf importing in our old code.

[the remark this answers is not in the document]
bzwei: @rochacbruno @Alex-Izquierdo Sorry I missed a few discussions above due to refresh issue before I added my comment above. To finalize the loading order:

1. Defaults.py [required]
2. development_defaults.py or production_defaults.py [conditionally]
3. DAB defaults coming from ansible_base.lib.dynamic_settings [required]
4. /etc/eda/settings.yaml 
5. /etc/ansible-automation-platform/*.yaml [optional]
6. any export EDA_* [optional]
7. constants.py (or rename to core.py)
8. Post loading modifications [required]

I am not sure about step 3. should it be after 7 since in 7 we define INSTALLED_APPS there?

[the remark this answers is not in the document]
rochacbruno: I agree that step 3 must go after step 7, DAB uses INSTALLED_APPS to make decisions on which settings to add

rochacbruno:
You can now point to DAB devel branch

zkayyali812 (approved):

rochacbruno (approved):

Dostonbek1 (approved):

PR #1225, opened 2025-02-27
chore: add more versions to startup log

Alex-Izquierdo:
Follow up https://github.com/ansible/eda-server/pull/1216 by adding some package version that I think are relevant. 
Refactor `get_eda_version` to `get_package_version` for a more general purpose helper. 
Update tests accordingly


ttuffin (approved):

ttuffin (approved):

src/aap_eda/utils/__init__.py
```diff
@@ -21,10 +21,10 @@ def str_to_bool(value: str) -> bool:
...
-def get_eda_version() -> str:
-    """Return EDA version as defined in the aap-eda package."""
+def get_package_version(package_name: str) -> str:
+    """Return version of the given package."""
     try:
-        return importlib.metadata.version("aap-eda")
+        return importlib.metadata.version(package_name)
     except importlib.metadata.PackageNotFoundError:
-        logger.error("Cannot read version from aap-eda package")
+        logger.error("Cannot read version from %s package", package_name)
```
Dostonbek1, line 29: Should we use f-string here, for better readability?

Dostonbek1 (approved):
looks good to me, except for one minor nitpick below.

src/aap_eda/utils/__init__.py
Alex-Izquierdo: Actually I believe we are doing it wrongly in most of the places with f-strings in logger since it is by default an unsafe method. 
Refs:
https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-format-interpolation.html
https://docs.python.org/3/library/logging.html#formatter-objects

PR #1226, opened 2025-02-27
AAP-40882: Add a credential type for analytics collection

hsong-rh:
https://issues.redhat.com/browse/AAP-40882

This PR adds a new credential type for analytics collection. Two associated credentials, `BASIC` and `OAUTH,` store the c.r.c credentials of the user/password or service account, respectively.


src/aap_eda/core/enums.py
```diff
@@ -174,3 +174,10 @@ class CustomEventStreamCredentialType(DjangoStrEnum):
     GITHUB = "GitHub Event Stream"
     SNOW = "ServiceNow Event Stream"
     DYNATRACE = "Dynatrace Event Stream"
+
+
+class AnalyticsCredentialType(DjangoStrEnum):
+    """Types of authentication for Analytics."""
+
+    BASIC = "Basic Analytics"
+    TOKEN = "Token Analytics"
```
mkanoor, line 183: @hsong-rh Would you have a separate Credential Type for Token Auth?

zkayyali812:
Awesome work @hsong-rh!!
If possible, can you note the steps required to test this out possibly and validate that analytics are sent properly?

src/aap_eda/core/management/commands/create_initial_data.py
```diff
@@ -913,6 +914,45 @@
     ]
 }
 
+ANALYTICS_CREDENTIAL_INPUTS = {
+    "fields": [
+        {
+            "id": "analytics_url",
```
bzwei, line 920: This field is not provided by customer, but from installer.

We need to add field for INSIGHTS_TRACKING_STATE, which is different from the feature flag.

src/aap_eda/core/management/commands/create_initial_data.py
```diff
@@ -913,6 +914,45 @@
...
+        },
+        {
+            "id": "auth_type",
+            "label": "Analytics Authentication Type",
+            "type": "string",
+            "default": "basic",
+            "hidden": True,
+        },
+        {
+            "id": "username",
```
bzwei, line 933: shouldn't we start to use service account and token?

src/aap_eda/analytics/utils.py
```diff
@@ -141,3 +157,77 @@ def extract_job_details(
...
+                getattr(setting, keys[0], None)
+                and getattr(setting, keys[1], None)
+            )
+
+    if not any(has_valid):
+        logger.error("Missing required credentials in settings")
+        raise MissingUserPasswordError("Valid credentials not found")
+
+
+@lru_cache(maxsize=1)
```
bzwei, line 226: should not cache it because the user can change the credential any time.

src/aap_eda/analytics/utils.py
```diff
@@ -141,3 +157,77 @@ def extract_job_details(
...
+def _get_analytics_credential():
+    analytics_type = models.CredentialType.objects.get(
+        name=enums.AnalyticsCredentialType.BASIC
+    )
+    return models.EdaCredential.objects.filter(
+        credential_type=analytics_type
+    ).first()
```
bzwei, line 233: how do we prevent the user from creating multiple credentials of this type?

src/aap_eda/analytics/utils.py
hsong-rh: will remove the cache when credentials are update

src/aap_eda/analytics/utils.py
hsong-rh: Added validation logic to prevent it

src/aap_eda/api/serializers/eda_credential.py
```diff
@@ -151,6 +151,22 @@ def validate(self, data):
             id=data.get("credential_type_id")
         )
 
+        # Analytics only allows one credential
+        if (
+            credential_type.name
+            in [
+                enums.AnalyticsCredentialType.BASIC,
```
mkanoor, line 158: @hsong-rh You might want to move this out to the top as a constant and it will come in handy if we need more Singleton credentials.

src/aap_eda/settings/default.py
```diff
@@ -214,7 +214,7 @@ def toggle_feature_flags(flags, settings):
     "FEATURE_EDA_ANALYTICS_ENABLED": [
         {
             "condition": "boolean",
-            "value": False,
+            "value": True,
```
zkayyali812, line 217: We should leave default to be `False`

src/aap_eda/analytics/collector.py
```diff
@@ -35,7 +35,10 @@ def _package_class() -> package.Package:
         return package.Package
 
     def _is_shipping_configured(self) -> bool:
-        if not application_settings.INSIGHTS_TRACKING_STATE:
+        if (
+            not flag_enabled("FEATURE_EDA_ANALYTICS_ENABLED")
+            or not application_settings.INSIGHTS_TRACKING_STATE
```
bzwei, line 40: `INSIGHTS_TRACKING_STATE` needs to be exposed in the credential. The users need to restart the service to toggle the flag, but they can easily turn on/off analytics through the credential.

src/aap_eda/api/views/eda_credential.py
```diff
@@ -208,6 +209,10 @@ def partial_update(self, request, pk):
                 model_to_dict(eda_credential),
             )
 
+        # clear cache if updating analytics credentials
```
bzwei, line 212: Will the caching and clearing logic work considering they are running on difference containers?

src/aap_eda/api/serializers/eda_credential.py
```diff
@@ -151,6 +151,18 @@ def validate(self, data):
...
 
+        # Analytics only allows one credential
+        if (
+            credential_type.name in enums.SINGLETON_CREDENTIAL_TYPES
+            and models.EdaCredential.objects.filter(
+                credential_type=credential_type
+            ).exists()
+        ):
+            raise serializers.ValidationError(
+                "Only one credential is allowed for analytics type: "
```
bzwei, line 162: `SINGLETON_CREDENTIAL_TYPES` doesn't suggest analytics. So the word analytics should be removed from the message.

src/aap_eda/api/views/eda_credential.py
hsong-rh: Your suggestions?

src/aap_eda/analytics/collector.py
hsong-rh: Added

src/aap_eda/analytics/utils.py
```diff
@@ -141,3 +158,185 @@ def extract_job_details(
...
+    _validate_credential()
+    return _get_credential_value(
+        "password",
+        (settings, "REDHAT_PASSWORD"),
+        (application_settings, "REDHAT_PASSWORD"),
+        (application_settings, "SUBSCRIPTIONS_PASSWORD"),
+    )
+
+
+def get_insights_tracking_state() -> str:
```
mkanoor, line 288: @hsong-rh since we are returning a boolean or a string from this function should the return type be fixed.

src/aap_eda/analytics/utils.py
```diff
@@ -141,3 +158,185 @@ def extract_job_details(
...
+        (application_settings, "AUTOMATION_ANALYTICS_GATHER_INTERVAL"),
+    )
+    try:
+        return int(interval)
+    except ValueError as e:
+        logger.error(
+            f"Invalid analytics interval value '{interval}'. "
+            f"Using default interval. Error: {str(e)}"
+        )
+        return 14400
```
mkanoor, line 311: @hsong-rh Can we make this default a constant

src/aap_eda/api/views/eda_credential.py
hsong-rh: It makes sense. The collection mostly happens every four hours or sometimes runs manually.

hsong-rh: The cache logic is removed

zkayyali812:
@hsong-rh Wondering if we should remove the Basic auth credential type, and just leave the Oauth?
Basic auth for analytics will be deprecated in the next few months, I think this May if Im not mistaken. If we remove it now, it will lead to less confusion later on.

hsong-rh:
> @hsong-rh Wondering if we should remove the Basic auth credential type, and just leave the Oauth? Basic auth for analytics will be deprecated in the next few months, I think this May if Im not mistaken. If we remove it now, it will lead to less confusion later on.

@zkayyali812 Ok, I will remove it, and also the basic user/password from the settings

hsong-rh:
> @zkayyali812 Ok, I will remove it, and also the basic user/password from the settings

I changed my mind: I will now only remove the basic auth from the UI in case any changes happen in the future.

zkayyali812 (approved):

zkayyali812 (approved):

mkanoor (dismissed):

src/aap_eda/core/management/commands/create_initial_data.py
```diff
@@ -913,6 +914,84 @@
...
+        {
+            "id": "gather_interval",
+            "label": "Analytics Gather Interval",
+            "type": "string",
+            "help_text": "The time interval between each collection (secs)",
+            "default": "14400",
+        },
+        {
+            "id": "insights_tracking_state",
+            "label": "Enable Red Hat Cloud Analytics",
```
bzwei, line 987: Both controller and gateway have setting page for this parameter, although they used different labels. We should not introduce the third label which brings more confusion to users. We should either align with gateway or controller.
Gateway label: Insights tracking state. Help_text: Enables the service to gather data on automation and send it to Automation Analytics.
Controller label: Gather data for Automation Analytics. No help text.
I prefer to align with gateway because eventually we will use the settings from gateway.

hsong-rh: @bzwei Both places are updated accordingly

bzwei (approved):

zkayyali812 (approved):

PR #1233, opened 2025-03-10
add x request id

ShaiahWren:
Ensure that an x-request-id in the form of a UUID is included in the header when an HTTP request is made through GW.
[AAP-40972](https://issues.redhat.com/browse/AAP-40972)


mkanoor:
@ShaiahWren The [x-request-id](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_conn_man/headers#x-request-id) originates from the Envoy server when the request arrives and then it flows thru the system. Any process that wants to handle it should grab it from the http headers and then store it in a ContextVar or Thread Local storage and the log formatters have to be adjusted to access it like it happens in the gateway. We should not have to generate a new one for every activation. There is another tracing id called log_tracing_id that we would be using to track the Activation from when it starts till when its deleted, that is part of @hsong-rh 's PR. Its generated once for every Activation and we have to have a migration to put one into the old Activations. To access the x-request-id from HTTP header there is a python django package that needs to be used.

PR #1234, opened 2025-03-11
fix: [AAP-37813] Replace EDA filters with DAB rest filters

Dostonbek1:
This PR replaces our existing filters with rest-filters app that DAB provides, which is more capable and helps us align across platform.

[AAP-37813](https://issues.redhat.com/browse/AAP-37813)

NOTE: I still need to run e2e tests and identify what tests need to be updated/skipped.


Dostonbek1:
e2e test run with updated api client: https://github.com/ansible/eda-server/actions/runs/13793345570

Dostonbek1:
@lgalis this might be a breaking change for the UI depending on how it is using the current EDA filters. Could you help us confirm?

src/aap_eda/api/views/activation.py
```diff
@@ -269,6 +267,9 @@ def retrieve(self, request, pk: int):
                 description="Return a list of Activations.",
             ),
         },
+        parameters=generate_query_params(
```
Alex-Izquierdo, line 270: what a good point now `generate_query_params` is it? :)

Alex-Izquierdo (approved):

Alex-Izquierdo:
@Dostonbek1 should we put it as draft until have confirmation it will not break the UI, just to prevent accidental merges?

src/aap_eda/api/views/activation.py
Dostonbek1: yessir, pretty dynamic. :)

zkayyali812 (approved):

mkanoor:
@Dostonbek1 this looks good but we need to ensure that it works with @lgalis's UI components. Plus she would have to make changes to take advantage of the new filters which can be combined  which wasn't allowed earlier. Would it be possible to build a quay.io image for UI testing

PR #1235, opened 2025-03-11
add logger messages to activation-worker log

ShaiahWren:
Ensure error and info logs are written to the activation-worker logs, as well as the DB/UI logs. 
[AAP-41782](https://issues.redhat.com/browse/AAP-41782)

### DIFF
--- a/src/aap_eda/services/activation/activation_manager.py
+++ b/src/aap_eda/services/activation/activation_manager.py
@@ -36,10 +36,10 @@ from aap_eda.services.activation.restart_helper import (
     system_restart_activation,
 )
 
-from .db_log_handler import DBLogger
 from .engine.common import ContainerableInvalidError, ContainerEngine
 from .engine.factory import new_container_engine
 from .status_manager import StatusManager, run_with_lock
+from .tee_log_handler import TeeSystemLogger
 
 LOGGER = logging.getLogger(__name__)
 
@@ -59,7 +59,7 @@ class ActivationManager(StatusManager):
         self,
         db_instance: models.Activation,
         container_engine: tp.Optional[ContainerEngine] = None,
-        container_logger_class: type[DBLogger] = DBLogger,
+        container_logger_class: type[TeeSystemLogger] = TeeSystemLogger,
     ):
         """Initialize the Activation Manager.
 
@@ -749,7 +749,7 @@ class ActivationManager(StatusManager):
         container_logger = self.container_logger_class(self.latest_instance.id)
         container_logger.write(user_msg, flush=True)
         LOGGER.info(
-            f"Stop operation activation id: {self.db_instance.id} "
+            f"Stop operation activation id: {self.db_instance.id, LOGGER} "
             "Activation stopped.",
         )
 
--- a/src/aap_eda/services/activation/tee_log_handler.py
+++ b/src/aap_eda/services/activation/tee_log_handler.py
@@ -1,0 +1,67 @@
+#  Copyright 2025 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import logging
+from datetime import datetime, timezone
+from typing import Union
+
+from aap_eda.services.activation.db_log_handler import DBLogger
+
+LOGGER = logging.getLogger(__name__)
+
+
+class TeeSystemLogger(DBLogger):
+    """
+    A logging class that extends DBLogger to log messages both to the database
+    and the system logger.
+    """
+
+    def __init__(self, activation_instance_id: int):
+        super().__init__(activation_instance_id)
+
+    @staticmethod
+    def _convert_to_asctime(timestamp: int) -> str:
+        """Convert a UNIX timestamp to asctime format in UTC."""
+        return datetime.fromtimestamp(timestamp, tz=timezone.utc).strftime(
+            "%Y-%m-%d %H:%M:%S,%f"
+        )[:-3]
+
+    def flush(self):
+        try:
+            for buffer in self.activation_instance_log_buffer:
+                line = buffer.log
+                rulebook_timestamp = self._convert_to_asctime(
+                    buffer.log_timestamp
+                )
+                # Determine the log level
+                if "ERROR" in line:
+                    log_level = logging.ERROR
+                elif "WARN" in line:
+                    log_level = logging.WARNING
+                elif "DEBUG" in line:
+                    log_level = logging.DEBUG
+                elif "CRITICAL" in line or "FATAL" in line:
+                    log_level = logging.CRITICAL
+                else:
+                    log_level = logging.INFO
+
+                extra_data = {
+                    "rulebook_timestamp": rulebook_timestamp,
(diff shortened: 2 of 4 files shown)


src/aap_eda/services/activation/engine/podman.py
```diff
@@ -285,6 +285,13 @@ def update_logs(self, container_id: str, log_handler: LogHandler) -> None:
                 log_parts = log.split(" ", 1)
                 last_timestamp = log_parts[0]
                 if len(log_parts) > 1:
+                    message = log_parts[1]
+
+                    if "ERROR" in message:
+                        LOGGER.error(message)
+                    else:
```
hsong-rh, line 292: we should also check for WARN, DEBUG, FATAL, and DEBUG. It's not good to log them as INFO level.

src/aap_eda/services/activation/tee_log_handler.py
```diff
@@ -0,0 +1,55 @@
...
+
+import logging
+from typing import Union
+
+from aap_eda.services.activation.db_log_handler import DBLogger
+
+LOGGER = logging.getLogger(__name__)
+
+
+class TeeLogger(DBLogger):
```
Alex-Izquierdo, line 23: Can you add a docstring clarifying the purpose of the class? Maybe use a more self descriptive name like "TeeSystemLogger" too ?

Alex-Izquierdo (commented):
Can you add a test?

src/aap_eda/services/activation/tee_log_handler.py
```diff
@@ -0,0 +1,55 @@
...
+    def __init__(self, activation_instance_id: int):
```
Alex-Izquierdo, line 24: If you don't need to override or extend the initialization of the object there is no need to define `__init__` here

BUT, using the LOGGER of this module probably is wrong since all the records would be tracked by this logger and we would be losing the track of the real module that is calling it. 

Instead I think it is better to pass the logger in the constructor in a more dependency injector pattern and allow to be used by different implementers, We consume the DBLogger from the engines and from the manager but this new logger I guess is intent to be called only from the engine. 

thoughts @hsong-rh @ShaiahWren

ShaiahWren: I think you are right @Alex-Izquierdo . I'm updating the pr now, and then I'll work on the test.

Alex-Izquierdo: Are you going to implement this new logger in this PR or in a separate PR?

ShaiahWren: I will implement it in this pr, if that's alright with you. I also need to write the test this afternoon.

src/aap_eda/services/activation/tee_log_handler.py
```diff
@@ -0,0 +1,60 @@
...
+            elif "WARN" in line:
+                log_level = logging.WARNING
+            elif "DEBUG" in line:
+                log_level = logging.DEBUG
+            elif "CRITICAL" in line or "FATAL" in line:
+                log_level = logging.CRITICAL
+            else:
+                log_level = logging.INFO
+
+            self.logger.log(log_level, line)
```
mkanoor, line 60: We need to pass in the date times from ansible rulebook 
We need to define a formatter for the TeeLogger which has a placeholder for rulebook_timestamp and we can pass that thru using extra args in the log call. The log format should align with the other formatters we have for activation worker but swap out the asctime with the rulebook_timestamp 

A sample would look something like this

import logging
# this would all be done in settings/default.py
# the format should align with the rest of our formatters please check with @hsong-rh for more details
# Configure logging
logging.basicConfig(
    level=logging.INFO,
    format='%(rulebook_timestamp)s - %(levelname)s - %(message)s'
)

# Create a logger in the teesystemlogger file 
logger = logging.getLogger(__name__)

# Log a message with extra parameters
extra_data = {'rulebook_timestamp': 'convert rulebook time'}
logger.log(log_level,'container id in', extra=extra_data)

src/aap_eda/services/activation/tee_log_handler.py
```diff
@@ -0,0 +1,60 @@
+#  Copyright 2023 Red Hat, Inc.
```
mkanoor, line 1: 2025

src/aap_eda/services/activation/tee_log_handler.py
```diff
@@ -0,0 +1,60 @@
...
+
+class TeeSystemLogger(DBLogger):
+    """
+    A logging class that extends DBLogger to log messages both to the database
+    and the system logger.
+    """
+
+    def __init__(self, activation_instance_id: int, logger: logging.Logger):
+        super().__init__(activation_instance_id)
+        self.logger = logger
```
mkanoor, line 29: I think the logger should be retrieved from the file name using
'''
import logging
'''
This allows us to make custom settings just for this file

src/aap_eda/services/activation/tee_log_handler.py
Alex-Izquierdo: We need to add there as well a field for the rulebook process id, we can not have in the log all those records without the rulebook process context.

Alex-Izquierdo: @mkanoor do you want to replace the existing timestamp or add new one from the rulebook one?

mkanoor: @Alex-Izquierdo we need to preserve the timestamp from ansible-rulebook and it is present in the log_timestamp argument being passed into this function we can use that except we need to convert it to the asctime format to be consistent

Alex-Izquierdo: Agree

src/aap_eda/services/activation/tee_log_handler.py
```diff
@@ -0,0 +1,75 @@
...
+                log_level = logging.WARNING
+            elif "DEBUG" in line:
+                log_level = logging.DEBUG
+            elif "CRITICAL" in line or "FATAL" in line:
+                log_level = logging.CRITICAL
+            else:
+                log_level = logging.INFO
+
+            extra_data = {"rulebook_timestamp": rulebook_timestamp}
+            LOGGER.log(log_level, f"{self.prefix} {line}", extra=extra_data)
```
Alex-Izquierdo, line 75: if you are going to create a teelogger formater, the prefix should be part of the format

src/aap_eda/settings/post_load.py
```diff
@@ -470,3 +472,40 @@ def post_loading(loaded_settings: Dynaconf):
...
+def configure_logging():
+    """
+    Configure logging and ensure that the rulebook_timestamp
+    replaces the default asctime field.
+    """
+    log_format = "%(rulebook_timestamp)s - %(levelname)s - %(message)s"
+    logging.basicConfig(level=logging.DEBUG, format=log_format)
+
+
+def get_logging_settings():
```
Alex-Izquierdo, line 486: how you will use this function?

src/aap_eda/services/activation/activation_manager.py
```diff
@@ -203,7 +203,9 @@ def _start_activation_instance(self):
         self._create_activation_instance()
 
         self.db_instance.refresh_from_db()
-        log_handler = self.container_logger_class(self.latest_instance.id)
+        log_handler = self.container_logger_class(
+            self.latest_instance.id, LOGGER
```
Alex-Izquierdo, line 207: Either, with a logger instance or a logger name, now we have two arguments for the constructor and one of them is the same in all the calls, only `activation_instance_id` is dynamic. There is no need no repeat the same argument in all the calls, you can be more concise with a factory
```
def get_log_handler(self, activation_instance_id, logger=LOGGER):
    return self.container_logger_class(activation_instance_id, LOGGER)
```
So the lines like:
```
log_handler = self.container_logger_class(self.latest_instance.id)
```
become
```
log_handler = self.get_log_handler(self.latest_instance.id)
```

ShaiahWren:
Closing the pr and replacing with https://github.com/ansible/eda-server/pull/1253

PR #1236, opened 2025-03-13
[AAP-41917] fix: default EDA_MODE to production

bzwei:
env EDA_MODE was not set. All settings from production/development/testing are imported.
This fix sets EDA_MODE to production by default to avoid conflict settings from different modes. Different mode can still be set by the env var.

Also directly call toggle_feature_flags function from DAB.

[AAP-41917](https://issues.redhat.com/browse/AAP-41917)

For testing pick a setting and set to different value in /etc/eda/settings.yaml under production/development/testing sections and prove the value is selected according to explicitly set EDA_MODE. When EDA_MODE is not set, the value under production is selected.

### DIFF
--- a/src/aap_eda/settings/default.py
+++ b/src/aap_eda/settings/default.py
@@ -11,17 +11,24 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+import os
+
 from ansible_base.lib.dynamic_config import (
     export,
     factory,
     load_dab_settings,
     load_envvars,
     load_standard_settings_files,
+    toggle_feature_flags,
 )
 
 from .post_load import post_loading
 
-EDA_SETTINGS_FILE = "/etc/eda/settings.yaml"
+EDA_SETTINGS_FILE = os.environ.get(
+    "EDA_SETTINGS_FILE", "/etc/eda/settings.yaml"
+)
+
+os.environ["EDA_MODE"] = os.environ.get("EDA_MODE", "production")
 
 DYNACONF = factory(
     __name__,
@@ -31,7 +38,12 @@ DYNACONF = factory(
     settings_files=["defaults.py"],
 )
 
-DYNACONF.load_file(EDA_SETTINGS_FILE)
+# Insert the custom file in the list of standard settings so DAB loads it
+# without requiring mode based sections.
+DYNACONF.set(
+    "ANSIBLE_STANDARD_SETTINGS_FILES", f"@insert 3 {EDA_SETTINGS_FILE}"
+)
+
 load_standard_settings_files(
     DYNACONF
 )  # /etc/ansible-automation-platform/*.yaml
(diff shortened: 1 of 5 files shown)


src/aap_eda/settings/default.py
```diff
@@ -21,17 +23,20 @@
 
 from .post_load import post_loading
 
-EDA_SETTINGS_FILE = "/etc/eda/settings.yaml"
+EDA_SETTINGS_FILE = os.environ.get(
```
Alex-Izquierdo, line 26: I'm missing something here. why we call directly the os.environ? it should not be handled by dynaconf?

src/aap_eda/settings/default.py
```diff
@@ -21,17 +23,20 @@
...
+    "EDA_SETTINGS_FILE", "/etc/eda/settings.yaml"
+)
+
+os.environ["EDA_MODE"] = os.environ.get("EDA_MODE", "production")
```
Alex-Izquierdo, line 30: if production is the default, we should be override the value in the docker-compose development envs?

src/aap_eda/settings/default.py
bzwei: This is before we call dynaconf to import settings. We ask dynaconf to load additional settings from this file. It was hardcoded. Now can be changed by env, also good for testing.

src/aap_eda/settings/default.py
bzwei: yes. docker-compose already overwrite the value to development mode

rochacbruno (approved):

src/aap_eda/settings/default.py
```diff
@@ -11,32 +11,48 @@
...
 
 DYNACONF = factory(
     __name__,
     "EDA",
     # Options passed directly to dynaconf
     environments=("production", "development", "testing"),
-    settings_files=["defaults.py"],
+    settings_files=["defaults.py", EDA_SETTINGS_FILE],
 )
```
rochacbruno, line 40: It looks like you caught a bug that I will need to fix on DAB, so I have an alternative solution to try here.

```suggestion
    settings_files=["defaults.py"],
)

# Insert the custom file in the list of standard settings so DAB loads it
# without requiring mode based sections.
DYNACONF.set("ANSIBLE_STANDARD_SETTINGS_FILES", f"@insert 3 {EDA_SETTINGS_FILE}")
```

Can you please try the above suggestion and see if it works? the result must be:




```console
$ dynaconf inspect -k ANSIBLE_STANDARD_SETTINGS_FILES
{
  "header": {
    "env_filter": "None",
    "key_filter": "ANSIBLE_STANDARD_SETTINGS_FILES",
    "new_first": "True",
    "history_limit": "None",
    "include_internal": "False"
  },
  "current": [
    "/etc/ansible-automation-platform/settings.yaml",
    "/etc/ansible-automation-platform/flags.yaml",
    "/etc/ansible-automation-platform/.secrets.yaml",
    "/etc/eda/settings.yaml",
    "/etc/ansible-automation-platform/eda/settings.yaml",
    "/etc/ansible-automation-platform/eda/flags.yaml",
    "/etc/ansible-automation-platform/eda/.secrets.yaml"
  ],
```

tests/unit/data/demo_settings.yaml
```diff
@@ -0,0 +1,2 @@
+development:
+    SECRET_KEY: "dev.key"
```
rochacbruno, line 2:
```suggestion
SECRET_KEY: "dev.key"
```

With the suggested change above the `development` here is not needed.

Alex-Izquierdo (approved):

hsong-rh (approved):

PR #1238, opened 2025-03-17
fix: WEBSOCKET_SSL_VERIFY and CONTROLLER_SSL_VERIFY to be a bool type

bzwei:
AAP-42126

WEBSOCKET_SSL_VERIFY and CONTROLLER_SSL_VERIFY in defaults.py were set to str type because they can be strings like "yes" or "no". But a value of true or false fails the type validation, preventing the server from starting up.

[AAP-42126](https://issues.redhat.com/browse/AAP-42126)

The fix changes the type from `str` to `Union[bool, str]`. The eda-server has code to convert `yes` or `no` to a bool. When they are sent to ansible-rulebook, they become `"True"` or `"False"` which are valid and expected.

Different inputs (`true`, `yes`, `True`) have been tested and proved working in activation.

### DIFF
--- a/src/aap_eda/settings/defaults.py
+++ b/src/aap_eda/settings/defaults.py
@@ -138,7 +138,7 @@ RENAMED_USERNAME_PREFIX: str = "eda_"
 # ---------------------------------------------------------
 DEPLOYMENT_TYPE: str = "podman"
 WEBSOCKET_BASE_URL: str = "ws://localhost:8000"
-WEBSOCKET_SSL_VERIFY: str = "yes"
+WEBSOCKET_SSL_VERIFY: Union[bool, str] = "yes"
 WEBSOCKET_TOKEN_BASE_URL: Optional[str] = None
 PODMAN_SOCKET_URL: Optional[str] = None
 PODMAN_SOCKET_TIMEOUT: int = 0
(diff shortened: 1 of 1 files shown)


src/aap_eda/settings/defaults.py
```diff
@@ -138,7 +138,7 @@
 # ---------------------------------------------------------
 DEPLOYMENT_TYPE: str = "podman"
 WEBSOCKET_BASE_URL: str = "ws://localhost:8000"
-WEBSOCKET_SSL_VERIFY: str = "yes"
+WEBSOCKET_SSL_VERIFY: bool = True
```
Alex-Izquierdo, line 141: I always understood that it should be `Union[str|bool]`
Where it is converted to string? because in container request and what it is passed to ansible-rulebook it is a string.

bzwei: When we call ansible-rulebook cli, all boolean becomes str. See this log:
```
Container args ['ansible-rulebook', '--worker', '--websocket-ssl-verify', 'False', '--websocket-url'
```

bzwei: I tested various combinations (including "yes" in yaml) and proved all work no matter what's the input. But changing to Union[bool, str] does not hurt.

hsong-rh: Before we called function `_get_boolean` for this kind of situation, right? it works for both boolean and str types.
```
def _get_boolean(name: str, default=False) -> bool:
    value = settings.get(name, default)
    if isinstance(value, str):
        value = utils.str_to_bool(value)
    if not isinstance(value, bool):
        raise ImproperlyConfigured("{name} setting must be a boolean value.")
    return value


DEBUG = _get_boolean("DEBUG")
```

zkayyali812 (approved):

rooftopcellist (approved):

rooftopcellist:
I tested this code out with the eda-server-operator and it fixes the settings error we were seeing when running migrations in the initContainer.

Alex-Izquierdo (approved):

src/aap_eda/settings/defaults.py
Alex-Izquierdo: thanks!

PR #1243, opened 2025-03-24
fix: process k8s_service_name during activation update

bzwei:
Fix an internal 500 error when calling api to update an activation on k8s

[AAP-42437](https://issues.redhat.com/browse/AAP-42437)

If no new name or k8s_service_name is provided when request updating the activation, the values from the existing activation are pulled. They are needed for validation. 
In addition, when the k8s_service_name is cleared through the update request, it will be regenerated with the default value.

To test the fix
1. Update an activation without changing name on k8s, it no longer raises a 500 internal error.
2. Create an activation with a customized k8s_service_name and clear it through an update, verify the activation's k8s_service_name is assigned to a default value generated based on the activation name.

### DIFF
--- a/src/aap_eda/api/serializers/activation.py
+++ b/src/aap_eda/api/serializers/activation.py
@@ -643,6 +643,10 @@ class ActivationUpdateSerializer(serializers.ModelSerializer):
         self.validated_data["user_id"] = self.context["request"].user.id
         self.validated_data["edited_at"] = timezone.now()
         self.validated_data["edited_by"] = self.context["request"].user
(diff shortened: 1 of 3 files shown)


zkayyali812:
Hey @bzwei, did you test this on OCP to validate that this fixes the ATF tests?

bzwei:
> Hey @bzwei, did you test this on OCP to validate that this fixes the ATF tests?

Not directly on OCP. I can reproduce the issue with pytest and verify the fix with the pytest too.

zkayyali812 (approved):

hsong-rh (approved):

PR #1248, opened 2025-03-26
fix: cleanup podman containers and handle not found exceptions AAP-42708

Alex-Izquierdo:
We were creating the container with "remove=True". This was due to the first design where we had a worker attached to the container and we relied on podman to get deleted the container once it is stopped. We found that this parameter did not work properly in the podman client, so we added a cleanup logic anyway. After the refactor of activations, this was never needed, since we fetch the logs in the monitor after stop the container, before remove it. 

Now after the upgrade of the podman client, this [has been fixed ](https://github.com/containers/podman-py/pull/435)and the container is actually removed when it is stopped. Our cleanup logic run under a conditional "if container.exists()" so we can incur in race conditions since there is a time window between podman stops the container and it is get deleted, then, we can enter in the conditional and later raises an exception because the container was deleted in the meantime. 

Even in the successful case (we don't enter in the condition) now the container gets deleted once it is stopped there is a potential loss of the last output not gathered by the monitor. 

1. Remove nested try/except block, it is a bug because NotFound exception is also an ApiError exception. 
2. Don't create the container with "remove=True"
3. Consolidate log level messages to warning. 

Jira: https://issues.redhat.com/browse/AAP-42708

### DIFF
--- a/src/aap_eda/services/activation/engine/podman.py
+++ b/src/aap_eda/services/activation/engine/podman.py
@@ -81,21 +81,20 @@ class Engine(ContainerEngine):
         try:
             if self.client.containers.exists(container_id):
                 container = self.client.containers.get(container_id)
-                try:
-                    container.stop(ignore=True)
-                    LOGGER.info(f"Container {container_id} is stopped.")
-                    self.update_logs(container_id, log_handler)
-                    self._cleanup(container_id, log_handler)
-                    log_handler.write(
-                        f"Container {container_id} is cleaned up.",
-                        flush=True,
-                    )
-                except NotFound:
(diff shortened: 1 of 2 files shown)


Alex-Izquierdo:
There is another aspect which is testing. Testing this is a little tricky with mocking. I really think we should have integration tests against a real podman socket. I want to discuss this internally to be addressed in later iterations. So far, I would need this fix with some urgency since I need it to unblock errors related with the dispatcher poc.

hsong-rh (approved):

zkayyali812 (approved):

mkanoor:
https://github.com/containers/podman-py/pull/435

PR #1249, opened 2025-03-26
Fix AAP EDA Dev Env

zkayyali812:
This PR is necessary for PR - https://github.com/ansible/aap-gateway/pull/743 to work. The alternative is to export all environment variables.

This method allows for cleanly using dynaconf


AlanCoding (approved):
I'm fairly sure that `_defaults.py` are coming in at lower variable precedence than what we care about for our purposes.

What we care about the _most_ is the `settings.yml` file giving `RESOURCE_SERVER:` as a yaml dictionary, and that this should be okay with that regarding precedence / format.

bzwei:
to use development_defaults.py, you need to update https://github.com/ansible/aap-gateway/pull/743 EDA_MODE=development

hsong-rh (approved):

PR #1255, opened 2025-04-01
[AAP-42756] fix: Generate default request id when it's not available

hsong-rh:
Request ID should have a default value when there is no Gateway.

https://issues.redhat.com/browse/AAP-42756

### DIFF
--- a/src/aap_eda/api/views/activation.py
+++ b/src/aap_eda/api/views/activation.py
@@ -112,11 +112,10 @@ class ActivationViewSet(
             )
 
         if response.is_enabled:
-            request_id = request.headers.get("x-request-id", "")
             start_rulebook_process(
                 process_parent_type=ProcessParentType.ACTIVATION,
                 process_parent_id=response.id,
-                request_id=request_id,
+                request_id=request.headers.get("x-request-id"),
             )
 
         logger.info(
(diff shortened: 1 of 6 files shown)


Alex-Izquierdo (approved):

mkanoor:
@hsong-rh I think the x-request-id is the domain of the Web Server (Envoy, NGINX, Apache) we shouldn't provide our own default just for Activation. We have all kinds of API requests coming in, in production in 2.5 there is a Envoy proxy doing this for us already. If this is for development env we should see how we can have the Web Server insert the x-request-id.

Alex-Izquierdo:
> @hsong-rh I think the x-request-id is the domain of the Web Server (Envoy, NGINX, Apache) we shouldn't provide our own default just for Activation. We have all kinds of API requests coming in, in production in 2.5 there is a Envoy proxy doing this for us already. If this is for development env we should see how we can have the Web Server insert the x-request-id.

@mkanoor I reported the task. The request id is the domain of any webapp (eda server is a webapp) and once it is included in the log format it should be present. Also for consistency since the application also assigns a request id for system requests. If you want to test that it takes the value that comes from the request, you only have to run a request with such a header and confirm the right value. 
This is not for the development environment, but also for the standalone deployment.

Alex-Izquierdo (approved):

mkanoor (approved):

PR #1262, opened 2025-04-08
feat(dispatcherd): add advisory lock for task uniqueness

Alex-Izquierdo:
For the PoC of dispatcherd we used [pg advisory_locks](https://www.postgresql.org/docs/current/explicit-locking.html#ADVISORY-LOCKS) to ensure job uniqueness. It will be part of the final implementation. It has been so far a robust and simpler way to ensure uniqueness in comparison to existing `unique_enqueue` helper, which would not have a replacement. 

Since the use of the advisory lock is not tied to dispatcherd, this PR explores its use as well for the current implementation with RQ redis, assuming that will simplify the later implementation under a feature flag of the dispatcherd itself. 

There is no functional change and everything should work transparently. I have been testing it extensively and tomorrow will do more battery of tests, I have confidence. I leave the open the draft so the rest of the team can look at the code as well as perform their own tests if you want. 

Jira: https://issues.redhat.com/browse/AAP-43538

### DIFF
--- a/src/aap_eda/tasks/analytics.py
+++ b/src/aap_eda/tasks/analytics.py
@@ -14,6 +14,7 @@
 
 import logging
 
+from ansible_base.lib.utils.db import advisory_lock
 from flags.state import flag_enabled
 
 from aap_eda.analytics import collector, utils
@@ -32,22 +33,36 @@ def schedule_gather_analytics(queue_name: str = ANALYTICS_TASKS_QUEUE) -> None:
         return
     interval = utils.get_analytics_interval()
     logger.info(f"Schedule analytics to run in {interval} seconds")
-    tasking.enqueue_delay(
-        queue_name,
-        ANALYTICS_SCHEDULE_JOB_ID,
-        interval,
-        auto_gather_analytics,
-    )
+    with advisory_lock(ANALYTICS_SCHEDULE_JOB_ID, wait=False) as acquired:
+        if not acquired:
+            logger.debug(
+                "Another instance of schedule_gather_analytics"
+                " is already running",
+            )
+            return
+
+        tasking.enqueue_delay(
+            queue_name,
+            ANALYTICS_SCHEDULE_JOB_ID,
+            interval,
+            auto_gather_analytics,
+        )
 
 
-def auto_gather_analytics(queue_name: str = ANALYTICS_TASKS_QUEUE) -> None:
-    gather_analytics(queue_name)
+def auto_gather_analytics() -> None:
     schedule_gather_analytics()
+    gather_analytics()
 
 
-def gather_analytics(queue_name: str = ANALYTICS_TASKS_QUEUE) -> None:
-    logger.info("Queue EDA analytics")
-    tasking.unique_enqueue(queue_name, ANALYTICS_JOB_ID, _gather_analytics)
+def gather_analytics() -> None:
+    logger.info("Running EDA gather analytics")
+    with advisory_lock(ANALYTICS_JOB_ID, wait=False) as acquired:
+        if not acquired:
+            logger.debug(
+                "Another instance of gather analytics is already running"
+            )
+            return
+        _gather_analytics()
 
 
 def _gather_analytics() -> None:
--- a/src/aap_eda/tasks/orchestrator.py
+++ b/src/aap_eda/tasks/orchestrator.py
@@ -20,6 +20,7 @@ from datetime import datetime, timedelta
 from typing import Optional
 
 import django_rq
+from ansible_base.lib.utils.db import advisory_lock
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 
@@ -42,6 +43,11 @@ from aap_eda.services.activation.activation_manager import (
 )
 
 from .exceptions import UnknownProcessParentType
+
+# Wrap the django_rq job decorator so its processing is within our retry
+# code.
+job = tasking.redis_connect_retry()(django_rq.job)
+
 
 LOGGER = logging.getLogger(__name__)
 
@@ -72,6 +78,26 @@ def get_process_parent(
 
 
 def _manage(process_parent_type: str, id: int, request_id: str = "") -> None:
+    """Wrap the _manage_no_lock function.
+
+    It ensures that only one task is executed at a time.
+    """
+    with advisory_lock(
+        _manage_process_job_id(process_parent_type, id), wait=False
+    ) as acquired:
+        if not acquired:
+            LOGGER.debug(
+                f"Method _manage({process_parent_type}, {id}) "
+                "already being ran, exiting",
+            )
+            return
+
+        _manage_no_lock(process_parent_type, id, request_id)
+
+
+def _manage_no_lock(
+    process_parent_type: str, id: int, request_id: str = ""
+) -> None:
     """Manage the activation with the given id.
 
     It will run pending user requests or monitor the activation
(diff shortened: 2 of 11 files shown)


src/aap_eda/tasks/analytics.py
```diff
@@ -32,22 +33,36 @@ def schedule_gather_analytics(queue_name: str = ANALYTICS_TASKS_QUEUE) -> None:
...
+            interval,
+            auto_gather_analytics,
+        )
 
-def auto_gather_analytics(queue_name: str = ANALYTICS_TASKS_QUEUE) -> None:
-    gather_analytics(queue_name)
+
+def auto_gather_analytics() -> None:
     schedule_gather_analytics()
+    gather_analytics()
```
Alex-Izquierdo, line 54: @bzwei @hsong-rh Note here that I have changed the command because there is no more enqueueing of the function. All is executed each time, so first we schedule the next execution of the gatherer and later run it, otherwise if gather_analytics fails for some reason, the next execution would not be scheduled.

src/aap_eda/tasks/orchestrator.py
```diff
@@ -72,6 +78,26 @@ def get_process_parent(
 
 
 def _manage(process_parent_type: str, id: int, request_id: str = "") -> None:
+    """Wrap the _manage_no_lock function.
+
+    It ensures that only one task is executed at a time.
+    """
+    with advisory_lock(
```
bzwei, line 85: it seems we could create a common method for the wrapping logic to check acquired, log debug message, or execute the actual method.

src/aap_eda/tasks/orchestrator.py
```diff
@@ -432,6 +475,7 @@ def start_rulebook_process(
         ActivationRequest.START,
         request_id,
     )
+    monitor_rulebook_processes.delay()
```
bzwei, line 478: Do we really need to start a monitor immediately after each user request? Although it does hurt, it also won't help since the request just gets to be added to the internal queue.

Alex-Izquierdo: This is the same approach that we already took in the weboscket. It will be dispatched in the monitor, but scheduling it may accelerate the process. In the best case we gain 5 seconds, but I think it does not hurt either. 
Do you want to remove it?

src/aap_eda/tasks/project.py
```diff
@@ -43,24 +58,45 @@ def import_project(project_id: int):
...
 
-    project = models.Project.objects.get(pk=project_id)
-    try:
-        ProjectImportService().sync_project(project)
-    except ProjectImportError as e:
-        logger.error(e, exc_info=settings.DEBUG)
 
-    logger.info(f"Task complete: Sync project ( project_id={project.id} )")
+def _sync_project(project_id: int):
+    with advisory_lock(f"import_project_{project_id}", wait=False) as acquired:
```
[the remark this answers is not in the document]
Alex-Izquierdo: yeah! good catch! let me fix it. thanks.

src/aap_eda/tasks/orchestrator.py
AlanCoding: That would be fine with me. We just need a clear grasp on what constitutes uniqueness. I expect this would be the method name and the arguments. To avoid args & kwargs confusion, we would ideally want this contract:

```python
def _manage(process_parent_type: str, id: int, /, *, request_id: str = "") -> None:
```

And we could have a decorator that would make it unique on
 - name
 - process_parent_type (AFAIK will always be the same)
 - id

```python
@advisory_unique
def _manage(process_parent_type: str, id: int, /, *, request_id: str = "") -> None:
```

The decorator could be added and tested in DAB with the rest of it. @Alex-Izquierdo did point out a library we could use that would be better than the runtime setting changes there, and I agree.

https://github.com/ansible/django-ansible-base/blob/f93c8aac04396ee8f2e54da415b3a68d3e46aee1/ansible_base/lib/utils/db.py

src/aap_eda/tasks/orchestrator.py
bzwei: Sorry. My original comment meant "it does *not* hurt". I am thinking by the time the monitor is running the task hasn't finished or even started yet. Sure it does not hurt.

chrismeyersfsu:
FYI maximum advisory locks `max_locks_per_transaction * (max_connections + max_prepared_transactions) =~ 64 * (1,024 + 0)` I _think_ the locks here are short lived, which is good.

src/aap_eda/tasks/project.py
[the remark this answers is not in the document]
Alex-Izquierdo: removed, @bzwei

src/aap_eda/tasks/orchestrator.py
Alex-Izquierdo: I would like to think carefully of the contract for such a shared utility and if it has real value setting the uniqueness from the positional arguments (clearly not for kwargs) or it is just an eventual coincidence only applicable to eda.

src/aap_eda/tasks/orchestrator.py
Alex-Izquierdo: >I am thinking by the time the monitor is running the task hasn't finished or even started yet. Sure it does not hurt.
Well, that's where the lock takes place. Anyway it would not be different from what we are already doing it in https://github.com/ansible/eda-server/pull/1262/files#diff-179935aa48d00084c8f2722f9b007cad6d09b24425ffe4121cc75f6c37b03c42R214

I will keep it if you agree.

src/aap_eda/tasks/orchestrator.py
```diff
@@ -460,12 +505,7 @@ def delete_rulebook_process(
         ActivationRequest.DELETE,
         request_id,
     )
-    dispatch(
-        process_parent_type,
-        process_parent_id,
-        ActivationRequest.DELETE,
-        request_id,
-    )
+    monitor_rulebook_processes.delay()
```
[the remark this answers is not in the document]
Alex-Izquierdo: I think it can be a good to have, sure. For the implementation of the feature flag right now I don't think it is a big deal. 
I'm saying it because I have seen you assigned to me the issue and I'm not sure If I will be able to work on it in the shortest term :)
```
def function()
   ...
if feature flag:
    function = task()(function)
```

src/aap_eda/tasks/orchestrator.py
Alex-Izquierdo: @AlanCoding @bzwei 

I've been looking into creating a decorator to encapsulate the advisory lock logic, but there are too many edge cases that would require passing a large number of parameters to the decorator to make it flexible enough.

For example, in the case of projects, we want both the sync and import operations to share the same lock ID, which means we'd need to support a customizable prefix. On the other hand, the orchestrator checks for the lock ahead of time before dispatching the method to a queue, which introduces yet another variation.

Enforcing the lock ID to be derived from positional arguments would restrict the caller’s ability to customize the lock. Eventually, we might need to pass additional parameters to advisory_lock, making the decorator even more complex.

Additionally, the approach would require inspecting the wrapped function to extract and process its arguments (e.g., via inspect.signature), which introduces an extra layer of complexity and runtime overhead. 

At that point, it feels like overengineering just to reuse a log message, and we’d lose a lot of flexibility in the process. In my opinion, it’s reasonable to continue using the advisory lock explicitly where needed, rather than forcing it through an abstraction.

bzwei:
Tested with eda-server image (quay.io/aizquier/eda-server:advisory) from this PR on OCP deployment and it worked.

bzwei (approved):

hsong-rh (approved):

PR #1264, opened 2025-04-09
add wait for pod deletion method

ShaiahWren:
[AAP-41447](https://issues.redhat.com/browse/AAP-41447)

The current Kubernetes 'cleanup' method has a weakness in that it finishes before ensuring that the pod at hand - and also its container resources - are stopped and deleted. Because the delete methods that _cleanup()_ employs run asynchronously, the method may (and has) returned whilst the pod is still running. Herein lies the race condition described in the ticket, where unexpected behavior can occur. 

The solution is to create a _wait_for_pod_deletion(pod_id) method that uses the Kubernetes Watch to observe the stream of events from the kubernetes API, filtered on the pod at hand, that waits until the event type has the value of 'DELETED'. After this method has successfully completed, the cleanup is safe to continue and return.

Screenshot 2025-04-11 at 6 07 20 PM

### DIFF
--- a/src/aap_eda/services/activation/engine/kubernetes.py
+++ b/src/aap_eda/services/activation/engine/kubernetes.py
@@ -12,7 +12,6 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-
 import base64
 import json
 import logging
@@ -23,6 +22,7 @@ from dateutil import parser
 from kubernetes import client as k8sclient, config, watch
 from kubernetes.client.rest import ApiException
 from kubernetes.config.config_exception import ConfigException
+from rest_framework import status
 
 from aap_eda.core.enums import ActivationStatus
 from aap_eda.services.activation.engine.exceptions import (
@@ -55,6 +55,7 @@ POD_FAILED_REASONS = [
     IMAGE_PULL_BACK_OFF,
     IMAGE_PULL_ERROR,
 ]
+POD_DELETE_TIMEOUT = 60
 
 
 @dataclass
@@ -114,6 +115,7 @@ class Engine(ContainerEngine):
         # f"{request.id}-{uuid.uuid4()}" #noqa: E800
         try:
             log_handler.write("Creating Job")
+            log_handler.write(f"Log tracking id: {request.log_tracking_id}")
             log_handler.write(f"Image URL is {request.image_url}", flush=True)
             self._create_job(request, log_handler)
             LOGGER.info("Waiting for pod to start")
@@ -197,6 +199,7 @@ class Engine(ContainerEngine):
         self._delete_services(log_handler)
         self._delete_job(log_handler)
 
+        self._wait_for_pod_to_be_deleted(container_id, log_handler)
         log_handler.write(f"Job {container_id} is cleaned up.", flush=True)
 
     def update_logs(self, container_id: str, log_handler: LogHandler) -> None:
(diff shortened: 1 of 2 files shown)


src/aap_eda/services/activation/engine/kubernetes.py
```diff
@@ -12,15 +12,14 @@
...
-
 import base64
 import json
 import logging
 from dataclasses import dataclass
 from datetime import datetime, timezone
 
 from dateutil import parser
-from kubernetes import client as k8sclient, config, watch
+from kubernetes import client, client as k8sclient, config, watch
```
Alex-Izquierdo, line 22: Why reimport client? you should use self.client

src/aap_eda/services/activation/engine/kubernetes.py
```diff
@@ -188,6 +187,37 @@ def get_status(self, container_id: str) -> ContainerStatus:
...
+        w = watch.Watch()
+
+        log_handler.write(
+            f"Waiting for pod '{pod_name}' to be deleted...", flush=True
+        )
+        try:
+            for event in w.stream(
+                core_v1.list_namespaced_pod,
+                namespace=namespace,
+                field_selector=f"metadata.name={pod_name}",
```
Alex-Izquierdo, line 206: We should add a timeout and catch the exception.

src/aap_eda/services/activation/engine/kubernetes.py
```diff
@@ -188,6 +187,37 @@ def get_status(self, container_id: str) -> ContainerStatus:
...
+            ):
+                if event["type"] == "DELETED":
+                    # Pod successfully deleted
+                    w.stop()
+                    log_handler.write(
+                        f"Pod '{pod_name}' has been deleted.", flush=True
+                    )
+                    return
+        except Exception as e:
```
Alex-Izquierdo, line 215: we should catch the ApiException of the client

src/aap_eda/services/activation/engine/kubernetes.py
```diff
@@ -188,6 +187,46 @@ def get_status(self, container_id: str) -> ContainerStatus:
...
+
+        log_handler.write(
+            f"Waiting for pod '{pod_name}' to be deleted...", flush=True
+        )
+        try:
+            for event in w.stream(
+                core_v1.list_namespaced_pod,
+                namespace=self.namespace,
+                field_selector=f"metadata.name={pod_name}",
+                timeout_seconds=60,
```
mkanoor, line 206: @ShaiahWren should we make the 60 a constant at the top of the file?

src/aap_eda/services/activation/engine/kubernetes.py
```diff
@@ -188,6 +187,46 @@ def get_status(self, container_id: str) -> ContainerStatus:
...
+            ):
+                if event["type"] == "DELETED":
+                    # Pod successfully deleted
+                    w.stop()
+                    log_handler.write(
+                        f"Pod '{pod_name}' has been deleted.", flush=True
+                    )
+                    return
+        except ApiException as e:
+            if e.status == 404:
```
mkanoor, line 216: @ShaiahWren Can this be replaced with the **status.HTTP_404_NOT_FOUND**

mkanoor (approved):

hsong-rh (approved):

PR #1265, opened 2025-04-10
chore: pin ansible-core in Dockerfile

ttuffin:
In light of upcoming changes in ansible-core, pinning the version to 2.16 while we investigate the impact.

[AAP-43723](https://issues.redhat.com/browse/AAP-43723)

https://github.com/ansible/ansible/pull/84621

### DIFF
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -5,6 +5,7 @@ ARG STATIC_ROOT=/var/lib/eda/static
 
 # python3.11 and python3.12 supported
 ARG PYTHON_BIN="python3.11"
+ARG ANSIBLE_CORE_VER=2.16
 
 RUN useradd --uid "$USER_ID" --gid 0 --home-dir /app --create-home eda \
     && mkdir -p /app/.local /var/lib/eda/ \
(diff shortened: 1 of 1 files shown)


kaiokmo (approved):

tools/docker/Dockerfile
```diff
@@ -57,7 +57,7 @@ RUN poetry install -E all --no-root --no-cache \
     && ${PYTHON_BIN} -m pip install gunicorn
 COPY . $SOURCES_DIR/
 RUN poetry install -E all --only-root
-RUN pip install ansible-core
+RUN pip install ansible-core==2.16
```
Alex-Izquierdo, line 60: Can we define a build arg with this version as default, so we can ship other versions dynamically? There will be a point when stable branches would have as default different versions

ttuffin: updated

tools/docker/Dockerfile
```diff
@@ -5,6 +5,7 @@ ARG STATIC_ROOT=/var/lib/eda/static
 
 # python3.11 and python3.12 supported
 ARG PYTHON_BIN="python3.11"
+ARG ANSIBLE_CORE_VER=${ANSIBLE_CORE_VER:-2.16}
```
Alex-Izquierdo, line 8: Build args are overridable by default. 
```suggestion
ARG ANSIBLE_CORE_VER=2.16
```

Alex-Izquierdo (changes requested):

tools/docker/Dockerfile
ttuffin: The idea was to allow it to be set via env var, like we do with USER_ID

Alex-Izquierdo (approved):

zkayyali812 (approved):

PR #1268, opened 2025-04-15
update variables for collection auth to be standard

sean-m-sullivan:
Update variables for collection auth to be standard. 

https://issues.redhat.com/browse/ANSTRAT-1181

### DIFF
--- a/src/aap_eda/core/management/commands/create_initial_data.py
+++ b/src/aap_eda/core/management/commands/create_initial_data.py
@@ -315,7 +315,7 @@ AAP_INJECTORS = {
         "aap_password": "{{password}}",
         "aap_token": "{{oauth_token}}",
         "aap_request_timeout": "{{request_timeout}}",
-        "aap_verify_ssl": "{{verify_ssl}}",
+        "aap_validate_certs": "{{verify_ssl}}",
     },
     "env": {
         "TOWER_HOST": "{{host}}",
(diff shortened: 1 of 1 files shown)


zkayyali812:
The following PR should fix the e2e test here - https://github.com/ansible/eda-qa/pull/466

zkayyali812:
I deployed a custom e2e run here to validate the update to our tests as well - 
https://github.com/ansible/eda-server/actions/runs/14479621253

zkayyali812 (approved):

mkanoor:
@zkayyali812 @sean-m-sullivan How do we handle migration of the existing Credentials, we have customers who might already be using the RH AAP Credential, since we are updating the Credential Type injectors here we need to account for how migrations are going to be handled. The Credential Type cannot be updated till all Credentials are removed in the regular UI flow. Moreover this particular Credential Type has some inert injector fields which doesn't get used by ansible-rulebook. So can we delay this till we have a Credential Service. When we started the Credentials in EDA we just copied the schema from the AAP controller so that we are not different, most of the Env vars in the injectors don't get used by EDA. Currently we dont have the EDA Server talking directly to the AAP controller, we only have ansible-rulebook that talks to AAP with the RH AAP Credential.

zkayyali812:
Closing this PR after discussion with @sean-m-sullivan and @mkanoor. This is no longer needed, and we can handle this functionality by just updating the collection.

PR #1269, opened 2025-04-18
[AAP-43742] fix: auto generate service name. Do not throw error

bzwei:
Enhance `create_k8s_service_name(activation_nam)` to accept any activation name

[AAP-43742](https://issues.redhat.com/browse/AAP-43742)

For k8s deployment we always prepare a service name in case it is needed for the rulebook. The current implementation of `create_k8s_service_name(activation_nam)` converts some non rfc_1035 compliant letters such as dot and _ into -, but rejects other non-qualified letters. This prevents the users from creating activation with a name containing such letters even if the service is not eventually needed.
The solution is to always create a rfc_1035_compliant service name. Never reject an activation name. Activation name is used only as a reference to create the service name. No special requirements should be imposed.

### DIFF
--- a/.github/workflows/e2e-dispatch.yml
+++ b/.github/workflows/e2e-dispatch.yml
@@ -72,7 +72,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.9"
+          python-version: "3.11"
 
       - name: Install dependencies
         working-directory: ${{ env.EDA_QA_PATH }}
@@ -154,7 +154,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v4
         with:
-          python-version: "3.9"
+          python-version: "3.11"
 
       - name: Install dependencies
         working-directory: ${{ env.EDA_QA_PATH }}
--- a/src/aap_eda/api/serializers/activation.py
+++ b/src/aap_eda/api/serializers/activation.py
@@ -48,10 +48,7 @@ from aap_eda.core import models, validators
 from aap_eda.core.enums import DefaultCredentialType, ProcessParentType
 from aap_eda.core.exceptions import ParseError
 from aap_eda.core.utils.credentials import get_secret_fields
-from aap_eda.core.utils.k8s_service_name import (
-    InvalidRFC1035LabelError,
-    create_k8s_service_name,
-)
+from aap_eda.core.utils.k8s_service_name import create_k8s_service_name
 from aap_eda.core.utils.rulebook import (
     build_source_list,
     get_rulebook_hash,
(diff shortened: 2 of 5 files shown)


mkanoor:
@bzwei 
We have two distinct fields called **service_name** and **name**. The user can provide the 2 properties separately. 
* When the service name is provided we have to ensure that the service name is RFC1035 compliant. So the RFC 1035 compliant check is needed we can't remove it.

* When the service name is not provided we have to auto generate the service name from the name. Thats where we have the issue when the name is [] 2 square brackets thats an invalid charachter so we try to convert it to a RFC1035 compliant name and can't create one.

If the user had
* name : []
* service_name: my_service_name
it works 

but if the user has

* name: []

Since we can't auto generate we need help from the customer to specify the name, if we picked up any name for this corner case it might not work all the time.

@hsong-rh wdyt?

Then our service name generator fails.

hsong-rh:
@mkanoor Yes, that's the backend logic for the k8s deployment only. We only check for RFC1035 compliance on the name when the service name is not specified. I agree with your suggestion.

bzwei:
@mkanoor @hsong-rh When the user explicitly provides a service name, we strictly validate whether it is rfc 1035 compliant.  The creation will fail with error "{service_name} must be a valid RFC 1035 label name". We don't convert it.

This fix is for the case when the user does NOT provide a service name, we auto generate a service name based on the activation name. When the activation name is not RFC 1035 compliant, the old code does partial conversion. It converts some invalid characters, but rejects others and raises an error. This fix unifies the user experience, i.e, the user can provide an activation name without restriction, we generate a compliant service name (when it is not given) based on the activation name. We never reject an activation name because it is not RFC 1035 compliant.

Why we don't force the user to provide a service name? First the service name may not be actually needed depending on the rulebook. Second it is not very important. The user will see the generated service name after the creation. Third the old code already does a partial conversion. The generated service name is not guaranteed to be equal to the activation name.

bzwei:
I added a handling for special case where there is no character from the activation name can be used for service. We then generate a service name using uuid. It is not user friendly to force them to manually correct the problem because the service name may not be actually needed.

mkanoor:
@bzwei @hsong-rh We also have this 
```
ALWAYS SAFE_PLUGINS_FOR_PORT_FORWARD: ['ansible.eda.webhook', 'ansible.eda.alertmanager']
```
If the rulebook contains one of these sources we need to open ports and need a service for all other cases a service might not be necessary. We also would have to check if the EventStreams are swapping one of the sources in the rulebook.

We cannot enforce service_name as a mandatory field since we don't know if its needed. Most of these predates the EventStreams so I am not sure if the service name is still being used. The most prominent use case was ansible.eda.webook.

bzwei:
This fix is following the existing pattern. We do not enforce service_name as a mandatory but generate a safe name in case it is needed. We avoid the work to check whether a service name is needed by analyzing the content of a rulebook and the use of event stream.

> @bzwei @hsong-rh We also have this
> 
> ```
> ALWAYS SAFE_PLUGINS_FOR_PORT_FORWARD: ['ansible.eda.webhook', 'ansible.eda.alertmanager']
> ```
> 
> If the rulebook contains one of these sources we need to open ports and need a service for all other cases a service might not be necessary. We also would have to check if the EventStreams are swapping one of the sources in the rulebook.
> 
> We cannot enforce service_name as a mandatory field since we don't know if its needed. Most of these predates the EventStreams so I am not sure if the service name is still being used. The most prominent use case was ansible.eda.webook.

hsong-rh (approved):

mkanoor (approved):

PR #1273, opened 2025-04-24
fix: [AAP-43898] update status endpoint response to DAB enum

Dostonbek1:
Change the response from `/v1/status/` endpoint to use enum provided by DAB.

JIRA: [AAP-43898](https://issues.redhat.com/browse/AAP-43898)

### DIFF
--- a/src/aap_eda/core/views.py
+++ b/src/aap_eda/core/views.py
@@ -12,6 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+from ansible_base.lib.constants import STATUS_FAILED, STATUS_GOOD
 from django.db import connection
 from django.db.utils import OperationalError
 from drf_spectacular.utils import OpenApiResponse, extend_schema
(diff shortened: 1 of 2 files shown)


src/aap_eda/core/views.py
```diff
@@ -12,6 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+from ansible_base.lib.constants import STATUS_GOOD
```
hsong-rh, line 15: Can you also import STATUS_FAILED? It should be used on https://github.com/ansible/eda-server/pull/1273/files#diff-b7f4c079e5cb3a3c01996199db0ee6474eb5618e60d85a85ac2a08795268908aL60

Dostonbek1: Updated. Thanks.

hsong-rh (approved):

dhaustein (approved):
LGTM FR FR

PR #1274, opened 2025-04-24
fix: change error field from rulebook to source_mappings

bzwei:
The error "Rulebook has changed since the sources were mapped. Please reattach event streams" is caused by validating field source_mappings.

[AAP-44583](https://issues.redhat.com/browse/AAP-44583)

This fix changes the validate error message of the 404 error body from key rulebook to source_mappings. The key must match the actual field in the creation/update body so that the UI can correctly highlight the field for user to fix error.

### DIFF
--- a/src/aap_eda/api/constants.py
+++ b/src/aap_eda/api/constants.py
@@ -14,4 +14,4 @@
 
 # EDA_SERVER_VAULT_LABEL is reserved for system vault password identifiers
 EDA_SERVER_VAULT_LABEL = "EDA_SERVER"
-SOURCE_MAPPING_ERROR_KEY = "rulebook"
+SOURCE_MAPPING_ERROR_KEY = "source_mappings"


hsong-rh:
@bzwei SOURCE_MAPPING_ERROR_KEY is a kind of anchor; UI uses it to decide where the error messages are displayed. If you change it to `source_mappings`, the UI must accept it. Otherwise, all these messages will be detached from UI fields.

bzwei:
Yes. I am working with Laura for the bug. The original key rulebook could not help the UI to display the error. If we change it to rulebook_id, the error will show under the rulebook box. The problem is not the rulebook but the mappings. So the desired place to show the error is under the event streams box.
> @bzwei SOURCE_MAPPING_ERROR_KEY is a kind of anchor; UI uses it to decide where the error messages are displayed. If you change it to `source_mappings`, the UI must accept it. Otherwise, all these messages will be detached from UI fields.

hsong-rh (approved):

PR #1276, opened 2025-04-28
fix: reschedule analytics when interval changes in credential

bzwei:
When the gather interval in the analytics credential changes, before the fix eda-server will wait until the scheduled analytics based on the old interval is executed, and then schedule the next job with the new interval. The fix immediately cancels the scheduled job and reschedules the next job to run at the new interval

[AAP-44570](https://issues.redhat.com/browse/AAP-44570)

### DIFF
--- a/src/aap_eda/tasks/analytics.py
+++ b/src/aap_eda/tasks/analytics.py
@@ -14,6 +14,7 @@
 
 import logging
 
+import django_rq
 from ansible_base.lib.utils.db import advisory_lock
 from flags.state import flag_enabled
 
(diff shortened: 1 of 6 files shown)


src/aap_eda/tasks/analytics.py
```diff
@@ -27,20 +28,30 @@
...
-    with advisory_lock(ANALYTICS_SCHEDULE_JOB_ID, wait=False) as acquired:
+    with advisory_lock(ANALYTICS_SCHEDULE_JOB_ID, wait=cancel) as acquired:
         if not acquired:
             logger.debug(
                 "Another instance of schedule_gather_analytics"
                 " is already running",
             )
             return
 
+        if cancel:
```
mkanoor, line 51: @bzwei What would happen if the interval is set to a low value like 2 seconds and if the gather job doesn't finish in 2 seconds will we end up killing the old running job. The next job would exhibit the same behavior and we will never get the gather task to finish successfully. Should we have a setting called
MIINIMUM_GATHER_TASK_TIME and ensure that the user cannot change the interval to anything less than MIINIMUM_GATHER_TASK_TIME

bzwei: @mkanoor A running job continues to run to finish, never be killed. I proved this by doing real tests.

hsong-rh (approved):

PR #1279, opened 2025-05-06
chore: remove replicas for docker-compose-dev

Alex-Izquierdo:
The use of replicas has some inconveniences and limitations. It does not allow to map ports if we need. It is not always compatible with other consumers of our compose file. This has been particularly relevant for the development of the integration with dispatcherd. This can be also considered one of the PR's part of the integration/onboarding of dispatcherd. 

This PR removes the replicas config and creates a second instance of the workers. This to make sure we develop and test against scenarios of redundancy where other race conditions can happen. It is necessary for dispatcherd but also valuable in general (which is why the default value of the replicas was 2)

### DIFF
--- a/tools/docker/docker-compose-dev.yaml
+++ b/tools/docker/docker-compose-dev.yaml
@@ -169,10 +169,8 @@ services:
       - '../../:/app/src:z'
     restart: always
 
-  eda-default-worker:
+  eda-default-worker-1:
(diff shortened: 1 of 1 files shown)


Dostonbek1 (approved):

bzwei (approved):

Alex-Izquierdo:
Needed also https://github.com/ansible/eda-qa/pull/469 for e2e multinode tests passing.

PR #1283, opened 2025-05-07
fix: [AAP-44969] Add Redis status check in project creating/syncing and endpoint

hsong-rh:
https://issues.redhat.com/browse/AAP-44969

This PR contains:
1. Add `redis_is_available` check when creating/syncing projects
2. Add the Redis status report in the endpoint `api/eda/v1/status`


Alex-Izquierdo (approved):
LGTM. What about the socket connection timeout? it is something that you are going to address in a different PR? Nothing of this is relevant if we hit the gunicorn timeout before the redis one.

hsong-rh:
> LGTM. What about the socket connection timeout? it is something that you are going to address in a different PR? Nothing of this is relevant if we hit the gunicorn timeout before the redis one.

@Alex-Izquierdo yes, I want to address it separately, after we test in an environment with this fix.

bzwei (approved):

PR #1284, opened 2025-05-07
Bug/remove container-aap 40077

B-Whitt:
Objectives for this bug [https://issues.redhat.com/browse/AAP-40077]:
- Update restart API endpoint such that, when a user attempts to restart, we detect if the deployment type is 'podman' and the 'activation-worker-node' is offline, we should block the restart and return an error warning the user that a running container will be orphaned.
- Enhance restart API endpoint with a force option where, if enabled, the restart will go through.

Activation containers are being left behind (orphaned) on nodes that go offline in an multi-node deployment. 

The user has no way of knowing they they are leaving behind these orphaned containers.

### DIFF
--- a/src/aap_eda/api/views/activation.py
+++ b/src/aap_eda/api/views/activation.py
@@ -16,6 +16,7 @@ import logging
 import redis
 from ansible_base.rbac.api.related import check_related_permissions
 from ansible_base.rbac.models import RoleDefinition
+from django.conf import settings
 from django.db import transaction
 from django.forms import model_to_dict
 from django_filters import rest_framework as defaultfilters
@@ -40,6 +41,7 @@ from aap_eda.tasks.orchestrator import (
     start_rulebook_process,
     stop_rulebook_process,
 )
+from aap_eda.utils import str_to_bool
 
 from .mixins import RedisDependencyMixin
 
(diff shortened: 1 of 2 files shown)


src/aap_eda/api/views/activation.py
```diff
@@ -476,15 +478,40 @@ def disable(self, request, pk):
                 None,
                 description="Activation not enabled.",
             ),
+            status.HTTP_409_CONFLICT: OpenApiResponse(
+                None, description="Decision Environment in use by Activations."
```
bzwei, line 482: description does not match

src/aap_eda/api/views/activation.py
```diff
@@ -476,15 +478,40 @@ def disable(self, request, pk):
...
+            settings.DEPLOYMENT_TYPE == "podman" and
+            activation.status == ActivationStatus.WORKERS_OFFLINE and not
+            force_delete
+        ):
+            # block the restart and return an error
+            raise api_exc.Conflict(
+                "An activation with an activation_status of 'Workers offline' "
+                "cannot be Restarted because this will leave an orphaned container "
+                "running on one of the 'activation-worker-node's. "
+                "If you want to force a restart, please add the /?force=True query param."
```
bzwei, line 513: should query param be `/?force=true` ?

src/aap_eda/api/views/activation.py
```diff
@@ -476,15 +478,40 @@ def disable(self, request, pk):
...
+                type=bool,
+            )
+        ],
     )
     @action(methods=["post"], detail=True, rbac_action=Action.RESTART)
     def restart(self, request, pk):
         activation = self.get_object()
-
         self._check_deleting(activation)
+        force_delete = str_to_bool(
```
bzwei, line 499: should the variable be `force_restart`?

src/aap_eda/api/views/activation.py
B-Whitt: updated comment

src/aap_eda/api/views/activation.py
B-Whitt: updated

PR #1286, opened 2025-05-07
wip: upd poetry, add dispatcher to analitics and orchestrator

Alex-Izquierdo:

### DIFF
--- a/src/aap_eda/tasks/analytics.py
+++ b/src/aap_eda/tasks/analytics.py
@@ -20,6 +20,8 @@ from flags.state import flag_enabled
 
 from aap_eda.analytics import collector, utils
 from aap_eda.core import tasking
+from flags.state import flag_enabled
+from django.conf import settings
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


src/aap_eda/tasks/analytics.py
```diff
@@ -60,9 +62,28 @@ def schedule_gather_analytics(
         )
 
 
-@job(ANALYTICS_TASKS_QUEUE)
 def reschedule_gather_analytics(
     queue_name: str = ANALYTICS_TASKS_QUEUE,
+) -> None:
+    """Reschedule the gather analytics job.
```
bzwei, line 68: Don't we need to do the same for `def schedule_gather_analytics`?
Also do we need to make line 35 `job = tasking.redis_connect_retry()(django_rq.job)` conditional based on the feature flag?

src/aap_eda/wsapi/consumers.py
```diff
@@ -209,9 +209,6 @@ def handle_heartbeat(self, message: HeartbeatMessage) -> None:
                 message.stats["ruleSetName"]
             ] = message.stats
             activation.save(update_fields=["ruleset_stats"])
-
```
bzwei, line 212: why delete this block.

src/aap_eda/tasks/analytics.py
Alex-Izquierdo: No, `schedule_gather_analytics` is agnostic, since it finally calls under the advisory_lock `tasking.enqueue_delay` which does the proxy between rq/dispatcherd. https://github.com/ansible/eda-server/pull/1286/files#diff-8919c4e8183e21876fcf7b5ad8884b1efbb37a0e08544fba092d83c6f733adfbR585

PR #1293, opened 2025-05-09
refactor: get feature flags from features module

bzwei:
Get whether a feature flag is on or off as a constant from the new features module.
It improves the code readability and slightly the performance as the flags are now cached.

This work is to prepare the merging of the dispatcherd feature which is guarded by a flag.

### DIFF
--- a/src/aap_eda/settings/features.py
+++ b/src/aap_eda/settings/features.py
@@ -1,0 +1,46 @@
+#  Copyright 2025 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


Alex-Izquierdo (approved):

src/aap_eda/settings/features.py
```diff
@@ -0,0 +1,37 @@
...
+
+from typing import Any
+
+from django.conf import settings
+from flags.state import flag_enabled
+
+DISPATCHERD: bool
+ANALYTICS: bool
+
+_caches = {}
```
ttuffin, line 23: Should this be the singular form, cache? Maybe a more descriptive name `_feature_flag_cache`

src/aap_eda/settings/features.py
```diff
@@ -0,0 +1,37 @@
...
+
+from typing import Any
+
+from django.conf import settings
+from flags.state import flag_enabled
+
+DISPATCHERD: bool
+ANALYTICS: bool
+
+_feature_flag_cache = {}
```
Alex-Izquierdo, line 23: Suggestion: no need to implement custom cache, you can use https://docs.python.org/3/library/functools.html#functools.cache

ttuffin (approved):

Alex-Izquierdo (approved):

PR #1295, opened 2025-05-12
chore: add helper to sanitize pg identifiers

Alex-Izquierdo:
For dispatcherd we need to sanitize the channel name as we already do for event streams. There is no need to fill the code with `.replace()` calls. 
Instead create a helper that can sanitize the name in a more reusable and reliable way.

### DIFF
--- a/src/aap_eda/utils/__init__.py
+++ b/src/aap_eda/utils/__init__.py
@@ -47,18 +47,26 @@ def sanitize_postgres_identifier(identifier: str) -> str:
     if not identifier:
         raise ValueError("Identifier cannot be empty.")
 
-    # Replace invalid characters with underscores
-    sanitized = re.sub(r"\W", "_", identifier)
+    if len(identifier) > max_identifier_length:
+        raise ValueError(
+            f"Identifier exceeds {max_identifier_length} characters."
+        )
+
+    sanitized = identifier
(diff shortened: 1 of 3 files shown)


mkanoor (approved):

src/aap_eda/utils/__init__.py
```diff
@@ -32,3 +34,31 @@ def get_package_version(package_name: str) -> str:
...
+
+    # Replace invalid characters with underscores
+    sanitized = re.sub(r"\W", "_", identifier)
+
+    # Ensure it starts with a valid character
+    if not re.match(r"[A-Za-z_]", sanitized[0]):
+        sanitized = f"_{sanitized}"
+
+    # Ensure length
+    if len(sanitized) > max_identifier_length:
```
bzwei, line 58: if initially the length of identifier is happened to be max_identifier_length, but we may add a prefix `_` to it to make the length to exceed the limit, which is not expected.

Alex-Izquierdo: Handled the case. Let me know if it looks good to you now

bzwei (approved):

Dostonbek1 (approved):

mkanoor (approved):

PR #1297, opened 2025-05-14
fix: Check queue name of worker before applying missing policy

mkanoor:
Trying to fix a race condition when the monitor triggers a missing policy if the container cannot be found

Checking if the queue name in the instance matches the nodes queue name

To avoid a race condition that ends up terminating the running container

Intermittent test failure

No

Running the eda-qa test suite.

https://issues.redhat.com/browse/AAP-45421

### DIFF
--- a/src/aap_eda/services/activation/activation_manager.py
+++ b/src/aap_eda/services/activation/activation_manager.py
@@ -906,7 +906,17 @@ class ActivationManager(StatusManager):
         # Activations in running status must have a container
         # This case prevents cases when the container is externally deleted
         if container_status is None:
-            self._missing_container_policy()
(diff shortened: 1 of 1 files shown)


Alex-Izquierdo (commented):
I'm really concerned about this change. 
If the monitor is running in a wrong queue, that means the root issue is in the queue_dispatcher which is scheduling the monitor in wrong queue. I don't think this is a fix, is more a mitigation. 

At the very least, we should have a comment saying that this is a temporal workaround or something like that. 

I feel we should not do this, we should address the real issue.

Alex-Izquierdo:
After several tests we can conclude that https://github.com/ansible/eda-server/pull/1262 already solves the issue.

PR #1306, opened 2025-05-16
fix: added missing settings import

Alex-Izquierdo:
we call settings.DEPLOYMENT_TYPE in restart, but settings are not imported. I'm not sure how this line can be lost, because I was testing also restart and it should have failed. But anyway, better later than never.


Alex-Izquierdo:
It was this PR, for some reason, I didn't fetched the import line when rebasing. https://github.com/ansible/eda-server/pull/1284/files

PR #1324, opened 2025-06-02
fix: [AAP-46563]  solve potential race condition issue when disabling activations

hsong-rh:
Move the status update from API to Activation workers to avoid race conditions in activation disabling.

https://issues.redhat.com/browse/AAP-46563


Alex-Izquierdo (commented):
IMO, `is_enabled` should not be domain of the activation_manager because it is a property of the activation as template, not part of the lifecycle of the activation. is_enabled triggers the stop operation, that's all. 

Without reading it in deep, this change seems a stronger change and it might be the cause of the failed e2e tests.

I think the fix is much more simpler and It is what I understood it would be from our previous conversation.  In the view we set `activation.status = ActivationStatus.STOPPING` only when the activation is already running (starting or running status). No need of migrations or changing the manager/orchestrator which already has the right logic.

Alex-Izquierdo (approved):

PR #1332, opened 2025-06-03
fix: Run noop loop in scheduler under dispatcherd flag

Alex-Izquierdo:
This PR changes the approach to handle the dispatcherd flag in the scheduler. Instead or returning an error and creating noise for users, that also need to modify their deployments to not deploy the scheduler which would be restarting all the time in default deployments,  it applies a less disruptive approach by just warning about it and running a noop loop.

### DIFF
--- a/src/aap_eda/core/management/commands/scheduler.py
+++ b/src/aap_eda/core/management/commands/scheduler.py
@@ -70,6 +70,8 @@ For a complete list of parameters refer to the
 https://github.com/rq/rq-scheduler/blob/master/README.rst
 """
 import logging
+import signal
+import time
 import typing
 from datetime import datetime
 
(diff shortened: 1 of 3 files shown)


Dostonbek1 (approved):

src/aap_eda/core/management/commands/scheduler.py
```diff
@@ -167,15 +170,27 @@ class Command(rqscheduler.Command):
...
+
+            def handle_exit(signum, frame):
+                raise GracefulExit()
+
+            signal.signal(signal.SIGTERM, handle_exit)
+            signal.signal(signal.SIGINT, handle_exit)
+
+            try:
+                while True:
+                    time.sleep(60)
```
bzwei, line 190: is a much larger number better?

PR #1334, opened 2025-06-03
test: fix flaky tests caused by multithreaded tests

Alex-Izquierdo:
We have sporadic unexpected failures in tests. I think this is caused by `test_job_uniqueness` which uses multithreading and seems to interfere intermittently with other mocks. Since I could not found a way to avoid it or a better way to do this test (which I think right now is clean and reliable except for the mock interference) I think the best way is to just keep them separately, which is a good and common pattern in general. 

It marks the test with a new marker "multithreaded" which is never executed by default. Then CI stuff is updated to execute both, default and multithreaded, separately. 

Depends on https://github.com/ansible/eda-server/pull/1333

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -99,10 +99,26 @@ jobs:
       - name: Check migrations are up to date
         run: poetry run /usr/bin/env aap-eda-manage makemigrations --dry-run --check
 
-      - name: Run tests
+      - name: Run default tests
         run: |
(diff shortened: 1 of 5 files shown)


ttuffin (approved):

kaiokmo (approved):

kaiokmo:
@Alex-Izquierdo now that #1333 is merged, should we merge this one too?

ptoscano:
> @Alex-Izquierdo now that #1333 is merged, should we merge this one too?

Yeah, although there are conflicts to solve.

PR #1335, opened 2025-06-03
fix: add more validations to project attributes

bzwei:
Now validate project url, branch, and refspec

https://issues.redhat.com/browse/AAP-47166

### DIFF
--- a/src/aap_eda/services/project/scm.py
+++ b/src/aap_eda/services/project/scm.py
@@ -71,6 +71,11 @@ if RUNNER_COMMAND is None:
     raise ExecutableNotFoundError("Cannot find ansible-runner executable")
 
 
+GIT_COMMAND = shutil.which("git")
+if GIT_COMMAND is None:
+    raise ExecutableNotFoundError("Cannot find git executable")
+
+
 class ScmRepository:
     """Represents a SCM repository."""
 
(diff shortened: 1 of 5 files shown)


src/aap_eda/services/project/scm.py
```diff
@@ -322,3 +327,32 @@ def __call__(
                     raise ScmAuthenticationError(err_msg)
                 raise ScmError(f"{self.ERROR_PREFIX} {err_msg}")
             raise ScmError(f"{self.ERROR_PREFIX} {outputs.getvalue().strip()}")
+
+
+URL_REGEX = "^((git|ssh|http(s)?)|git\\+ssh|(git@[\\w\\.-]+))(:(//)?)([\\w\\.@\\:/\\-~]+)(\\.git)?(/)?$"  # noqa: E501
```
mkanoor, line 332: @bzwei should this be prefixed with r"...."

src/aap_eda/services/project/scm.py
```diff
@@ -322,3 +327,32 @@ def __call__(
...
+        args = [GIT_COMMAND, "check-ref-format", refspec]
+    with tempfile.TemporaryDirectory(prefix="EDA_REFSPEC") as home_dir:
+        result = subprocess.run(args, cwd=home_dir)
+        if result.returncode != 0:
+            logger.error(
+                f"git check-ref-format {refspec} failed with exit code "
+                f"{result.returncode}"
+            )
+            return False
+        if any(arg in refspec for arg in ["{{", "}}", "lookup"]):
```
mkanoor, line 355: @bzwei Is this for branch test or ref spec. I think the git check-ref-format already says these are invalid with and without spaces.

mkanoor:
@bzwei do we need to check the integration [tests](https://github.com/ansible/eda-server/blob/e6e49f146ace9fdc9d220de041017736613dd720/tests/integration/tasks/test_projects.py#L44) to see if they error comes in the correct field names so UI can display it.

src/aap_eda/services/project/scm.py
Alex-Izquierdo: It is always better to compile the expression for performance reasons. 
Docs: https://docs.python.org/3/library/re.html#re.compile 
Example: https://github.com/ansible/eda-server/blob/e6e49f146ace9fdc9d220de041017736613dd720/src/aap_eda/core/utils/rulebook_process_logs.py#L21

bzwei:
> @bzwei do we need to check the integration [tests](https://github.com/ansible/eda-server/blob/e6e49f146ace9fdc9d220de041017736613dd720/tests/integration/tasks/test_projects.py#L44) to see if the error comes in the correct field names so UI can display it.

Added

bzwei:
The e2e failure is fixed by https://github.com/ansible/eda-qa/pull/474.

Alex-Izquierdo (approved):
LGTM
Regarding the [comment](https://github.com/ansible/eda-server/pull/1335#discussion_r2124979007) from @mkanoor raw strings `r'...'` affects to how the backslashes are interpreted. Here they are escaped so it would not be necessary. 

I personally prefer raw strings because it is easier to read and debug and less error prone. I suggest to change it but I won't block the PR for this. Up to you bill.

Alex-Izquierdo:
sonar is complaining about a non covered line `which git`. I'm fine to ignore it, but it is also true that the test for it would be also simple.

ptoscano:
There are two sonarqube reports on the "http" urls used in tests, because `Using http protocol is insecure. Use https instead`. It seems there is no way to disable a specific rule with an inline comment (like with flake8/pylint/etc):
https://community.sonarsource.com/t/python-add-inline-disabling-of-specific-rule/64184
(https://portal.productboard.com/sonarsource/3-sonarqube-server/c/755-python-users-can-suppress-specific-issues-with-nosonar-with-a-rule-key seems to be a way to request it more)
Another option, as mentioned in the above forum, is to disable sonarqube altogether for those two lines, eg:
```python
        ("http://git.example.com/repo.git/sub/r2.git", True),  # /NOSONAR
```
IMHO either this, or let's ignore the issue altogether (which will show up in case those lines are changed again in the future).

Alex-Izquierdo (approved):

PR #1343, opened 2025-06-09
fix: write extra_vars into a file when call ansible-runner

bzwei:
Treat all str values in extra_vars as unsafe. Write them to a yaml file to be consumed by the ansible-runner.

### DIFF
--- a/src/aap_eda/core/utils/safe_yaml.py
+++ b/src/aap_eda/core/utils/safe_yaml.py
@@ -1,0 +1,57 @@
+#  Copyright 2025 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import yaml
+
+
+class UnsafeString(str):
+    pass
+
+
+def unsafe_string_representer(
+    dumper: yaml.SafeDumper, obj: UnsafeString
+) -> yaml.ScalarNode:
+    return dumper.represent_scalar("!unsafe", obj)
+
+
+def get_dumper() -> yaml.SafeDumper:
+    """Add representers to a YAML seriailizer."""
+    safe_dumper = yaml.SafeDumper
+    safe_dumper.add_representer(UnsafeString, unsafe_string_representer)
+    return safe_dumper
+
+
+def dump(filename: str, data: dict) -> None:
+    """Write data to a file as YAML with unsafe strings."""
+    if not isinstance(data, dict):
+        raise TypeError("Data must be a dictionary")
+
+    def transform_strings(data):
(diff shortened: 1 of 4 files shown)


src/aap_eda/core/utils/safe_yaml.py
```diff
@@ -0,0 +1,58 @@
...
+    """Add representers to a YAML seriailizer."""
+    safe_dumper = yaml.SafeDumper
+    safe_dumper.add_representer(UnsafeString, unsafe_string_representer)
+    return safe_dumper
+
+
+def dump(filename: str, data: dict) -> None:
+    """Write data to a file as YAML with unsafe strings."""
+    if not isinstance(data, dict):
+        raise TypeError("Data must be a dictionary")
```
mkanoor, line 39: @bzwei can you add a test fot this line

src/aap_eda/core/utils/safe_yaml.py
```diff
@@ -0,0 +1,58 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import yaml
+
+
+class UnsafeString(str):
+    def __new__(cls, value):
```
Alex-Izquierdo, line 19: What's the purpose of overriding the constructor if we are not doing anything?

src/aap_eda/core/utils/safe_yaml.py
```diff
@@ -0,0 +1,58 @@
...
+    safe_dumper.add_representer(UnsafeString, unsafe_string_representer)
+    return safe_dumper
+
+
+def dump(filename: str, data: dict) -> None:
+    """Write data to a file as YAML with unsafe strings."""
+    if not isinstance(data, dict):
+        raise TypeError("Data must be a dictionary")
+
+    def transform_strings(data):
```
Alex-Izquierdo, line 41: Do we need an else statement here. Dict could contain unsupported nested types.

src/aap_eda/services/project/scm.py
```diff
@@ -355,7 +360,4 @@ def is_refspec_valid(refspec: str, is_branch: bool) -> bool:
                 f"{result.returncode}"
             )
             return False
-        if is_branch and all(arg in refspec for arg in ["{{", "}}", "lookup"]):
```
Alex-Izquierdo, line 358: Is this removal related to this PR? it does not looks like.

src/aap_eda/core/utils/safe_yaml.py
mkanoor: Tested with this and it works
```
class UnsafeString(str):
    pass
```

src/aap_eda/services/project/scm.py
mkanoor: Git allows a branch that looks like this
"{{lookup('lines','id')}}" 
with no spaces. If there are spaces the git check-ref-format rejects it.

While we were waiting on the [unsafe](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_advanced_syntax.html) approach we need a temporary fix and now that we have unsafe strings we should allow users to create the branches with those names.

src/aap_eda/core/utils/safe_yaml.py
bzwei: the last line `return data` is the else statement. Remaining types will be serialized by `yaml.dump()` itself. The main focus of this PR is for unsafe string type.

PR #1350, opened 2025-06-23
refactor: enhance db existence validation

Alex-Izquierdo:
Some refactoring in validators. Some of them are not used anymore or they do check for the existence of the DB by fetching the record, which is inefficient. DRF validators are expected to raise `ValidationError` or return None.  In case of the `organization_id` has been created a mixin to avoid code duplication. 

Refactored:
- organization
- rulebook
- credential_type
- awx_token

Removed:
- event_streams
- extra_vars

Technically, in most of the cases a PrimaryKey field would be better, but that would require a more extended refactor to keep 100% backward compatibility, since we would have to implement custom error messages among other changes. There are multiple fields candidate to be a PrimaryKeyField and such a conversion is not the scope of this PR. 
Here I only address the fields where we are duplicating DB queries, duplicating field definition and removing unused validators.

### DIFF
--- a/src/aap_eda/tasks/orchestrator.py
+++ b/src/aap_eda/tasks/orchestrator.py
@@ -178,8 +178,32 @@ def _run_request(
 def queue_dispatch(
     process_parent_type: ProcessParentType,
     process_parent_id: int,
+    request_type: Optional[ActivationRequest] = None,
+    request_id: str = "",
+) -> None:  # pragma: no cover
+    job_id = _manage_process_job_id(process_parent_type, process_parent_id)
+    with advisory_lock(job_id, wait=False) as acquired:
+        if not acquired:
+            LOGGER.debug(
+                f"queue_dispatch({job_id}) already being ran, "
+                f"not dispatching request {request_type}",
+            )
+            return
+        queue_dispatch_no_lock(
+            process_parent_type,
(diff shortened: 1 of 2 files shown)


hsong-rh (dismissed):

src/aap_eda/api/serializers/activation.py
```diff
@@ -620,12 +616,6 @@ class Meta:
             "skip_audit_events",
         ]
 
-    organization_id = serializers.IntegerField(
-        required=False,
-        allow_null=False,
-        validators=[validators.check_if_organization_exists],
-        error_messages={"null": "Organization is needed"},
-    )
```
ptoscano, line 628: The newly added `OrganizationIdFieldMixin` mixin has `required=True` whereas this field (replaced by the mixin) has `required=False`; is the change correct here? If so, I'd at least mention that in the PR/commit message

bzwei (approved):

mkanoor (approved):

src/aap_eda/api/serializers/activation.py
Alex-Izquierdo: Hi @ptoscano the requirement is applied only for the creation, it does not require any field in PATCH unless it is explicitily set (because of `partial=True`)

PR #1353, opened 2025-07-01
Fix test failures due to uptream DAB WIP

AlanCoding:
For https://github.com/ansible/django-ansible-base/pull/749

### DIFF
--- a/src/aap_eda/core/management/commands/create_initial_data.py
+++ b/src/aap_eda/core/management/commands/create_initial_data.py
@@ -17,12 +17,12 @@ import os
 
 from ansible_base.rbac import permission_registry
 from ansible_base.rbac.models import DABPermission, RoleDefinition
+from django.apps import apps
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.core.management import BaseCommand
 from django.db import transaction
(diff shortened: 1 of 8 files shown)


src/aap_eda/core/management/commands/create_initial_data.py
```diff
@@ -1291,6 +1291,14 @@ def handle(self, *args, **options):
         self._remove_deprecated_credential_kinds()
         enable_redis_prefix()
 
+    def content_type_model(self):
+        try:
+            # DAB RBAC migrated to a custom type model, try to use that here
+            return apps.get_model("dab_rbac", "DABContentType")
+        except LookupError:
```
Alex-Izquierdo, line 1298: Hi @AlanCoding If we finally depend on a new revision of DAB that includes this change, this fallback should not be necessary. Am I right?

AlanCoding: Correct.

AlanCoding:
I kept this separate as documentation that checks were passing, setting the expectation that it fixes the downstream check failures in the DAB PR.

Screenshot From 2025-07-02 14-41-37

the other PR is merged, this remains good to have documented, as closed PR.

PR #1355, opened 2025-07-02
Fix test failures due to uptream DAB ContentType changes

AlanCoding:
This is another version of https://github.com/ansible/eda-server/pull/1353, but with the dependency changes reverted.

That linked PR (draft) shows that this is effective. I expect tests will pass here (which current DAB) as well. But that PR shows that, if we merge this, it will make downstream tests for https://github.com/ansible/django-ansible-base/pull/749 will pass.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -660,14 +660,14 @@ files = [
 
 [[package]]
 name = "django"
-version = "4.2.16"
+version = "4.2.23"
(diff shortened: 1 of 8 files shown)


Alex-Izquierdo (approved):

tvo318 (approved):
Looks alright to me!

PR #1358, opened 2025-07-06
feat: add support for mTLS EventStreams

mkanoor:
This depends on
 - Gateway PR which adds mTLS Routes
 - Platform collection which can be used by the installer
 - Installer changes to create a mTLS route

Adding a new event stream for mTLS support. A new credential has been added which can be used in conjuction with a new mTLS EventStream. When the mTLS EventStream is created if the user has added a Certificate we add it to the Gateway which can then be pulled by Envoy.

Customers in the financial sector have a requirement to use Certificate based auth when sending events to EventStreams.

This change uses the permissive mTLS feature in Envoy to dynamically update the certificates without having to restart the server. As long as Gateway has the latest Certifcate Envoy can pull it in via Secret Discovery Service (SDS) so the data flow is EDA --> Gateway <-- Envoy

This is not a breaking change but its an optional feature that requires the Gateway changes 

Create a Certificate using OpenSSL
```
openssl req -new -newkey rsa:4096 -days 365 -nodes -x509 \
    -subj "/C=US/ST=NJ/L=Mahwah/O=Demo/OU=R&D Department/CN=agent1.example.com" \
    -keyout agent1.key  -out agent1.crt
```
Add a new mTLS Credential (my_mtls) in EDA and load the above certificate
Add a new mTLS EventStream using the mTLS Credential (my_mtls)
Send an Event Payload to the above EventStream and it should be successful.

### DIFF
--- a/src/aap_eda/services/sync_certs.py
+++ b/src/aap_eda/services/sync_certs.py
@@ -1,0 +1,197 @@
+#  Copyright 2025 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+"""Synchronize Certificates with Gateway."""
+import hashlib
+import logging
+from typing import Optional
+from urllib.parse import urljoin
+
+import requests
+import yaml
+from ansible_base.resource_registry import resource_server
+from django.conf import settings
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+from rest_framework import status
+
+from aap_eda.core import enums, models
+from aap_eda.core.exceptions import GatewayAPIError, MissingCredentials
+
+LOGGER = logging.getLogger(__name__)
+SLUG = "api/gateway/v1/ca_certificates/"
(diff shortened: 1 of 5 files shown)


src/aap_eda/services/sync_certs.py
```diff
@@ -0,0 +1,197 @@
...
+        headers = self._prep_headers()
+        response = requests.delete(
+            url,
+            headers=headers,
+            verify=self.gateway_ssl_verify,
+            timeout=DEFAULT_TIMEOUT,
+        )
+        if response.status_code == status.HTTP_200_OK:
+            LOGGER.debug("Certificate object deleted")
+        if response.status_code == status.HTTP_404_NOT_FOUND:
```
Alex-Izquierdo, line 136: This should be `elif` otherwise we would be raising GatewayAPIError incorrectly, 
```suggestion
        elif response.status_code == status.HTTP_404_NOT_FOUND:
```

src/aap_eda/services/sync_certs.py
```diff
@@ -0,0 +1,197 @@
...
+            LOGGER.warning("Certificate object missing during delete")
+        else:
+            LOGGER.error("Couldn't delete certificate object in gateway")
```
Alex-Izquierdo, line 139: Here should we not print the response, status code and reason of the error?

src/aap_eda/services/sync_certs.py
```diff
@@ -0,0 +1,197 @@
...
+
+            if response.status_code in [
+                status.HTTP_200_OK,
+                status.HTTP_201_CREATED,
+            ]:
+                LOGGER.debug("Certificate updated")
+            elif response.status_code == status.HTTP_400_BAD_REQUEST:
+                LOGGER.error("Update failed")
+            else:
+                LOGGER.error("Couldn't update certificate")
```
Alex-Izquierdo, line 105: should we not print the response and reason of the error?

PR #1359, opened 2025-07-07
feat: Not allow local resource management by default

bzwei:
AAP-48659 asks to remove ALLOW_LOCAL_RESOURCE_MANAGEMENT in EDA. This flag is no longer used across AAP. 

https://issues.redhat.com/browse/AAP-48659

This PR converts ALLOW_LOCAL_RESOURCE_MANAGEMENT to an internal flag. The default value is False meaning EDA is not allowed to create/modify shared resources. PDE does not need to manipulate the flag. The flag is not completely removed for development purpose. It can be turn on to run EDA standalone.


Alex-Izquierdo:
@bzwei Changing the default value is a breaking change and I think it is not what is required if you look at wgat AWX did, which is precisely the opposite, they allow to modify it: https://github.com/ansible/awx/pull/16033

If we merge this change the default standalone deployment done by the Operator would not work which would create a big distortion in the upstream community.

bzwei:
> @bzwei Changing the default value is a breaking change and I think it is not what is required if you look at wgat AWX did, which is precisely the opposite, they allow to modify it: [ansible/awx#16033](https://github.com/ansible/awx/pull/16033)
> 
> If we merge this change the default standalone deployment done by the Operator would not work which would create a big distortion in the upstream community.

AAP-48659 and its parent require only AWX can create local resources. It requires the new change to be committed to 2.6 branch only, which is also our upstream main. This PR preserves the capability to create resources in EDA. Nothing needs to be changed if the operator already configures ALLOW_LOCAL_RESOURCE_MANAGEMENT.

Alex-Izquierdo:
> > @bzwei Changing the default value is a breaking change and I think it is not what is required if you look at wgat AWX did, which is precisely the opposite, they allow to modify it: [ansible/awx#16033](https://github.com/ansible/awx/pull/16033)
> > If we merge this change the default standalone deployment done by the Operator would not work which would create a big distortion in the upstream community.
> 
> AAP-48659 and its parent require only AWX can create local resources. It requires the new change to be committed to 2.6 branch only, which is also our upstream main. This PR preserves the capability to create resources in EDA. Nothing needs to be changed if the operator already configures ALLOW_LOCAL_RESOURCE_MANAGEMENT.

If you have to change the operator to enforce the value of ALLOW_LOCAL_RESOURCE_MANAGEMENT, that means that it should need to be reverted for the downstream builds who should set it to `false` again and you would ending up to not meet the original goal which is to remove the ALLOW_LOCAL_RESOURCE_MANAGEMENT. Even in that case, this PR should not be merged before update the operator, otherwise it would break it. So, at least, the documentation and CRD's examples of the upstream operator should be updated prior this change, assuming that the operator would not work OoB with the default values.

bzwei:
> If you have to change the operator to enforce the value of ALLOW_LOCAL_RESOURCE_MANAGEMENT, that means that it should need to be reverted for the downstream builds who should set it to `false` again and you would ending up to not meet the original goal which is to remove the ALLOW_LOCAL_RESOURCE_MANAGEMENT. Even in that case, this PR should not be merged before update the operator, otherwise it would break it. So, at least, the documentation and CRD's examples of the upstream operator should be updated prior this change, assuming that the operator would not work OoB with the default values.

We either need to create a 2.6 branch for this change, or document how to enable local resource management in the operator repo.

Alex-Izquierdo (approved):

hsong-rh (approved):

ttuffin (approved):

PR #1366, opened 2025-07-23
Update DAB to July 22

AlanCoding:
Normal periodic bump of DAB.

It is suspected that this version may be failing E2E tests, so we request that label is added so we get pre-merge results here.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -924,7 +924,7 @@ bcrypt = ["bcrypt"]
 
 [[package]]
 name = "django-ansible-base"
-version = "2025.5.8"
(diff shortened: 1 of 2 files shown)


poetry.lock
```diff
@@ -924,7 +924,7 @@ bcrypt = ["bcrypt"]
 
 [[package]]
 name = "django-ansible-base"
-version = "2025.5.8"
+version = "25.7.21.0.dev768+g58a1b55"
```
AlanCoding, line 927: I don't know why this changed, and wonder if it was wrong before. It could have been due to packaging bugs on the DAB side, I don't know.

PR #1368, opened 2025-07-29
feat: Enable static event stream URLs with user-configurable UUIDs

hsong-rh:
Allow users to provide custom UUIDs when creating event streams for consistent, static URLs. Users can also change UUIDs during updates for maximum flexibility.

https://issues.redhat.com/browse/AAP-50564


src/aap_eda/api/serializers/event_stream.py
```diff
@@ -41,6 +42,22 @@ class EventStreamInSerializer(serializers.ModelSerializer):
         ],
         error_messages={"null": "EdaCredential is needed"},
     )
+    uuid = serializers.UUIDField(required=False, allow_null=True)
+
+    def validate_uuid(self, value):
```
mkanoor, line 47: @hsong-rh Wont Django's unique=True and we have a default set is this code still needed

hsong-rh: @mkanoor Use this validation function; we can handle exceptions better. Otherwise, we need to change in views to catch errors if they happen.

PR #1370, opened 2025-07-31
feat: Add configurable image pull policy for Decision Environments

hsong-rh:
This PR implements configurable image pull policies for Decision Environments, allowing users to control when container images are pulled before activation execution.

https://issues.redhat.com/browse/AAP-24056


src/aap_eda/services/activation/engine/podman.py
```diff
@@ -143,9 +143,15 @@ def start(self, request: ContainerRequest, log_handler: LogHandler) -> str:
 
         try:
             LOGGER.info(f"Image URL is {request.image_url}")
-            if request.pull_policy == "Always" or not self._image_exists(
-                request.image_url,
+            image_existed = self._image_exists(request.image_url)
```
Alex-Izquierdo, line 146: it's still missing the logic for `never`, am I right?

Alex-Izquierdo: I disagree, that's incorrect. If we support `never` policy it must behave as expected. With never policy we should not try to pull the image never, it is supposed to be there. 
https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy

Alex-Izquierdo: @hsong-rh Podman actually accepts the pull policy [when running a container](https://docs.podman.io/en/v5.0.1/markdown/podman-run.1.html#pull-policy) in CLI or as [image operation](https://docs.podman.io/en/latest/_static/api.html#tag/images/operation/ImagePullLibpod) in the rest api (which is what the python client consumes)

In our current implementation we take care explicitly of the live cycle of the image so we need to implement the logic for all three options (only `never` would be missing)

src/aap_eda/core/enums.py
```diff
@@ -138,6 +138,14 @@ class ActivationRequest(DjangoStrEnum):
     AUTO_START = "auto_start"
 
 
+class ImagePullPolicy(DjangoStrEnum):
+    """Image pull policy for decision environments."""
+
+    ALWAYS = "Always"
```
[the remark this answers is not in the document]
Alex-Izquierdo: We need to create mappers from EDA to k8s and podman since the values are differents.

B-Whitt (approved):

B-Whitt (dismissed):

src/aap_eda/api/serializers/decision_environment.py
```diff
@@ -70,6 +82,26 @@ class DecisionEnvironmentCreateSerializer(serializers.ModelSerializer):
...
+
+    def validate_pull_policy(self, value):
+        """Normalize pull policy input and validate."""
+        if value:
+            normalized = convert_pull_policy_from_frontend(value)
+            if normalized not in ImagePullPolicy.values():
+                valid_options = ["always", "never", "missing"] + list(
+                    ImagePullPolicy.values()
+                )
+                raise serializers.ValidationError(
```
bzwei, line 99: Why show to the user 6 total options? We should show only 3 options. The backend can convert them in a case insensitive way.

src/aap_eda/core/utils/decision_environment.py
```diff
@@ -0,0 +1,39 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+from aap_eda.core.enums import ImagePullPolicy
+
+
+def convert_pull_policy_to_frontend(pull_policy: str) -> str:
+    """Convert internal pull policy enum to frontend-friendly string."""
+    policy_mapping = {
```
bzwei, line 20: Why do we need to define a set of frontend values? The UI has freedom convert the enums to any string. How does the API doc look like? How does other API endpoints handle enum?

src/aap_eda/core/utils/decision_environment.py
```diff
@@ -0,0 +1,39 @@
...
+        ImagePullPolicy.ALWAYS: "always",
+        ImagePullPolicy.NEVER: "never",
+        ImagePullPolicy.IF_NOT_PRESENT: "missing",
+    }
+    return policy_mapping.get(pull_policy, "always")
```
Alex-Izquierdo, line 25: Even if you want to continue with the current approach of these utils, this line is a potential bug that masks the default policy.

src/aap_eda/services/activation/engine/common.py
```diff
@@ -155,6 +155,11 @@ def get_container_request(self) -> ContainerRequest:
         """Return ContainerRequest used for creation."""
         self.validate()
 
+        pull_policy = (
+            self.decision_environment.pull_policy
+            or settings.DEFAULT_PULL_POLICY
```
Alex-Izquierdo, line 160: `settings.DEFAULT_PULL_POLICY` should be now validating the value as well.

src/aap_eda/core/utils/decision_environment.py
hsong-rh: > Right now for podman we are implementing the policy in our side but as you say in comment, once the pypodman client support it, we are going to pass the values for podman, so you would use "convert_pull_policy_to_frontend" for it?
> 
All these changes are temporary. When podman 5.6 is available, they will not be needed. We simply pass the pull_policy to podman-py APIs, like what we do now for kubernetes.

Alex-Izquierdo: > 
> All these changes are temporary. When podman 5.6 is available, they will not be needed. We simply pass the pull_policy to podman-py APIs, like what we do now for kubernetes.

exactly, and then you would need to convert them as for the UI. Since podman values are what controller and ui uses, makes sense to use those values by default, and convert only k8s.

src/aap_eda/api/serializers/decision_environment.py
```diff
@@ -70,6 +79,37 @@ class DecisionEnvironmentCreateSerializer(serializers.ModelSerializer):
...
+    pull_policy = serializers.CharField(
+        required=False,
+        default=settings.DEFAULT_PULL_POLICY,
+        help_text="Pull policy for the container image",
+    )
+
+    def validate_pull_policy(self, value):
+        """Validate and normalize pull policy input."""
+        if not value:
+            return ImagePullPolicy.ALWAYS.value
```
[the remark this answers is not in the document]
Alex-Izquierdo: My bad, you are right. https://www.django-rest-framework.org/api-guide/serializers/#field-level-validation

src/aap_eda/api/serializers/decision_environment.py
```diff
@@ -70,6 +74,28 @@ class DecisionEnvironmentCreateSerializer(serializers.ModelSerializer):
...
+    )
+
+    def validate_pull_policy(self, value):
+        """Validate and normalize pull policy input."""
+        valid_inputs = {
+            ImagePullPolicy.ALWAYS.value,
+            ImagePullPolicy.NEVER.value,
+            ImagePullPolicy.MISSING.value,
+            settings.DEFAULT_PULL_POLICY,
+        }
```
Alex-Izquierdo, line 90: The enum should be the source of truth, we should not hardcode the values here. 
`settings.DEFAULT_PULL_POLICY` must be validated to meet a valid enum, it should not be here. 
The open api spec should be honored.

```suggestion
        valid_inputs = {policy.value for policy in ImagePullPolicy}
```

src/aap_eda/core/models/decision_environment.py
```diff
@@ -40,5 +48,12 @@ class Meta:
         default=None,
         on_delete=models.SET_NULL,
     )
+    pull_policy = models.CharField(
+        max_length=16,
+        choices=PULL_CHOICES,
```
Alex-Izquierdo, line 53: choices must come from the enum.

src/aap_eda/services/activation/engine/common.py
```diff
@@ -116,7 +116,7 @@ class ContainerRequest(BaseModel):
     process_parent_id: int
     credential: tp.Optional[Credential] = None
     ports: tp.Optional[list[tuple]] = None
-    pull_policy: str = settings.DEFAULT_PULL_POLICY  # Always by default
+    pull_policy: str = ImagePullPolicy.ALWAYS
```
Alex-Izquierdo, line 119: We should not define a default value here, it comes from the model.

src/aap_eda/api/serializers/decision_environment.py
```diff
@@ -70,6 +79,37 @@ class DecisionEnvironmentCreateSerializer(serializers.ModelSerializer):
             validators.check_credential_registry_username_password,
         ],
     )
+    pull_policy = serializers.CharField(
+        required=False,
+        default=settings.DEFAULT_PULL_POLICY,
```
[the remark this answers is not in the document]
Alex-Izquierdo: this has not been resolved.

src/aap_eda/api/serializers/decision_environment.py
hsong-rh: removed

src/aap_eda/settings/post_load.py
```diff
@@ -507,6 +507,16 @@ def post_loading(loaded_settings: Dynaconf):
         settings
     )
 
+    settings.DEFAULT_PULL_POLICY = settings.DEFAULT_PULL_POLICY.lower()
```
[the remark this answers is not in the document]
hsong-rh: Let's remove it completely to reduce the complexity

PR #1381, opened 2025-08-14
fix: moving project out of stuck pending or running state

bzwei:
Fix bug in existing monitor_project_tasks for pending state. 
Add new logic to handle running state. 
Add new django command to manually change the import state of a project

https://issues.redhat.com/browse/AAP-51139

Fix the bug in requeueing pending projects that are no longer in the queue.
Detect running projects that are no longer in the queue or already failed but not updated in database.
The django command can be used as a last resort helping the project to move out of a stuck state.

Once a project is stuck in pending or running, it cannot be restarted. This work helps to move such projects out of the situation.

### DIFF
--- a/src/aap_eda/core/management/commands/fix_project_state.py
+++ b/src/aap_eda/core/management/commands/fix_project_state.py
@@ -1,0 +1,80 @@
+#  Copyright 2025 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


Alex-Izquierdo (commented):
What is the use case of the command that can not be covered by just re-syncing the project?

Alex-Izquierdo (approved):

mkanoor (approved):

PR #1384, opened 2025-08-21
fix: when doing credential test inputs and metadata are mandatory

mkanoor:
When doing external sms credential test we need both inputs and metadata.

There is no UI field related to this unlike other ValidationErrors We return a 400 when either of them is missing


kaiokmo (approved):

Alex-Izquierdo (approved):

hsong-rh (approved):
This ATF test also needs to be updated: https://gitlab.cee.redhat.com/ansible/testing/eda-test-suite/-/blob/main/tests/api/test_credential_types.py?ref_type=heads#L204

PR #1387, opened 2025-08-25
fix: add forward_events field - AAP-49285

Alex-Izquierdo:
## Problem
/event-stream REST API resource currently has the field "test_mode" that sets if the event stream sends the events to the attached activations. 
This name mismatches the name used by the UI and the collection, which is "forward_events"
This causes confusion to customer who expects name parity in the interfaces. This confusion is even greater given that the "test_mode" boolean represents a NOT operator of the "forward_events" option in the clients. 

## Includes
- Add the new field `forward_events` for GET as reverse of `test_mode`
- Add the new field `forward_events` for POST and PATCH setting it `test_mode` as reverse value. Are mutually exclusive. 
- Update/Add description text for the fields
- Add API tests. 

Jira: https://issues.redhat.com/browse/AAP-49285

### DIFF
--- a/src/aap_eda/api/serializers/event_stream.py
+++ b/src/aap_eda/api/serializers/event_stream.py
@@ -43,6 +43,23 @@ class EventStreamInSerializer(serializers.ModelSerializer):
         error_messages={"null": "EdaCredential is needed"},
     )
     uuid = serializers.UUIDField(required=False, allow_null=True)
+    test_mode = serializers.BooleanField(
+        required=False,
+        help_text=(
+            "Enable test mode for the event stream. When true, events are "
+            "captured for testing but not forwarded to activations. "
+            "Mutually exclusive with forward_events."
+        ),
+    )
+    forward_events = serializers.BooleanField(
+        required=False,
+        help_text=(
+            "Whether to forward events to activations (inverse of "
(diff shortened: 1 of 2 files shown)


tests/integration/api/test_event_stream.py
```diff
@@ -491,6 +518,146 @@ def test_update_event_stream_uuid_change_allowed(
     assert response.data["uuid"] == str(new_uuid)
 
 
+@pytest.mark.django_db
+def test_create_event_stream_with_forward_events_true(
```
bzwei, line 522: the testing logic is very similar in multiple tests. Can you use pytest.mark.parametrize to consolidate them?

src/aap_eda/api/serializers/event_stream.py
```diff
@@ -43,6 +43,23 @@ class EventStreamInSerializer(serializers.ModelSerializer):
         error_messages={"null": "EdaCredential is needed"},
     )
     uuid = serializers.UUIDField(required=False, allow_null=True)
+    test_mode = serializers.BooleanField(
+        required=False,
+        help_text=(
```
bzwei, line 48: Is there a plan to obsolete test_mode in the future? If so we may reflect it in the help_text.
Can we accept both test_mode and forward_events as long as they are not contradict each other?

PR #1396, opened 2025-09-18
fix: handling of HTTP headers in EventStreams

mkanoor:
To include all headers you have to explicitly set the

additional_data_headers = *

* If the header key is used for Authentication it will be redacted

* To pick and choose which headers get included, list them as a comma delimited string in additional_data_headers e.g Host,Origin,X-Request-ID

* Any header that starts with X-Envoy will be dropped
* X-Trusted-Proxy,X-Forwarded-For,X-Real-IP will be dropped

### DIFF
--- a/src/aap_eda/api/views/external_event_stream.py
+++ b/src/aap_eda/api/views/external_event_stream.py
@@ -16,6 +16,7 @@
 import datetime
 import logging
 import urllib.parse
+from typing import Any
 
 import yaml
 from django.conf import settings
@@ -46,6 +47,8 @@ from aap_eda.core.utils.credentials import get_resolved_secrets
 from aap_eda.services.pg_notify import PGNotify
 
 logger = logging.getLogger(__name__)
+UNSAFE_HEADER_KEYS = {"X-Trusted-Proxy", "X-Forwarded-For", "X-Real-IP"}
+REDACTED_STRING = "********"
 
 
 class ExternalEventStreamViewSet(viewsets.GenericViewSet):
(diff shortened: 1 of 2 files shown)


Alex-Izquierdo:
this change would need also update our product documentation, right?

Alex-Izquierdo (dismissed):

src/aap_eda/api/views/external_event_stream.py
```diff
@@ -235,13 +251,17 @@ def post(self, request, *_args, **kwargs):
             logger.warning("Error fetching external secrets %s", str(err))
             return Response(status=status.HTTP_500_INTERNAL_SERVER_ERROR)
 
+        event_headers = self._redacted_headers(
```
bzwei, line 254: in line 245 `logger.debug("Headers %s", request.headers)`, should we just remove or redact sensitive headers?

src/aap_eda/api/views/external_event_stream.py
```diff
@@ -107,20 +109,34 @@ def _parse_body(self, content_type: str, body: bytes) -> dict:
...
-                if value:
-                    event_headers[key] = value
-        else:
-            event_headers = dict(headers)
-            if header_key in event_headers:
-                event_headers.pop(header_key)
+            if self.event_stream.additional_data_headers == "*":
+                event_headers = dict(headers)
+                if header_key in event_headers:
+                    event_headers[header_key] = "Redacted"
```
bzwei, line 120: should we write literally "Redacted", or use typical way as "*****"?

bzwei (approved):

PR #1523, opened 2026-04-10
Prevents port numbers from changing datatype.

AlexSCorey:
PDE reporting an issue where the new event persistence feature was breaking the CI job.  turns out Dynaconf was loading the `EVENT_PERSISTENCE_DB_PORT` env var as an integer but it was typed as an `Optional[str]`.  This pr fixes that problem, and adds tests to ensure it doesn't return.

## Summary by CodeRabbit

* **Bug Fixes**
  * Settings validation now accepts integer inputs for values expected as strings by converting them to strings, trims surrounding whitespace from string inputs, and preserves explicit null values. No public configuration interfaces were changed.

* **Tests**
  * Added unit tests verifying integer-to-string coercion, whitespace trimming, and null-preservation during settings post-load.

### DIFF
--- a/src/aap_eda/settings/post_load.py
+++ b/src/aap_eda/settings/post_load.py
@@ -302,10 +302,10 @@ def _get_default_pg_notify_dsn_server(settings: Dynaconf) -> str:
         f"dbname={settings.DATABASES['default']['NAME']} "
         f"user={settings.DATABASES['default']['USER']} "
         f"password={settings.DATABASES['default']['PASSWORD']} "
-        f"sslmode={db_options.get('sslmode','allow')} "
-        f"sslcert={db_options.get('sslcert','')} "
-        f"sslkey={db_options.get('sslkey','')} "
-        f"sslrootcert={db_options.get('sslrootcert','')} "
+        f"sslmode={db_options.get('sslmode', 'allow')} "
+        f"sslcert={db_options.get('sslcert', '')} "
+        f"sslkey={db_options.get('sslkey', '')} "
+        f"sslrootcert={db_options.get('sslrootcert', '')} "
     )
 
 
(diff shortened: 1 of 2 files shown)


src/aap_eda/settings/post_load.py
```diff
@@ -330,10 +330,15 @@ def _enforce_types(settings: Dynaconf) -> None:
             settings[key] = _get_int(settings, key)
         elif key_type is defaults.UrlSlash:
             settings[key] = _get_url_end_slash(settings, key)
+        elif key_type is str or key_type is Optional[str]:
+            # Convert int to str for Optional[str] types (e.g., port numbers)
+            value = settings[key]
+            if value is not None and isinstance(value, int):
+                settings[key] = str(value)
+            else:
+                settings[key] = _get_stripped_str(settings, key)
```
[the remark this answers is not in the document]
ttuffin: I don't think this is a valid concern. We already match boolean types earlier in the loop on L327 and it's unrealistic a user would try to pass a boolean value instead of an int. The proposed fix adds unnecessary complexity.

ttuffin:
@AlexSCorey I think this approach is fine. Just need to address the failing tests.

jneedle (approved):

src/aap_eda/settings/post_load.py
```diff
@@ -330,10 +351,12 @@ def _enforce_types(settings: Dynaconf) -> None:
             settings[key] = _get_int(settings, key)
         elif key_type is defaults.UrlSlash:
             settings[key] = _get_url_end_slash(settings, key)
+        elif key_type is Optional[str]:
+            _enforce_optional_str_type(settings, key, settings[key])
+        elif key_type is str:
+            _enforce_str_type(settings, key, settings[key])
         elif not isinstance(settings[key], key_type):
```
wfealdel, line 358: @AlexSCorey  
Would be possible to update this line to use the code from line 339, by doing this I think we could get ride of the `_enforce_str_type()`

AlexSCorey: @wfealdel I'm not sure I understand.  I broke this function apart into several functions because sonar cloud was complaining about it being logically complex.

wfealdel: @AlexSCorey I was asking if it's possible to have something like this:
 ```python
elif key_type is str:
    settings[key] = _get_stripped_str(settings, key)
elif not isinstance(settings[key], key_type):
    _validate_type(key, settings[key], key_type)
```
 In such case we could remove the `_enforce_str_type` functions as most of the types are validated by the `_enforce_types` one

ttuffin (approved):

src/aap_eda/settings/post_load.py
AlexSCorey: @wfealdel no this won't work because other env var fields like `EVENT_PERSISTENCE_DB_PORT` are `Optional[str]` and  your recommendation entirely removes the logic needed to convert optional integers to strings.  Further, `str`  like `RULEBOOK_QUEUE_NAME` should be rejected with `ImproperlyConfigured`.  Using `_get_stripped_str()` would cause a `AttributeError` crash.

PR #1528, opened 2026-04-14
Improves RBAC role naming conventions and increasing consistency.

AlexSCorey:
This addresses https://redhat.atlassian.net/browse/AAP-59318 partially.  There are some reamining bits regarding rbac roles within awx, and possibly within DAB that would need to be addressed.  For this PR the work was to improve the consistency of the RBAC role names.  

Tests were also updated/added.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized EDA Credential role names from "Eda" to "EDA" for consistent display and access.
  * Added an automated rename to update existing role entries safely, avoiding duplicate conflicts.

* **Tests**
  * Updated integration tests to assert the new "EDA Credential" role name casing.


ttuffin (approved):

AlanCoding:
Just to link this:

https://github.com/ansible-automation-platform/aap-gateway/blob/devel/aap_gateway_api/management/commands/migrate_service_data.py#L158-L164

The concern is that `RoleDefinition` is a synchronized model. On first-install, entries are created by the data loading script here, and then after that `aap-gateway-manage migrate_service_data` runs and puts a copy of the entries into Gateway.

Customers will upgrade from prior 2.6.z versions to this. It's unclear if that will result in the desired behavior or not. It depends on the behavior of the sync from migrate_service_data, and I do not know this right now.

AlexSCorey:
Following advice from @AlanCoding I spun up a devel based aap-dev instance and found role names like `Organization Eda Credential Admin` which is consistent with what's reported in the bug.  Then, I spun up aap-dev against this branch.  It built just fine, with no problem and saw that `Organization EDA Credential Admin` was there.  Notice that in the original role `Eda` vs. in the role created by this branch where we have `EDA` all caps in the name.  It seems that this might be safe to merge, and we do get some consistency, where EDA is all caps'd in stead of Eda.

AlexSCorey:
Linking my conversation with @AlanCoding where he offers suggestions about how to test this pr. https://redhat-internal.slack.com/archives/C06CM09FQSV/p1776094482726909

ptoscano (approved):

PR #1537, opened 2026-04-27
fix: prevent secret redaction oracle attack in project sync errors (AAP-72813)

hsong-rh:
## Summary

Fixes [[AAP-72813](https://redhat.atlassian.net/browse/AAP-72813)] — project sync error redaction creates an oracle attack vector.

The previous code did `msg.replace(secret, "****")` on the entire error message. An attacker controlling a git server could craft responses containing candidate strings. When one matches the password, it gets replaced with `****`, revealing the secret.

Example: password `abc123`, branch `first-abc123-second` → error shows `first-****-second`.

## Changes

Replaced secret-based substring matching with URL-based replacement:
- Replace the full credential-embedded URL with the clean URL
- Handle both URL-encoded and decoded password forms for special characters
- Branch names, server responses, and other error content are left untouched

## Test plan

- [x] `test_git_clone_leak_password` — credentials sanitized from URL in error
- [x] `test_git_clone_no_oracle_attack` — secret in branch name NOT redacted (oracle blocked)
- [x] `test_git_clone_sanitizes_decoded_url` — special-char passwords sanitized in both encoded and decoded forms
- [x] All 31 existing scm tests pass (zero regressions)
- [x] Reproduced locally: branch name containing password preserved as-is, credential URL properly sanitized

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved credential redaction in source-control clone error messages: embedded credentials (including URL-decoded and URL-encoded forms) are fully removed while preserving host/repo details.
* **Tests**
  * Expanded tests to verify redaction behavior, ensure secret-like text outside credential fields remains visible, and confirm sanitization for passwords with special characters.

### DIFF
--- a/src/aap_eda/services/project/scm.py
+++ b/src/aap_eda/services/project/scm.py
@@ -243,9 +243,17 @@ class ScmRepository:
                 git_hash = _executor(extra_vars=extra_vars, env_vars=env_vars)
         except ScmError as e:
             msg = str(e)
-            if secret:
-                msg = msg.replace(secret, "****", 1)
-                msg = msg.replace(quote(secret, safe=""), "****", 1)
+            # Replace credential-embedded URL with clean URL instead
+            # of redacting the password substring (which would create
+            # an oracle attack vector — see AAP-72813).
+            if final_url != url:
+                msg = msg.replace(final_url, url)
+            # Handle URL-decoded form for passwords with special chars
+            # (e.g., "p@ss" is encoded as "p%40ss" in final_url, but
+            # git may print the decoded form in error messages).
+            if secret and secret != quote(secret, safe=""):
+                raw_url = final_url.replace(quote(secret, safe=""), secret)
+                msg = msg.replace(raw_url, url)
             logger.warning("SCM clone failed: %s", msg)
             raise e.__class__(msg) from None
         finally:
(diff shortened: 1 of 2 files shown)


AlexSCorey (dismissed):
This looks good to me.

src/aap_eda/services/project/scm.py
```diff
@@ -243,9 +243,11 @@ def clone(
                 git_hash = _executor(extra_vars=extra_vars, env_vars=env_vars)
         except ScmError as e:
             msg = str(e)
-            if secret:
-                msg = msg.replace(secret, "****", 1)
-                msg = msg.replace(quote(secret, safe=""), "****", 1)
+            if final_url != url:
```
wfealdel, line 249: @hsong-rh 
Pls, consider to add inline comments explaining those URL replacements.

wfealdel (dismissed):

src/aap_eda/services/project/scm.py
hsong-rh: Added inline comments. One explains why we replace the full URL instead of the password substring (oracle attack), and another for the decoded-URL case with special characters.

wfealdel (approved):

jcraiglo1 (approved):
Looks great thank you for the comments!

PR #1539, opened 2026-04-29
fix: api remains reachable when workers are offline

ttuffin:
This change reworks the logic of the /status endpoint so that the EDA API will continue to respond when the workers are offline. Currently, if all activation workers or all default workers are offline EDA will report HTTP 500. This presents a problem for upstream proxies such as Gateway, that interpret this to mean that the EDA API is non-functional and therefore removes the node from processing and returns HTTP 503 with 'no healthy upstream' for any EDA requests.

With the change, if the workers are offline then the API will return HTTP 200 and `STATUS_DEGRADED`, so that the upstream proxy will continue to send requests to the EDA node but is aware that its in a degraded state. The API will respond HTTP 500 if the database is unreachable.

https://redhat.atlassian.net/browse/AAP-73535

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved system health status reporting. The status endpoint now correctly distinguishes between degraded states (when worker services are unavailable but the database remains healthy) and failed states, returning appropriate HTTP status codes.

### DIFF
--- a/src/aap_eda/core/views.py
+++ b/src/aap_eda/core/views.py
@@ -12,7 +12,11 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from ansible_base.lib.constants import STATUS_FAILED, STATUS_GOOD
+from ansible_base.lib.constants import (
+    STATUS_DEGRADED,
+    STATUS_FAILED,
+    STATUS_GOOD,
+)
 from django.db import connection
 from django.db.utils import OperationalError
 from drf_spectacular.utils import OpenApiResponse, extend_schema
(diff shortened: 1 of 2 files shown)


ptoscano (approved):

B-Whitt (approved):

src/aap_eda/core/views.py
```diff
@@ -56,16 +60,20 @@ def _check_database():
         responses={
             status.HTTP_200_OK: OpenApiResponse(
                 StatusResponseSerializer,
+                description="EDA is healthy or degraded (workers down "
```
kaiokmo, line 63: I think this is confusing. If we return 200 and in the openapi response say that "EDA is healthy or degraded", which state should the caller believe in? Below, in lines 85-92 we're checking things correctly. Can it be reflected in the above description somehow?

kaiokmo (changes requested):

src/aap_eda/core/views.py
kaiokmo: please disregard

kaiokmo (approved):

hsong-rh (approved):

kaiokmo (approved):

PR #1543, opened 2026-05-01
fix: update source_mappings hash on project sync to prevent stale validation

### LINKED ISSUE #1542: Rulebook activation restart fails with stale source_mappings hash after project sync to new Git ref
### Bug Summary

When an EDA project's `scm_branch` is updated to a different Git tag (e.g. `v1.0.0` to `v1.1.0`) and the project is synced, existing rulebook activations with event stream source mappings cannot be restarted. The API returns:

```
{'source_mappings': 'Rulebook has changed since the sources were mapped. Please reattach event streams'}
```

The root cause is that `_sync_rulebook()` in `services/project/imports.py` updates `Rulebook.rulesets` and `Activation.rulebook_rulesets` but never updates the `rulebook_hash` embedded inside `Activation.source_mappings`. The stale hash then fails the comparison in `_validate_sources_with_event_streams()`.

### Environment

- eda-server (`main` branch)

### Steps to reproduce

1. Create an EDA project pointing to a Git repo at tag `v1.0.0`
2. Create an event stream
3. Create a rulebook activation using a rulebook from that project, with event stream source mappings attached (the API stores a `rulebook_hash` inside `source_mappings`)
4. Confirm the activation runs successfully
5. Update the EDA project's `scm_branch` from `v1.0.0` to `v1.1.0` (where the rulebook file has any text change, even a comment or whitespace)
6. Sync the project
7. Attempt to restart the activation

### Actual results

Restart fails with HTTP 400:
```json
{"errors": "{'source_mappings': 'Rulebook has changed since the sources were mapped. Please reattach event streams'}"}
```
The activation is set to `error` state. The user must manually edit the activation and re-attach event streams, even when the source definitions are functionally identical.

### Expected results

When a project sync updates a rulebook's content, the `rulebook_hash` stored inside `source_mappings` on affected activations should be updated to match. Restart and enable operations should continue to work without manual re-attachment when the sources are structurally unchanged.

### Additional information

**Root cause** spans two code paths:

1. `src/aap_eda/services/project/imports.py`, `_sync_rulebook()`: updates `rulebook.rulesets` and `Activation.rulebook_rulesets` via bulk `update()`, but does not touch `Activation.source_mappings`.

2. `src/aap_eda/api/serializers/activation.py`, `_validate_sources_with_event_streams()`: recomputes `get_rulebook_hash(rulebook.rulesets)` and compares it against the now-stale `rulebook_hash` in `source_mappings`.

**Proposed fix:** In `_sync_rulebook()`, when the content has changed, iterate over affected activations and rewrite the `rulebook_hash` inside each `source_mappings` entry to match the new hash. A draft patch is available.

amasolov:
## Summary

- Gate `source_mappings` `rulebook_hash` updates during project sync on a **structural source comparison** between the previous `Rulebook.rulesets` (captured before save) and the newly synced content
- Use `build_source_list()` for comparison so unnamed sources get the same positional `__SOURCE_N` names as mapping validation; rules-only changes are ignored
- **Sources unchanged:** update `rulebook_hash` in `source_mappings` so enable/restart validation passes without manual re-attachment (#1542)
- **Sources changed:** leave `rulebook_hash` stale; existing `warnings` field and manual serializer re-mapping flow apply (AAP-65811)
- Do **not** use `Activation.rulebook_rulesets` for comparison (post-`swap_event_stream_sources()` blob)
- Preserve post-swap `rulebook_rulesets` on source-mapped activations; keep auto-restart skip guard
- Mark activations ERROR when `source_mappings` cannot be parsed during a hash-eligible sync
- Project sync already runs inside `transaction.atomic()` in `_sync_project_no_lock()`

## Test plan

- [x] `rulebook_sources_unchanged()` unit tests (rules-only vs source change)
- [x] `test_sync_updates_source_mappings_hash` (rules-only project sync updates hash)
- [x] `test_sync_preserves_source_mappings_hash_when_sources_change` (source config change leaves hash stale)
- [x] `test_sync_marks_error_on_invalid_source_mappings` (bad YAML marks ERROR)
- [x] `_sync_rulebook` integration tests for both paths
- [ ] Full CI suite

## Tracking

- Tracked in Jira: https://redhat.atlassian.net/browse/AAP-74014
- Affected version: AAP 2.6

Closes #1542
Also fixes the bug in https://redhat.atlassian.net/browse/AAP-88826

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved project synchronization for rulebook-based activations.
  - Rule changes now update mapped rulebook references while preserving activation content where appropriate.
  - Source changes are detected more reliably, preventing outdated mappings from being reused.
  - Invalid or unsupported source mappings are marked **Error** with guidance to reattach them.
  - Failed or errored activations can resume automatic restart after successful synchronization, while running activations remain undisturbed.
  - Invalid rulebook structures are now detected and reported more clearly.

### DIFF
--- a/src/aap_eda/core/utils/rulebook.py
+++ b/src/aap_eda/core/utils/rulebook.py
@@ -21,41 +21,80 @@ from aap_eda.core.exceptions import ParseError
 
 LOGGER = logging.getLogger(__name__)
 DEFAULT_SOURCE_NAME_PREFIX = "__SOURCE_"
+_PARSE_ERROR_MSG = "Failed to parse rulebook data"
+
+PG_NOTIFY_DSN = (
+    "host={{postgres_db_host}} port={{postgres_db_port}} "
+    "dbname={{postgres_db_name}} user={{postgres_db_user}} "
+    "password={{postgres_db_password}} sslmode={{postgres_sslmode}} "
+    "sslcert={{eda.filename.postgres_sslcert|default(None)}} "
+    "sslkey={{eda.filename.postgres_sslkey|default(None)}} "
+    "sslpassword={{postgres_sslpassword|default(None)}} "
+    "sslrootcert={{eda.filename.postgres_sslrootcert|default(None)}}"
+)
+
+
+def _parse_rulesets(rulesets_data: str) -> list:
+    """Parse and validate the top-level rulesets YAML structure."""
+    try:
+        rulesets = yaml.safe_load(rulesets_data)
+    except yaml.MarkedYAMLError as ex:
+        LOGGER.error("Invalid rulesets: %s", str(ex))
+        raise ParseError(_PARSE_ERROR_MSG) from ex
+
+    if not isinstance(rulesets, list):
+        raise ParseError(_PARSE_ERROR_MSG)
+    return rulesets
+
+
+def _validated_sources(ruleset: object) -> list:
+    """Return the sources list from a ruleset after validation."""
+    if not isinstance(ruleset, dict):
+        raise ParseError(_PARSE_ERROR_MSG)
+    sources = ruleset.get("sources") or []
+    if not isinstance(sources, list):
+        raise ParseError(_PARSE_ERROR_MSG)
+    return sources
+
+
+def _validate_source(source: object) -> None:
+    """Raise ParseError if a source entry is structurally invalid."""
+    if not isinstance(source, dict):
+        raise ParseError(_PARSE_ERROR_MSG)
+    if _get_source_type(source) is None:
+        raise ParseError(_PARSE_ERROR_MSG)
+    if "name" in source and not isinstance(source["name"], str):
+        raise ParseError(_PARSE_ERROR_MSG)
 
 
 def build_source_list(rulesets_data: str) -> list[dict]:
-    """
-    Parse rulesets to build sources.
+    """Parse rulesets to build sources.
 
     Args:
         rulesets_data: rulesets of the rulebook
 
     Returns:
         list of sources defined in the ruleset
-
     """
-    results = []
     if rulesets_data is None:
-        return results
+        return []
 
-    try:
-        rulesets = yaml.safe_load(rulesets_data)
-    except yaml.MarkedYAMLError as ex:
-        LOGGER.error("Invalid rulesets: %s", str(ex))
-        raise ParseError("Failed to parse rulebook data") from ex
-
+    rulesets = _parse_rulesets(rulesets_data)
     rulebook_hash = get_rulebook_hash(rulesets_data)
-    current_names = set()
+    current_names: set[str] = set()
     counter = 1
+    results: list[dict] = []
 
     for ruleset in rulesets:
-        for source in ruleset.get("sources", []):
+        for source in _validated_sources(ruleset):
+            _validate_source(source)
             default_name = f"{DEFAULT_SOURCE_NAME_PREFIX}{counter}"
             counter += 1
 
-            src_record = {}
-            src_record["rulebook_hash"] = rulebook_hash
-            src_record["source_info"] = source
+            src_record: dict = {
+                "rulebook_hash": rulebook_hash,
+                "source_info": source,
+            }
 
             src_name = source.get("name", default_name)
             if src_name in current_names:
@@ -64,10 +103,77 @@ def build_source_list(rulesets_data: str) -> list[dict]:
                 src_record["name"] = source.get("name", default_name)
 
             current_names.add(src_record["name"])
-
             results.append(src_record)
 
     return results
+
+
+def _get_source_type(source: dict) -> str | None:
+    """Extract the plugin type key from a source dict.
+
+    The source dict uses the plugin type as a top-level key
+    (e.g. ``ansible.eda.webhook``).  ``name`` and ``filters``
+    are reserved sibling keys, so the type is the first key
+    that is neither of those.
+    """
+    for key in source:
+        if key not in ("name", "filters"):
+            return key
+    return None
+
+
+def rulebook_sources_unchanged(old_rulesets: str, new_rulesets: str) -> bool:
+    """Return True when rulebook sources are structurally identical.
+
+    Compares only the attributes that define source identity:
+      1. Positional index (source count and order)
+      2. Plugin type key (e.g. ``ansible.eda.webhook``)
+
+    Source args, filters, and explicit names are intentionally
+    ignored.  Args and filters are swapped/preserved independently
+    by ``swap_event_stream_sources()``.  Name changes are handled
+    separately via ``build_source_name_updates()``.
+    """
+    try:
+        old_sources = build_source_list(old_rulesets)
+        new_sources = build_source_list(new_rulesets)
+    except ParseError:
+        LOGGER.warning("Failed to parse rulebook sources for comparison")
+        return False
+
+    if len(old_sources) != len(new_sources):
+        return False
+
+    for old, new in zip(old_sources, new_sources):
+        old_type = _get_source_type(old["source_info"])
+        new_type = _get_source_type(new["source_info"])
+        if old_type != new_type:
+            return False
+
+    return True
+
+
+def build_source_name_updates(
+    old_rulesets: str, new_rulesets: str
+) -> dict[str, str]:
+    """Return a mapping of old source name to new source name.
+
+    Only includes entries where the resolved name actually changed.
+    Positional pairing uses the same ordering as ``build_source_list()``.
+    Callers should invoke ``rulebook_sources_unchanged()`` first to
+    confirm the source structure is compatible.
+    """
+    try:
+        old_sources = build_source_list(old_rulesets)
+        new_sources = build_source_list(new_rulesets)
+    except ParseError:
+        return {}
+
+    updates: dict[str, str] = {}
+    for old, new in zip(old_sources, new_sources):
+        if old["name"] != new["name"]:
+            updates[old["name"]] = new["name"]
+    return updates
 
 
 def get_rulebook_hash(rulebook: str) -> str:
@@ -80,6 +186,40 @@ def get_rulebook_hash(rulebook: str) -> str:
     Returns: the hexadecimal representation of the hash
     """
     return hashlib.sha256((rulebook or "").encode("utf-8")).hexdigest()
+
+
+def build_rulebook_with_event_streams(validated_data: dict) -> str:
+    """Build swapped rulesets from source_mappings and rulebook content.
+
+    Queries EventStream objects referenced in ``source_mappings``,
+    builds the ``pg_listener`` source configuration for each, and
+    calls ``swap_event_stream_sources`` to produce the final rulesets.
+
+    Args:
+        validated_data: dict with ``source_mappings`` (YAML string)
+            and ``rulebook_rulesets`` (YAML string) keys.
+
+    Returns:
+        YAML string with event-stream sources swapped in.
+    """
+    from aap_eda.core import models
+
+    source_mappings = yaml.safe_load(validated_data["source_mappings"])
+    sources_info = {}
+    for source_map in source_mappings:
+        event_stream_id = source_map.get("event_stream_id")
+        obj = models.EventStream.objects.get(id=event_stream_id)
+
+        sources_info[obj.name] = {
+            "ansible.eda.pg_listener": {
+                "dsn": PG_NOTIFY_DSN,
+                "channels": [obj.channel_name],
+            },
+        }
+
+    return swap_event_stream_sources(
+        validated_data["rulebook_rulesets"], sources_info, source_mappings
+    )
 
 
 def swap_event_stream_sources(
--- a/src/aap_eda/services/project/imports.py
+++ b/src/aap_eda/services/project/imports.py
@@ -26,6 +26,7 @@ from django.conf import settings
 from django.core import exceptions
 
 from aap_eda.core import models
+from aap_eda.core.enums import ActivationStatus
 from aap_eda.core.types import StrPath
 from aap_eda.core.utils.rulebook import get_rulebook_hash
 from aap_eda.services.project.scm import ScmRepository
@@ -56,7 +57,7 @@ class MalformedError(Exception):
 
 
 def _project_import_wrapper(
-    func: Callable[[ProjectImportService, models.Project], None]
+    func: Callable[[ProjectImportService, models.Project], None],
 ):
     @wraps(func)
     def wrapper(self: ProjectImportService, project: models.Project):
@@ -206,25 +207,92 @@ class ProjectImportService:
         rulebook.rulesets = rulebook_info.raw_content
         rulebook.rulesets_sha256 = new_sha256
         rulebook.save(update_fields=["rulesets", "rulesets_sha256"])
-        # Update activations without auto-restart. Those with
-        # restart_on_project_update=True are handled by
-        # _auto_restart_activations which needs to detect
-        # the change via SHA256 comparison.
-        base_qs = models.Activation.objects.filter(
+        self._update_activations_for_rulebook(rulebook, new_sha256, git_hash)
+
+    def _update_activations_for_rulebook(
+        self,
+        rulebook: models.Rulebook,
+        new_sha256: str,
+        git_hash: str,
+    ):
+        """Update activations after a rulebook content change.
+
+        For activations without auto-restart and without source_mappings,
+        update the cached rulesets and hash immediately (bulk).
+
+        For activations with source_mappings, only update the embedded
+        rulebook_hash inside source_mappings so that the enable/restart
+        validation does not reject the activation with a stale hash.
+        The cached rulebook_rulesets are intentionally left unchanged
+        because they contain the post-swap content produced by
+        swap_event_stream_sources() at activation creation time.
+
+        Activations with restart_on_project_update=True and no
+        source_mappings are handled entirely by
+        _auto_restart_activations() and are not touched here.
+        """
+        models.Activation.objects.filter(
             rulebook=rulebook,
             restart_on_project_update=False,
-        )
-        base_qs.filter(source_mappings="").update(
+            source_mappings="",
+        ).update(
             rulebook_rulesets=rulebook.rulesets,
             rulebook_rulesets_sha256=new_sha256,
             git_hash=git_hash,
         )
-        # Source-mapped activations: preserve SHA256 for stale
-        # warning detection (AAP-72873).
-        base_qs.exclude(source_mappings="").update(
-            rulebook_rulesets=rulebook.rulesets,
-            git_hash=git_hash,
-        )
+
+        mapped_activations = models.Activation.objects.filter(
+            rulebook=rulebook,
+        ).exclude(source_mappings="")
+
+        for activation in mapped_activations:
+            updated = self._update_source_mappings_hash(
+                activation.source_mappings, new_sha256
+            )
+            if updated is not None:
+                activation.source_mappings = updated
+                activation.save(update_fields=["source_mappings"])
+            else:
+                logger.error(
+                    "Activation %s (id=%s) has unparseable "
+                    "source_mappings; marking as error",
+                    activation.name,
+                    activation.id,
+                )
+                activation.status = ActivationStatus.ERROR
+                activation.status_message = (
+                    "source_mappings could not be updated after "
+                    "project sync. Please reattach event streams."
+                )
+                activation.save(update_fields=["status", "status_message"])
+
+    @staticmethod
+    def _update_source_mappings_hash(
+        source_mappings: str, new_hash: str
+    ) -> str | None:
+        """Rewrite the rulebook_hash in each source mapping entry.
+
+        Returns the updated YAML string, or None if parsing failed.
+        """
+        try:
+            mappings = yaml.safe_load(source_mappings)
+        except yaml.YAMLError:
+            logger.warning("Failed to parse source_mappings for hash update")
+            return None
+
+        if not isinstance(mappings, list):
+            return None
+
+        for mapping in mappings:
+            if not isinstance(mapping, dict):
+                logger.warning(
+                    "source_mappings contains non-dict entry: %s",
+                    type(mapping),
+                )
+                return None
+            mapping["rulebook_hash"] = new_hash
+
+        return yaml.dump(mappings, default_flow_style=False)
 
     def _find_rulebooks(self, repo: StrPath) -> Iterator[RulebookInfo]:
         rulebooks_dir = None
--- a/src/aap_eda/services/project/imports.py
+++ b/src/aap_eda/services/project/imports.py
@@ -26,8 +26,14 @@ from django.conf import settings
 from django.core import exceptions
 
 from aap_eda.core import models
+from aap_eda.core.enums import ActivationStatus
 from aap_eda.core.types import StrPath
-from aap_eda.core.utils.rulebook import get_rulebook_hash
+from aap_eda.core.utils.rulebook import (
+    build_rulebook_with_event_streams,
+    build_source_name_updates,
+    get_rulebook_hash,
+    rulebook_sources_unchanged,
+)
 from aap_eda.services.project.scm import ScmRepository
 
 logger = logging.getLogger(__name__)
@@ -56,7 +62,7 @@ class MalformedError(Exception):
 
 
 def _project_import_wrapper(
-    func: Callable[[ProjectImportService, models.Project], None]
+    func: Callable[[ProjectImportService, models.Project], None],
 ):
     @wraps(func)
     def wrapper(self: ProjectImportService, project: models.Project):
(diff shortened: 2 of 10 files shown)


hsong-rh:
@amasolov Thanks for the contribution! The skip guard this PR removes was added intentionally in AAP-65811 (PR #1480).

The reason auto-restart is skipped for activations with `source_mappings` is that `source_mappings` are tied to the OLD rulebook version. After sync, the new rulebook may have different source names, making the existing mappings invalid. More critically, `swap_event_stream_sources()` replaces `ansible.eda.webhook` sources with `ansible.eda.pg_listener` at activation creation time and persists the swapped content to `activation.rulebook_rulesets`. Overwriting `rulebook_rulesets` with the raw new rulebook content (as this PR does at line 288) destroys that swap — `ansible-rulebook` would then try to bind to the original webhook port, which is the exact AAP-72873 crash this is trying to fix.

Updating `rulebook_hash` inside `source_mappings` keeps the metadata in sync but doesn't re-apply the source swap. The activation would restart with unswapped rulesets containing `ansible.eda.webhook` + port binding instead of `ansible.eda.pg_listener`.

The designed solution (AAP-65811) is: skip auto-restart for source-mapped activations, leave them running with old-but-working swapped content, and surface a computed `warnings` field on the activation detail API when SHA256 divergence is detected. The user then resolves by updating the activation through the normal serializer flow, which re-applies the swap safely. This is already implemented and live in `ActivationReadSerializer.to_representation()`.

hsong-rh:
@AlexSCorey, could you please check if the warning message is visible in the UI?

amasolov:
@hsong-rh Thank you for the detailed review. You're absolutely right that the skip guard was intentional, and I missed the `swap_event_stream_sources()` contract. Overwriting `rulebook_rulesets` for source-mapped activations would destroy the `pg_listener` swap.

I've revised the approach in d7427ea:

- **Restored the auto-restart skip guard** for source-mapped activations
- **Only update `rulebook_hash` inside `source_mappings`** during project sync (not `rulebook_rulesets`) so the post-swap content is preserved
- Non-mapped activations without auto-restart still get the bulk rulesets update as before
- Activations with unparseable `source_mappings` are now marked as ERROR

This fixes the original bug (stale `rulebook_hash` in `source_mappings` blocks manual enable/restart via `_validate_sources_with_event_streams`) while preserving the AAP-65811 warnings design and the swap contract.

hsong-rh (dismissed):

B-Whitt:
Missing from description: 

- Tracked in Jira https://redhat.atlassian.net/browse/AAP-74014
- Affected version: AAP 2.6

src/aap_eda/services/project/imports.py
```diff
@@ -206,19 +207,88 @@ def _sync_rulebook(
...
+            mappings = yaml.safe_load(source_mappings)
+        except yaml.YAMLError:
+            logger.warning("Failed to parse source_mappings for hash update")
+            return None
+
+        if not isinstance(mappings, list):
+            return None
+
+        for mapping in mappings:
+            mapping["rulebook_hash"] = new_hash
```
wfealdel, line 414: Hi @amasolov!
Are we confident that the list does not contains non-dict types, maybe it's safe to check for types before the assignation to avoid a `TypeError` exception:
```python
if not isinstance(mapping, dict):
    logger.warning("source_mappings contains non-dict entry: %s", type(mapping))
    return None
mapping["rulebook_hash"] = new_hash

Comment (amasolov) [reply src/aap_eda/services/project/imports.py:414]: Good catch @wfealdel thanks! Added the guard in `d695eaa`. It now returns None for non-dict entries (which triggers the existing error-handling path that marks the activation as ERROR with a descriptive message).

**Comment by ttuffin** [comment]
@mkanoor @hsong-rh please review this

**Review by mkanoor — COMMENTED** [review]
@amasolov The Source Mappings are a manual process so that a user can decide which sources in a ruleset have been swapped out. A rulebook can have multiple rulesets and each ruleset can have multiple sources. So when a user does the source mapping to event streams its possible that there are other sources in the rulebook say like Kafka or Azure in addition to the **ansible.eda.webhook** and they might selectively pick which one they swap out and which ones they leave alone. The source mapping captures which ones have been  swapped out. Once a rulebook has changed we dont know which section of the rulebook has changed is it the sources section or the rules section. So we use the whole rulebooks hash to detect changes and request users help in re mapping the sources. Its possible that they might have switched to Kafka and they might not want to use EventStreams anymore. Automatically swapping the sources will cover some specific cases but it is not a generic solution. We dont have an option that says Always swap out the sources when the rulebook is updated. If there is only 1 source and its always **ansible.eda.webhook** the swap might be easy but with multiple sources it becomes error prone. We had discussed this with the product team to have an option that says always create the new source mappings when the rulebook hash changes or do it automatically if there is only 1 source and it is **ansible.eda.webhook**.
The other point is we dont want to update the Sources Mappings rulebook_hash during a Project Update since that is not what is in the rulebook copy that is currently being used for the running Activation. The running Activation has a copy of the rulebook stored as a data blob when we overwrite the hash in the source mappings we are saying that the rulebook in the Database is the same as the copy stored inside of the Rulebook which is not true.

``` 
source_mappings: 
   - source_name: __SOURCE_1  
     event_stream_id: 1  
     event_stream_name: demo  
     rulebook_hash: 391c7468ab3d081e9e1ab4c121ac5979018092f053cf1a4fb9257752d83a42b0',
   - source_name: __SOURCE_2  
     event_stream_id: 100  
     event_stream_name: demo100  
     rulebook_hash: 391c7468ab3d081e9e1ab4c121ac5979018092f053cf1a4fb9257752d83a42b0',
```

The rulebook_hash is the hash when the source mapping was created when the user does the re-mapping this will adjust if we overwrite it with what is in the Database the source mapping check will be redundant.

**Comment by amasolov** [comment]
@mkanoor Thanks for the detailed review — this helps clarify the design intent behind `rulebook_hash` in `source_mappings`.

I agree that a blanket hash rewrite during project sync is too aggressive. The hash is meant to record the rulebook state at mapping time, and silently updating it would undermine the signal that the user needs to re-map when sources have actually changed.

That said, #1542 highlights a real pain point: when a rulebook update changes only rules (or other non-source content), users are blocked from restart with "Please reattach event streams" even though their source mappings are still valid. The existing `warnings` field compares `rulebook_rulesets_sha256` divergence, but the hard validation in `_validate_sources_with_event_streams()` still rejects enable/restart based on the stale `rulebook_hash` in `source_mappings`.

**Proposed narrower approach (Option B):** only auto-update `rulebook_hash` during project sync when the **source structure for mapped sources is unchanged**, using a structural comparison rather than a full rulebook hash match.

Concretely, on sync we would:

1. Parse the activation's `source_mappings` to get the list of mapped `source_name` values.
2. Call `build_source_list()` on the new rulebook content and the previous rulebook content (or the activation's pre-sync cached rulesets).
3. For each mapped source, compare structural identity: source name, plugin type (e.g. `ansible.eda.webhook` vs `ansible.eda.kafka`), and relevant config keys — ignoring rules section changes entirely.
4. **If all mapped sources are structurally identical:** update `rulebook_hash` in `source_mappings` so enable/restart validation passes without manual re-attachment.
5. **If any mapped source changed structurally** (renamed, removed, type changed, or new unmapped sources added that affect the mapping contract): leave `rulebook_hash` stale and rely on the existing warnings + manual re-mapping flow through the serializer.

This preserves the manual mapping contract for the cases you described (multiple sources, selective swapping, user switching from Event Streams to Kafka, etc.) while fixing the case where only rules changed and the mappings are still valid.

Does this align with the product direction you mentioned (auto-remap only when safe)? Happy to revise the PR along these lines or close it if the team prefers to wait for the explicit product option instead.

**Comment by mkanoor** [comment]
@amasolov I agree with the direction you are taking it would have to be an exhaustive comparison of the sources to see if they have changed. We would have to difference the rulebooks sources section per ruleset and since not all sources have names we have to create temporary names based on the positional index. There are a couple of issues I am not sure how that has been addressed in your PR.

When a Project sync is run we update the aap_eda.core.models.Rulebook#rulesets with the latest data, here we will update the old rulebook and that old data is lost
So before we update the Rulebook#rulesets we will have to compare the old source data with the new source data to see if they are the same. Maybe we need to save the old source data in a separate column or the old rulebook in its entirety.

The Activation model aap_eda.core.models.Activation#rulebook_rulesets has a copy of the old data except that the sources have been swapped. So this cloned data cannot be used for comparison since we have swapped out the source.

If the Project Sync fails for some reason we might have to undo the changes since the commit sha has to match what is stored in the database. So the whole call would have to be wrapped in a transaction.

I do see there is special logic in _update_activations_for_rulebook but that is not checking the sources old value and new value. We would have to pass in the old rulebook#rulesets from the Database and the new rulebook#rulesets that we are trying to update to and compare the sources and if the sources have not been updated then we could safely reuse the source mappings.

**Comment by amasolov** [comment]
@mkanoor Thanks for the follow-up — agreed on all points. I've revised the PR in 8d689a0 to implement the structural comparison approach:

**What changed**

1. **Capture old content before overwrite:** `_sync_rulebook()` now saves `old_rulesets = rulebook.rulesets` before updating `Rulebook.rulesets`, and passes it to `_update_activations_for_rulebook()`.

2. **Structural source comparison:** Added `rulebook_sources_unchanged()` in `core/utils/rulebook.py`. It uses `build_source_list()` on both old and new content, comparing assigned source names and `source_info` dicts per position. Rules section changes are ignored.

3. **Gated hash update:** `rulebook_hash` inside `source_mappings` is only rewritten when `rulebook_sources_unchanged(old_rulesets, new_rulesets)` is true. When sources change, the hash is left stale so the existing validation and warnings flow still prompts re-mapping.

4. **No activation blob comparison:** `Activation.rulebook_rulesets` is not used for comparison (still holds post-swap content).

5. **Transactions:** `_sync_project_no_lock()` already wraps `ProjectImportService().sync_project()` in `transaction.atomic()`, so a failed sync rolls back rulebook and activation updates together.

**Tests added**

- Unit tests for `rulebook_sources_unchanged()` (rules-only change vs source config change)
- Integration test: rules-only sync updates hash (`project-02` → `project-07`)
- Integration test: source config change preserves stale hash (`project-02` → `project-08`)
- `_sync_rulebook` direct tests for both paths

Let me know if the comparison should be scoped further (e.g. only mapped `source_name` entries rather than all sources globally).

File: src/aap_eda/core/utils/rulebook.py [hunk 2]
```diff
@@ -70,6 +70,35 @@ def build_source_list(rulesets_data: str) -> list[dict]:
...
+) -> bool:
+    """Return True when rulebook sources are structurally identical.
+
+    Uses the same positional naming logic as build_source_list().
+    Rules section changes are ignored.
+    """
+    try:
+        old_sources = build_source_list(old_rulesets)
+        new_sources = build_source_list(new_rulesets)
+    except ParseError:
```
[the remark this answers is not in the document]
amasolov: Addressed in 746a3b4: `build_source_list()` now validates that parsed rulesets are a list of dicts with list-of-dict sources before calling `.get()`. Structurally invalid YAML (e.g. plain strings) raises `ParseError`, so `rulebook_sources_unchanged()` returns `False` safely. Added `test_build_source_list_raises_for_structurally_invalid_yaml`.

src/aap_eda/services/project/imports.py
```diff
@@ -203,29 +207,116 @@ def _sync_rulebook(
...
+                    "Skipping source_mappings hash update for activation "
+                    "%s (id=%s): rulebook sources changed during sync",
+                    activation.name,
+                    activation.id,
+                )
+                continue
+
+            updated = self._update_source_mappings_hash(
+                activation.source_mappings, new_sha256
+            )
```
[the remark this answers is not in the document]
amasolov: Addressed in 746a3b4: mapped activations now validate `source_mappings` via `_parse_source_mappings()` before the `sources_unchanged` branch. Malformed mappings are marked `ActivationStatus.ERROR` in both paths; valid mappings with source changes still keep a stale hash. Added `test_sync_marks_error_on_invalid_source_mappings_when_sources_change` (sync to `project-08`).

amasolov:
@coderabbitai Both actionable comments from the `8d689a0` review are addressed in `746a3b4`:

1. **`build_source_list()` invalid YAML** — structural validation + `ParseError` for non-list/dict shapes
2. **Malformed `source_mappings` on source-change skip** — validate before branch; ERROR in both paths

Please re-review when convenient.

mkanoor:
@amasolov Great work looks good. Couple more subtleties
* A source can have zero or more [filters](https://github.com/ansible/ansible-rulebook/blob/4d6e3f78121b2c6d5eeea0a803f3388c87deae73/tests/examples/93_event_splitter.yml#L16) underneath it, that is **not** swapped out. These come in very handy when we have to deal with Prometheus/Alerts which come in as an array of events and we use a source filter to split these out into individual events.

* The Source **Index** and Source **Type** and Source **Name** if present are the key comparison points. Even the source args can be ignored when doing the comparison. Since the entire Source args gets swapped out with the ansible.eda.pg_listener.

For single source rulebook this is easy to resolve but if the Type changes then we have to raise our manual update needed message
For multiple source rulebook across multiple [rulesets](https://github.com/ansible/ansible-rulebook/blob/main/tests/examples/36_multiple_rulesets_both_fired.yml) we might have to ensure that the type and index dont change.

So during differencing the key compartors are
1. Index
2. Source Type
3. Optionally Name if they have provided one, if they just changed the name we can fix that in the source mappings. This source name is used when doing reporting on Rule Audit.

The things to ignore are

* Source Args
* Source Filters

We might need tests to test these corner cases.

The Source Node has some quirks:
 * It doesn't have a separate node called args.
 * So the type is the key and every child key after that is args unless you get to a filters which is an array.
 * The name node is also optional hence we have to resort to creating names using the index format e.g __SOURCE_1
 
Another suggestion would be to move this Rulebook Differencing into a separate file we could use it in other places.


I appreciate all your efforts in getting this to a closure.

amasolov:
@mkanoor Thanks for the detailed breakdown. Addressed in `cb75851` + `685701e`:

**Source comparison narrowed to index + type only**

`rulebook_sources_unchanged()` now compares only:
1. Source count (positional index)
2. Plugin type key (first key that is not `name`/`filters`)

`_get_source_type()` helper extracts the plugin key using the same convention as `_updated_event_stream_source()`.

Source args, filters, and explicit names are all ignored during comparison.

**Name changes auto-fixed in source_mappings**

New `build_source_name_updates()` helper computes old-name -> new-name for positionally paired sources. `_update_source_mappings_hash()` now accepts an optional `name_updates` dict and rewrites `source_name` entries alongside `rulebook_hash`.

**Test coverage**

Unit tests (`test_rulebook.py`):
- args-only change -> unchanged
- filter-only change -> unchanged
- name added -> unchanged
- type change -> changed
- source added -> changed
- `build_source_name_updates()` for no-change, rename, and invalid YAML
- `_get_source_type()` for webhook, named+filtered, and empty

Integration tests:
- `test_sync_updates_source_name_in_mappings_on_rename` (project-02 -> project-09): verifies both `rulebook_hash` and `source_name` are updated
- `test_sync_rulebook_updates_source_name_on_rename` (`_sync_rulebook` direct): same via direct call
- project-08 fixture updated to use type change (`ansible.eda.range` -> `ansible.eda.kafka`)

**Re: separate differencing file** - I agree this would be cleaner for reuse. Happy to extract `_get_source_type`, `rulebook_sources_unchanged`, and `build_source_name_updates` into `src/aap_eda/core/utils/rulebook_diff.py` in a follow-up if you prefer to keep this PR focused.

src/aap_eda/core/utils/rulebook.py
```diff
@@ -44,12 +44,24 @@ def build_source_list(rulesets_data: str) -> list[dict]:
...
+        if not isinstance(ruleset, dict):
+            raise ParseError("Failed to parse rulebook data")
+        sources = ruleset.get("sources", [])
+        if sources is None:
+            sources = []
+        if not isinstance(sources, list):
+            raise ParseError("Failed to parse rulebook data")
+        for source in sources:
+            if not isinstance(source, dict):
+                raise ParseError("Failed to parse rulebook data")
```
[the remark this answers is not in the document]
amasolov: Addressed in `16be4d3`: added both validations (`_get_source_type(source) is None` and non-string `name`) with `ParseError`, plus unit tests for empty source and `name: []`.

src/aap_eda/services/project/imports.py
```diff
@@ -203,28 +208,162 @@ def _sync_rulebook(
...
+                activation.name,
+                activation.id,
+            )
+            return
+
+        updated = self._update_source_mappings_hash(
+            activation.source_mappings, new_sha256, name_updates
+        )
+        if updated is not None:
+            activation.source_mappings = updated
```
mkanoor: @amasolov Another thing that is missing here is that the Activation#rulebook_rulesets has the ansible.eda.webhook swapped with ansible.eda.pg_listener.
In your rulebook you might have rules and conditions that have changed but the sources remain the same. So in this code we are fixing the source mappings with the new hash but the sources also have to be preserved in Activation#rulebook_rulesets but the rest of the ruleset meaning rules/conditions/actions have to be updated. Otherwise if we just replace the entire rulebook we will loose the ansible.eda.pg_listener changes. This is currently handled by https://github.com/amasolov/eda-server/blob/37a75f7a90e454bb07ae13e4381bf3b1087760d2/src/aap_eda/core/utils/rulebook.py#L85 so we would have to update the rulebook in the Activation after we have updated it here

amasolov:
**SonarCloud Quality Gate note**

The 7 remaining violations and the 0% coverage metric are all pre-existing or infrastructure-related, not introduced by this PR:

| Issue | File | Notes |
|-------|------|-------|
| S3776 Cognitive Complexity 16 | `imports.py:_find_rulebooks` | Pre-existing function, untouched by this PR |
| S1481 unused `queue` (x2) | `tasks/project.py:56,125` | Pre-existing, not modified by this PR |
| S9073 composite assertion (x4) | `test_purge_log_records.py` | Pre-existing, not modified by this PR |

Coverage at 0.0% appears to be a fork PR artifact (CI does not upload coverage reports for external PRs). The previous SonarCloud scan in May passed with 96.6% coverage and 0 new issues on the same PR.

src/aap_eda/services/project/imports.py
amasolov: Good catch. Addressed in \`44804cd\`.

Added \`rebuild_swapped_rulesets(new_rulesets, old_swapped_rulesets)\` which merges by positional index:

- **Rules, conditions, actions**: from the new rulebook
- **Sources** (e.g. \`pg_listener\` type + args + name): from the old swapped rulesets
- **Filters**: from the new rulebook (in case they changed)

\`_update_mapped_activation()\` now calls this after updating the source_mappings hash and saves the rebuilt content to \`activation.rulebook_rulesets\`. \`rulebook_rulesets_sha256\` is intentionally left unchanged (it tracks the original pre-swap hash used for validation).

Unit tests cover: preserved swapped sources, updated rules, new filters applied, old filters dropped when new rulebook has none.

Integration tests updated to use realistic swapped YAML (with \`pg_listener\`) and verify both hash updates and rulesets rebuilding.

mkanoor:
@amasolov I have tested the PR it looks good but it doesn't restart the Activation due to this check
https://github.com/ansible/eda-server/blob/d05189a94e96a1d4d4fa88de29fe750c0cb0e8dd/src/aap_eda/tasks/project.py#L279
Since it returns early without restarting.
Since the copy of the Rulebook has been updated by your code we also need to guard against this code which ends up undoing your changes.
https://github.com/ansible/eda-server/blob/d05189a94e96a1d4d4fa88de29fe750c0cb0e8dd/src/aap_eda/tasks/project.py#L287. This code should only handle the Activations without the source_mappings.
There was another issue that I found unrelated to your changes. Which is in JIRA https://redhat.atlassian.net/browse/AAP-88826
Another thing that I noticed is if an Activation was stuck with the Source Mapping error and manual update we cannot fix it and restart it. That might have to be tackled outside of this PR

src/aap_eda/services/project/imports.py
```diff
@@ -203,28 +209,208 @@ def _sync_rulebook(
...
+        old_rulesets = rulebook.rulesets
         rulebook.rulesets = rulebook_info.raw_content
         rulebook.rulesets_sha256 = new_sha256
         rulebook.save(update_fields=["rulesets", "rulesets_sha256"])
-        # Update activations without auto-restart. Those with
-        # restart_on_project_update=True are handled by
-        # _auto_restart_activations which needs to detect
-        # the change via SHA256 comparison.
-        base_qs = models.Activation.objects.filter(
+        self._update_activations_for_rulebook(
```
mkanoor, line 217: @amasolov 
Here we need the update from db
```
rulebook.refresh_from_db()
```
Otherwise it is using the old object before the update and that will be used to update the rulebook

mkanoor:
@amasolov There is an issue on this line https://github.com/ansible/eda-server/blob/d05189a94e96a1d4d4fa88de29fe750c0cb0e8dd/src/aap_eda/tasks/project.py#L366 it unconditionally overwrites the Activation.rulebook_rulesets This would have to be updated to something like this
```
# For activations with source_mappings, rebuild with event streams
       366 +          if activation.source_mappings:
       367 +              from aap_eda.core.utils.rulebook import build_rulebook_with_event_streams
       368 +              try:
       369 +                  activation.rulebook_rulesets = build_rulebook_with_event_streams({
       370 +                      "source_mappings": activation.source_mappings,
       371 +                      "rulebook_rulesets": rulebook.rulesets,
       372 +                  })
       373 +                  activation.rulebook_rulesets_sha256 = rulebook.rulesets_sha256
       374 +              except Exception:
       375 +                  logger.warning(
       376 +                      f"Failed to rebuild rulebook_rulesets for activation "
       377 +                      f"'{activation.name}' with event streams; keeping existing",
       378 +                      exc_info=True,
       379 +                  )
       380 +          else:
       381 +              # For unmapped activations, use raw rulebook
       382 +              activation.rulebook_rulesets = rulebook.rulesets or ""
       383                activation.rulebook_rulesets_sha256 = rulebook.rulesets_sha256
       384 +  
```

amasolov:
## Fixes from @mkanoor's review

Squashed the following fixes into the single commit:

1. **`rulebook.refresh_from_db()` after save** (`imports.py:216`) - Ensures the rulebook object passed to `_update_activations_for_rulebook` reflects the latest DB state, preventing stale data from being used during activation updates.

2. **Conditional rebuild in `_update_activation_content()`** (`tasks/project.py`) - When `activation.source_mappings` exists, rebuild `rulebook_rulesets` via `build_rulebook_with_event_streams()` instead of unconditionally overwriting with raw rulebook content. This preserves `pg_listener` source swaps during sync-triggered restarts (the "update revision on launch" path).

## Testing

### Environment
- AAP 2.7 podman-based deployment on AWS EC2 (m5.2xlarge, RHEL 9)
- Patched EDA containers (`eda-api`, `eda-worker`, `eda-scheduler`) with the PR changes

### Integration tests (44/44 passed)
Ran a comprehensive test suite inside the EDA worker container covering:

- `_parse_rulesets` helper and `ParseError` handling
- `build_source_list()` structural validation (isinstance checks, empty sources, non-string names)
- `rulebook_sources_unchanged()` structural comparison (index + plugin type only)
- `build_source_name_updates()` rename detection
- `PG_NOTIFY_DSN` template string
- `swap_event_stream_sources()` with real source mappings
- `build_rulebook_with_event_streams()` shared utility (queries EventStream, builds pg_listener config, calls swap)
- `ProjectImportService._update_activations_for_rulebook()`:
  - Source-mapped activations: hash update, source name update, structural comparison
  - Unmapped activations: no source_mappings changes
  - Malformed source_mappings: activation marked ERROR
  - Source structure changes: activation correctly flagged for re-mapping

### Playwright UI tests (6/6 passed)
Ran against the live AAP 2.7 UI from local Playwright:

1. **EDA Projects page** - "EDA Sample Project" visible with "Completed" status
2. **Rulebook Activations page** - Both source-mapped and regular activations listed
3. **Activation detail** - Shows correct rulebook (`webhook_with_vars.yml`), event stream (`Test Webhook Stream`), status "Pending" (not Error)
4. **Event Streams page** - "Test Webhook Stream" visible
5. **Post-sync verification** - Triggered project sync, verified activation source_mappings preserved, status remains "Pending" (not Error)
6. **Full API verification** - Project completed, source-mapped activation has correct `source_mappings` with `rulebook_hash` and `event_stream_id`, regular activation has empty `source_mappings`, rulebook sources correct

All tests confirm that:
- `pg_listener` sources are preserved in `rulebook_rulesets` after project sync
- Source mappings are not stomped when "update revision on launch" triggers a content refresh
- Activations with unchanged source structure correctly get updated `rulebook_hash` and rebuilt rulesets
- Activations with changed source structure are correctly flagged for manual re-mapping

mkanoor:
@amasolov 
The project.py is skipping the Activation when AutoRestart is enabled because of this check
https://github.com/ansible/eda-server/blob/d05189a94e96a1d4d4fa88de29fe750c0cb0e8dd/src/aap_eda/tasks/project.py#L279
This needs to be deleted otherwise the Auto Restart will not restart a Rulebook with EventStreams swapped.
Then on this line we should only do it for non Event Streams rulebooks otherwise it will overwrite what you had changed in the Activation during the project import.
https://github.com/ansible/eda-server/blob/d05189a94e96a1d4d4fa88de29fe750c0cb0e8dd/src/aap_eda/tasks/project.py#L287

This line needs be be updated to
```
if (content_changed or hash_changed) and not activation.source_mappings:
```
The net changes I had were this
```
-    if content_changed and activation.source_mappings:
-        logger.warning(
-            f"Skipping auto-restart for activation "
-            f"'{activation.name}' - has event stream "
-            f"source mappings that need manual update"
-        )
-        return 0
-
-    if content_changed or hash_changed:
+    if (content_changed or hash_changed) and not activation.source_mappings:
```

amasolov:
@mkanoor Good catch on the auto-restart path, thank you. Applied your suggested changes in the latest push:

1. Removed the early return that was skipping auto-restart for source-mapped activations
2. Changed the content update to only overwrite `rulebook_rulesets` and `rulebook_rulesets_sha256` for non-source-mapped activations, preserving the pg_listener-swapped rulesets

I kept `git_hash` updating unconditionally (for both mapped and unmapped) to prevent hash-only changes from re-triggering on every subsequent check.

Updated tests:
- `test_auto_restart_restarts_source_mapped_without_overwrite`: verifies source-mapped activations DO get restarted but rulesets are NOT overwritten
- `test_auto_restart_mixed_source_mappings_activations`: both activations now restart (`call_count == 2`), only unmapped gets rulesets overwritten
- `test_update_activation_content_rebuilds_for_source_mapped`: new test with real EventStream verifying the happy path for `_update_activation_content`
- `test_update_activation_content_preserves_on_rebuild_failure`: verifies both rulesets and SHA256 are preserved when rebuild fails

mkanoor:
@amasolov Thanks for all the effort on getting this thru. I tested the different scenarios
1. Update the rulebook with AutoRestart enabled with EventStreams mapped, no source changes the Activation Restarts
2. Upadte the rulebook with AutoRestart enabled with EventStreams mapped, with changed source, fails the Activation once manually updated and restarted it starts properly.
3. With the project "update revisions on launch" enabled successfully restarts an EventStream mapped Activation (AAP-88826)
4. Non EventStream mapped Activations cleanly update the rulebook.

💯 🥇 ✅

mkanoor (dismissed):

src/aap_eda/tasks/project.py
```diff
@@ -276,25 +276,16 @@ def _check_and_restart_activation(
...
+        update_fields = ["git_hash"]
         activation.git_hash = current_git_hash
-        activation.save(
-            update_fields=[
-                "rulebook_rulesets",
-                "rulebook_rulesets_sha256",
-                "git_hash",
-            ]
-        )
+        if not activation.source_mappings:
```
hsong-rh, line 283: **Source-mapped auto-restart activations restart on every project sync after the first content change**

After the first rulebook-content-changing sync, `activation.rulebook_rulesets_sha256` stays at its pre-sync value indefinitely — `_update_mapped_activation` omits it from `update_fields`, and here `rulebook_rulesets_sha256 = current_sha256` is gated behind `if not activation.source_mappings`. On every subsequent project sync (even when the rulebook is byte-identical since the last change), `_check_and_restart_activation` computes `content_changed = old_hash != new_hash = True` and triggers a restart, killing in-flight event processing on every sync.

The `_update_activation_content` resume path (line ~369) already updates sha256 for source-mapped activations, so `awaiting_project_sync` activations converge correctly — but `restart_on_project_update` activations do not.

**Suggested fix:** Move `rulebook_rulesets_sha256 = current_sha256` outside the `source_mappings` guard. Since `content_changed` is captured before the sha256 write, the current-cycle restart decision is unaffected — only the sha256 field converges so future syncs don't see a spurious diff:

```python
if content_changed or hash_changed:
    update_fields = ["git_hash", "rulebook_rulesets_sha256"]
    activation.git_hash = current_git_hash
    activation.rulebook_rulesets_sha256 = current_sha256
    if not activation.source_mappings:
        activation.rulebook_rulesets = current_rulebook.rulesets or ""
        update_fields.append("rulebook_rulesets")
    activation.save(update_fields=update_fields)
```

`test_auto_restart_restarts_source_mapped_without_overwrite` will need updating (it currently asserts the old sha256 is preserved). A two-sync regression test would also prevent recurrence.

amasolov: Good catch @hsong-rh, applied your suggested fix exactly: moved \`rulebook_rulesets_sha256 = current_sha256\` outside the \`source_mappings\` guard so sha256 always converges. Only \`rulebook_rulesets\` remains protected for source-mapped activations.

Also added the two-sync regression test (\`test_auto_restart_source_mapped_no_spurious_restart_on_second_sync\`) that verifies the first sync triggers a restart but the second identical sync does not.

hsong-rh (dismissed):
👍 Great job!

src/aap_eda/services/project/imports.py
```diff
@@ -203,28 +209,209 @@ def _sync_rulebook(
...
+        )
+        if updated is None:
+            self._mark_activation_source_mappings_error(activation)
+            return
+
+        activation.source_mappings = updated
+        swapped = self._reswap_rulesets(activation, new_rulesets)
+        if swapped is not None:
+            activation.rulebook_rulesets = swapped
+        activation.save(
```
wfealdel, line 317: **`git_hash` not updated for source-mapped activations (regression)**

The old bulk `.update()` set `git_hash` for all non-auto-restart activations, including source-mapped ones. This save omits it — non-auto-restart, source-mapped activations will show a stale git commit hash in the API after sync.

`git_hash` needs to be passed into `_update_mapped_activation`, set on the activation object, and included in `update_fields`.

```suggestion
        activation.save(
            update_fields=[
                "source_mappings",
                "rulebook_rulesets",
                "git_hash",
                "modified_at",
            ]
        )
```

src/aap_eda/services/project/imports.py
```diff
@@ -203,28 +209,209 @@ def _sync_rulebook(
...
+            return
+
+        if not sources_unchanged:
+            logger.info(
+                "Skipping source_mappings hash update for activation "
+                "%s (id=%s): rulebook sources changed during sync",
+                activation.name,
+                activation.id,
+            )
+            return
```
wfealdel, line 303: Same `git_hash` gap here — when sources changed, the old code still updated `git_hash` via the bulk `.update()`. This early return skips it entirely. Consider updating `git_hash` before the return so the activation reflects the current project commit even when source re-mapping is needed.

src/aap_eda/services/project/imports.py
```diff
@@ -203,28 +209,209 @@ def _sync_rulebook(
...
+        """
+        try:
+            return build_rulebook_with_event_streams(
+                {
+                    "source_mappings": activation.source_mappings,
+                    "rulebook_rulesets": new_rulesets,
+                }
+            )
+        except Exception:
+            logger.warning(
```
wfealdel, line 347: **Test coverage gap**: The `except Exception` fallback path (returning `None` so the activation keeps its old rulesets) is tested for `_update_activation_content` in `test_projects.py` (`test_update_activation_content_preserves_on_rebuild_failure`) but not for this call site through `_update_mapped_activation`. Consider adding an integration test where `build_rulebook_with_event_streams` fails during sync (e.g. deleted EventStream) to lock in the graceful degradation behavior.

wfealdel (commented):
@amasolov 
Nice contribution! Well-structured with thorough test coverage and extensive maintainer feedback incorporated. One minor regression and one test gap flagged inline. Overall the structural source comparison approach is solid and aligns with the product direction discussed by @mkanoor.

src/aap_eda/services/project/imports.py
amasolov: Fixed: `git_hash` is now set on the activation and included in `update_fields` in `_update_mapped_activation`. Also added `git_hash` parameter to the method signature so it receives the value from `_update_activations_for_rulebook`.

src/aap_eda/services/project/imports.py
amasolov: Fixed: the early return when sources change structurally now updates `git_hash` before returning, so the activation reflects the current project commit even though source re-mapping is still needed.

src/aap_eda/services/project/imports.py
amasolov: Added `test_sync_rulebook_reswap_failure_keeps_existing_rulesets`: creates an activation with source_mappings referencing an EventStream, deletes the EventStream, then runs `_sync_rulebook`. Verifies the activation keeps its existing rulesets, is not marked ERROR, and `git_hash` is updated.

mkanoor (approved):

hsong-rh (approved):

wfealdel (approved):

ttuffin (approved):

PR #1548, opened 2026-05-04
[AAP-51874] fix: enable persistent DB connections to reduce PostgreSQL log spam

AlexSCorey:
## What

Add `CONN_MAX_AGE=600` and `CONN_HEALTH_CHECKS=True` to the default DATABASES
configuration in `_get_databases_settings()`.

## Why

In containerized AAP deployments, PostgreSQL logs are flooded with:

```
could not receive data from client: Connection reset by peer
```

PID correlation in PostgreSQL logs confirms all 17,403 resets over 19 hours
trace to `user=eda database=eda`. Two sources:

1. **Envoy health checks** (~50%): Hitting `/api/eda/v1/status/` every ~10s.
   Each request calls `connection.ensure_connection()`, opens a fresh DB
   connection (CONN_MAX_AGE defaults to 0), and closes it after the response
   with a TCP RST.

2. **Dispatcherd scheduled tasks** (~50%): `monitor_rulebook_processes` runs
   every 5 seconds, querying the DB from forked worker processes. Connection
   cleanup during process lifecycle produces RSTs.

Setting `CONN_MAX_AGE=600` enables Django to reuse connections for up to 10
minutes, eliminating the open/close churn. `CONN_HEALTH_CHECKS=True` validates
reused connections before use, so stale connections are caught cleanly.

Jira: AAP-51874

## Test plan

- [ ] Deploy containerized AAP and verify PostgreSQL logs are clean after 5+ minutes idle
- [ ] Verify `podman logs postgresql 2>&1 | grep "connection reset"` returns no new entries
- [ ] Confirm actual DB failures are still logged (stop EDA, check PostgreSQL logs)
- [ ] Run existing test suite

## Summary by CodeRabbit

* **Chores**
  * Improved database connection stability by enabling health checks and extending connection reuse time to reduce reconnects.
  * Expected result: fewer transient database interruptions and more reliable application behavior during sustained or burst traffic.


jcraiglo1 (dismissed):
LGTM. Was there any testing done to see why 10 minutes was the best amount of time?

ttuffin:
Further to @jcraiglo1's question, what are other components using? I had a quick look at awx and I don't see a definition for `CONN_MAX_AGE`, so I assume they are using the default of 0. Do we know why this is only an issue with eda-server?

AlexSCorey:
@ttuffin @jcraiglo1 No, I didn't really do testing to determine which CONN_MAX_AGE is best.  I'm going off of recommendations from django and it seems that 600 seconds is the recommended connection age limit.  Currently eda connects and disconnects at minimum every 5 seconds.  awx's task manager connects and disconnects to the db every 20 seconds. The goal with this pr is to reduce the noise in the logs so that customers don't get so bogged down in connect/disconnect messages. To do that we leverage the conn max age, and the conn health check to first, make sure the db is accessible (and fail gracefully if not), and then establish the connection after age out.

@dleehr what are your thoughts?

dleehr (approved):
> @dleehr what are your thoughts?

+1 to @ttuffin's point - Sounds like our other django applications do not trigger the flood of log messages and do not generally require these options, so it would be good to understand why we need this custom postgres configuration for EDA but not the others -  and how we got to 600 - specifically the `CONN_MAX_AGE`.

On `CONN_HEALTH_CHECKS`, it does look like that is enabled in jewel https://github.com/ansible/jewel/blob/devel/aap_gateway_api/defaults.py#L101, so it may be worth syncing with that team if the decision was similar.

Only other thing I noticed - this PR changes the configuration that's applied in `post_load.py`. I haven't looked at EDA installation/configuration before, but a quick scan suggests this is **the** way EDA is configured (e.g. this function creates the `databases` dictionary for settings based on env vars, and the installer sets those env vars )

So my feedback here is that it doesn't look like we provide a way for users (or the installer) to **change** these new settings (`CONN_MAX_AGE` / `CONN_HEALTH_CHECKS`). Since everything else in `databases["default"]` reads an env var by default, I think it would be wise to read env vars for these too. Are we confident that these two options should be hard-coded with these specific values in every EDA installation?

AlexSCorey:
@dleehr Good feedback — pushed a fix. Both settings now read from Dynaconf (env vars) with the same pattern as the rest of the function:

```python
"CONN_MAX_AGE": settings.get("CONN_MAX_AGE", 600),
"CONN_HEALTH_CHECKS": settings.get("CONN_HEALTH_CHECKS", True),
```

Defaults stay the same (600s / True) but operators can override via `CONN_MAX_AGE` and `CONN_HEALTH_CHECKS` env vars if needed.

To address the broader question of "why EDA and not others" — EDA's dispatcherd runs `monitor_rulebook_processes` every 5 seconds from forked workers, and Envoy health checks hit the status endpoint every ~10s. Both create rapid connect/disconnect cycles. AWX's task manager polls every 20s with longer-lived connections, so the default `CONN_MAX_AGE=0` doesn't produce the same volume of resets.

AlexSCorey:
@ttuffin Elijah made a similar change as part of some testing work he did for controller and saw an increase in performance for eda.  See https://redhat.atlassian.net/browse/AAP-64679?focusedCommentId=3587730

jcraiglo1 (approved):

B-Whitt:
## Code Review: [AAP-51874] fix: enable persistent DB connections to reduce PostgreSQL log spam

**Verdict**: READY_FOR_HUMAN_REVIEW
**Score**: 9.7/10 (Functionality 10, Security 10, Quality 9)

### Key Findings

No Critical or Major issues. Two Minor quality items:

1. **[Quality] Missing documentation** — `EDA_CONN_MAX_AGE` and `EDA_CONN_HEALTH_CHECKS` are not documented in `src/aap_eda/settings/defaults.py` alongside the other DB env vars (`DB_HOST`, `DB_PORT`, etc.). Operators won't know these are configurable without reading source.

2. **[Quality] No test coverage** — No tests verify the default values (600 / True) or override behavior of the new settings. The existing `tests/unit/test_settings.py` patterns provide a template.

### Path to 10/10

1. Add two lines to the Database settings docstring in `defaults.py`:
   ```
   * CONN_MAX_AGE - Database connection max age in seconds (default: 600)
   * CONN_HEALTH_CHECKS - Enable DB connection health checks (default: True)
   ```

2. Add tests verifying defaults and env var override using the existing `mock_settings` / `post_loading()` pattern.

### Observations

- Correct pattern adherence — uses same `settings.get("KEY", default)` as all adjacent settings
- Fork safety handled — `dispatcherd_pre_fork.py` closes connections before fork, `CONN_HEALTH_CHECKS=True` validates reused connections
- Sensible defaults — 600s and health checks align with Django recommendations and `jewel` (Gateway)
- Escape hatch — `EDA_CONN_MAX_AGE=0` restores close-after-request behavior
- dleehr's env var configurability concern is addressed by the `settings.get()` pattern

---
*🤖 AI code review — not a substitute for human judgment. [Full review stored locally.]*

AlexSCorey:
@B-Whitt @ttuffin @jcraiglo1 I got expert support for this situation and it seems that making these settings changes won't cause problems for other component, and will resolve much of the noise in the logs.  Here is the last conversation I've had.  https://redhat-internal.slack.com/archives/C044A6R5F51/p1779128961965359

ttuffin (approved):
Thanks for following up on this. I think we have enough information to merge this into main and keep an eye on the devel pipelines. I would hold off merging this change into the release branches before we have properly observed the devel pipeline results.

PR #1551, opened 2026-05-05
fix: increase wait() timeout for multi-message consumer tests

B-Whitt:
## Summary
- Tests that send multiple websocket messages before `wait()` can flake under CI load — the default 1-second timeout cancels the consumer task before all messages are processed
- `test_multiple_rules_for_one_event` was observed failing with `assert 1 == 2` because only one of two actions was processed within the 1s window
- Added explicit `timeout=TIMEOUT` (5s) to the 5 multi-message tests only, matching the pattern already used for `receive_json_from(timeout=TIMEOUT)` in the same file
- Single-message tests keep the default timeout for fast failure detection

## Test plan
- [x] `test_multiple_rules_for_one_event` passes locally
- [x] All 33 tests in `test_consumer.py` pass (72s)
- [x] Changed file passes flake8 cleanly
- [x] CI passes on this PR

🤖 Assisted by Claude Opus

### DIFF
--- a/tests/integration/wsapi/test_consumer.py
+++ b/tests/integration/wsapi/test_consumer.py
@@ -292,7 +292,7 @@ async def test_handle_actions_multiple_firing(
     )
     await ws_communicator.send_json_to(payload1)
     await ws_communicator.send_json_to(payload2)
-    await ws_communicator.wait()
+    await ws_communicator.wait(timeout=TIMEOUT)
 
     assert (await get_audit_rule_count()) == 2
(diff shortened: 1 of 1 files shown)


jcraiglo1 (approved):
Nice!

jcraiglo1 (approved):
Nice and elegant

ptoscano:
@B-Whitt would you please link logs of jobs that fail, and would be fixed by this? nothing potentially wrong with this, I simply want to check what is the issue supposed to be fixed 

thanks!

PR #1558, opened 2026-05-12
fix: context-aware activation worker health check

ttuffin:
Health checks now target the activation's own queue instead of always sampling the first configured queue, preventing false 503s in multi-node podman deployments when an unrelated node is unhealthy.

https://redhat.atlassian.net/browse/AAP-75092

## Summary by CodeRabbit

* **Refactor**
  * Worker health checks are now queue-scoped; activation operations (delete/enable/disable/restart) use the activation’s resolved queue so health is evaluated per queue rather than globally, improving reliability in multi-queue setups.

* **Tests**
  * Added integration tests covering queue-scoped health checks, any-queue healthy short-circuiting, and error raising for unhealthy specified queues.

### DIFF
--- a/src/aap_eda/core/health.py
+++ b/src/aap_eda/core/health.py
@@ -13,6 +13,7 @@
 #  limitations under the License.
 
 import logging
+from typing import Optional
 
 from django.conf import settings
 
@@ -23,19 +24,26 @@ from aap_eda.tasks.project import check_default_worker_health
 logger = logging.getLogger(__name__)
 
 
-def check_activation_worker_health() -> bool:
+def check_activation_worker_health(
+    queue_name: Optional[str] = None,
+) -> bool:
     """Check activation worker health (rulebook queues only).
+
+    Args:
+        queue_name: If provided, check only this specific queue.
+                    If None, check all configured queues and return
+                    True if any is healthy.
 
     Returns:
         bool: True if activation workers are healthy, False otherwise.
     """
     try:
-        # Check activation workers (sample one rulebook queue)
+        if queue_name is not None:
+            return check_rulebook_queue_health(queue_name)
+
         rulebook_queues = getattr(settings, "RULEBOOK_WORKER_QUEUES", [])
         if rulebook_queues:
-            # Check first configured rulebook queue as
-            # representative sample
-            return check_rulebook_queue_health(rulebook_queues[0])
+            return any(check_rulebook_queue_health(q) for q in rulebook_queues)
 
         # If no rulebook queues configured, activation workers are considered
         # healthy
@@ -48,12 +56,18 @@ def check_activation_worker_health() -> bool:
         return False
 
 
-def check_dispatcherd_workers_health(raise_exceptions=False) -> bool:
+def check_dispatcherd_workers_health(
+    raise_exceptions=False,
+    queue_name: Optional[str] = None,
+) -> bool:
     """Check dispatcherd worker health for both default and activation workers.
 
     Args:
         raise_exceptions: If True, raises WorkerUnavailable with
                          specific worker_type. If False, returns boolean.
+        queue_name: If provided, check only this specific activation queue.
+                    If None, check all configured queues (healthy if any
+                    is up).
 
     Returns:
         bool: True if both worker types are healthy, False otherwise.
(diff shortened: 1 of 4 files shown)


hsong-rh (dismissed):
Nice fix. This also has a good side effect: `create()` now checks `any()` queue instead of only `rulebook_queues[0]`, which means creating an activation no longer fails when an unrelated node is down.

A few non-blocking observations:

1. **`_get_activation_queue_name` has no direct tests** — it's simple enough that the risk is low, and a failure degrades safely to the old any-queue check. But since it's the core logic of this PR, a quick unit test covering the three branches (has queue, has instance but no queue, no instance) would be nice.

2. **No test for all-queues-unhealthy with `any()` fallback** — `test_check_dispatcherd_workers_health_any_queue_healthy` covers the positive case, but there's no multi-queue test where all return False. The existing single-queue unhealthy test covers the essential contract, so this is minor.

3. **Behavioral change in `create()` path is worth a release note** — the shift from first-queue-only to any-queue-healthy is an improvement, but it's not mentioned in the PR description. Might be worth calling out since it changes when 503s are returned.

jcraiglo1:
This looks great and happy to approve.

Question though, is there any reason to leave queue_name as an optional arg? What is the use of having the "check any queue" functionality if we are adding the ability to be more specific in our health checks?

ttuffin:
@jcraiglo1 it depends on the action being performed. For an action such as disable() or restart(), the activation is already associated with a queue, so the healthcheck must target that specific queue. However, for create(), we are creating a brand new activation which is not associated with any queue and therefore we do not care which queue to use, we just want the health check to ensure at least one queue is healthy and leave it up to the orchestrator to choose which one to use.

jcraiglo1 (dismissed):
Succinct and makes sense, great stuff. Thank you for the rundown!

ttuffin:
Thanks for the reviews  @hsong-rh @jcraiglo1. I've added an integration test to extend coverage of the code changes. Please review.

src/aap_eda/core/health.py
```diff
@@ -23,19 +24,26 @@
...
-        # Check activation workers (sample one rulebook queue)
+        if queue_name is not None:
+            return check_rulebook_queue_health(queue_name)
+
         rulebook_queues = getattr(settings, "RULEBOOK_WORKER_QUEUES", [])
         if rulebook_queues:
-            # Check first configured rulebook queue as
-            # representative sample
-            return check_rulebook_queue_health(rulebook_queues[0])
+            return any(check_rulebook_queue_health(q) for q in rulebook_queues)
```
wfealdel, line 46: @ttuffin 
Here if the first queue  check throws an error, remaining queues aren't checked. Could this cause some issue, or is this behavior acceptabe?

kaiokmo (approved):
LGTM. but I wanted to confirm that, apart from adding the unit tests, did you test this scenario:

- start a 2-node deployment
- create an activation, let it land on node-B
- kill node-A's workers
- attempt to delete/disable/restart the activation on node-B

before the fix, this would 503, right? with the fix now, it should succeed because node-B's queue is healthy. is that a fair assessment?

I don't remember how the deployment looks like on eda-qa, or if could try to replicate the above scenario there. nevertheless, I think we can try to confirm this manually

src/aap_eda/core/health.py
ttuffin: @wfealdel `any()` will iterate over each queue until it hits True which indicates a healthy queue has been found. If check_rulebook_queue_health() returns False it indicates the queue is unhealthy and the iteration continues.

ttuffin:
> LGTM. but I wanted to confirm that, apart from adding the unit tests, did you test this scenario:
> 
> * start a 2-node deployment
> * create an activation, let it land on node-B
> * kill node-A's workers
> * attempt to delete/disable/restart the activation on node-B
> 
> before the fix, this would 503, right? with the fix now, it should succeed because node-B's queue is healthy. is that a fair assessment?
> 
> I don't remember how the deployment looks like on eda-qa, or if could try to replicate the above scenario there. nevertheless, I think we can try to confirm this manually

@kaiokmo that is correct. Prior to this fix `check_activation_worker_health()` checks the first queue in the list, which in this example would be node-A. That check will fail and return false for the healthcheck, which causes the action to 503 even though the activation is already running on a perfectly healthy node.

jcraiglo1 (approved):

PR #1559, opened 2026-05-18
chore: Update github actions to use OIDC for codecov. Assisted-by: Claude

jcraiglo1:
EDA-Server is already connected to codecov. So here I changed around some configuration to use the recommended OIDC method of connecting. It is simpler to maintain as less configuration is required and github itself handles all provisioning. Alongside this we are removing unused env parameters. 

Note: As part of this change, flag analytics was enabled in codecov

## Summary by CodeRabbit

* **Chores**
  * Improved security of the CI/CD infrastructure by implementing OIDC-based authentication for code coverage reporting.

### DIFF
--- a/.github/actions/test/action.yml
+++ b/.github/actions/test/action.yml
@@ -77,7 +77,7 @@ runs:
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v5
       with:
-        env_vars: OS,PYTHON
+        use_oidc: true
         fail_ci_if_error: false
         files: ./coverage.xml
         flags: "unit-int-tests-${{ inputs.python-version }}"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -54,6 +54,9 @@ jobs:
 
   test:
     runs-on: ubuntu-latest
+    # Required for OIDC: https://github.com/codecov/codecov-action?tab=readme-ov-file#using-oidc
+    permissions:
+      id-token: write
     strategy:
       matrix:
         python-version:


AlexSCorey (approved):
This looks good to me.

.github/actions/test/action.yml
```diff
@@ -77,7 +77,7 @@ runs:
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v5
       with:
-        env_vars: OS,PYTHON
```
ptoscano, line 80: Hm is this parameter not needed anymore?

.github/workflows/ci.yaml
```diff
@@ -54,6 +54,9 @@ jobs:
 
   test:
     runs-on: ubuntu-latest
+    # Required for OIDC: https://github.com/codecov/codecov-action?tab=readme-ov-file#using-oidc
+    permissions:
+      id-token: write
```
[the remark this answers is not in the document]
ptoscano: It looks like `contents: read` makes it possible to use these endpoints:
https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents

Not sure whether any of them is needed right now (it seems not according to a quick skim of mine); feel free to add it if unsure.

ptoscano (changes requested):
Mostly LGTM, see my question about the removed parameter.

.github/actions/test/action.yml
jcraiglo1: It's not necessary because we have that information in the flags. We can keep it for potentially more granular sorting of tests in codecov. But again this is mostly handled already with the flags. With that being said it was removed because of a miscommunication between claude and I. But after more researching I don't think it's necessary.

ptoscano: More than that, it looks like `env_vars` is actually not working: https://github.com/codecov/codecov-action/issues/1828

So I'm fine to remove it, but please mention it explicitly in the commit/PR message, saying something about this change being an update to the codecov config removing a non-working parameter and switching to OIDC. Otherwise it is not clear what's the relation between OIDC and `env_vars`, or whether the removal of the latter was an accidental change.

jcraiglo1: PR description updated

ptoscano (approved):

PR #1577, opened 2026-06-01
[AAP-76179] fix(security): enforce sandboxed Jinja2 rendering in EDA server

B-Whitt:
## What

Replace unsandboxed `jinja2.nativetypes.NativeTemplate` with
`jinja2.sandbox.ImmutableSandboxedEnvironment` in all three rendering
call sites to prevent server-side template injection (SSTI):

- `strings.py`: sandbox `_render_string()`; remove Django settings object
  (`SECRET_KEY`, `DATABASES`, `REDIS_URL`) from `substitute_extra_vars()` context
- `credentials.py`: sandbox `_check_jinja_string()` in credential validation
- `ports.py`: sandbox `render_string()` in port resolution

## Why

The EDA server uses unsandboxed `NativeTemplate` which allows arbitrary
Python object traversal via template expressions. The `substitute_extra_vars()`
function also passes the entire Django settings object into the template
context, exposing secrets to any user who can craft activation extra_vars.

This matches the approach AWX already uses
(`jinja2.sandbox.ImmutableSandboxedEnvironment` at `credential/__init__.py:519`).

## Test plan

- [x] 9 regression tests verifying SSTI payloads are blocked across all three call sites
- [x] Existing port tests pass (10/10)
- [x] Existing credential validation tests pass (schema/injector tests; DB-dependent tests require PostgreSQL)
- [x] 96 additional unit tests pass with no regressions

Resolves: AAP-76179
Assisted-by: Claude Code / Opus 4.6 (Anthropic)

## Summary by CodeRabbit

* **Security**
  * Stronger template sandboxing to block template injection and remove application settings from template rendering context.
* **Tests**
  * New regression and coverage tests confirming malicious templates are blocked while legitimate templates render.
* **Bug Fixes**
  * Adjusted WebSocket payload formatting for a file-template scenario to match updated rendering behavior.

### DIFF
--- a/src/aap_eda/core/utils/credentials.py
+++ b/src/aap_eda/core/utils/credentials.py
@@ -31,13 +31,13 @@ from cryptography.x509.oid import NameOID
 from django.core.exceptions import ValidationError
 from django.forms import model_to_dict
 from django.utils.translation import gettext_lazy as _
-from jinja2.nativetypes import NativeTemplate
 from rest_framework import serializers
 
 from aap_eda.api.constants import EDA_SERVER_VAULT_LABEL
 from aap_eda.core import enums
 from aap_eda.core.utils.crypto.base import SecretValue
 from aap_eda.core.utils.external_sms import get_external_secrets
+from aap_eda.core.utils.strings import _SANDBOXED_ENV
 
 if typing.TYPE_CHECKING:
     from aap_eda.core import models
@@ -512,13 +512,13 @@ def _default_context(schema: dict, injectors: dict) -> dict:
 def _check_jinja_string(value: str, context: dict) -> str:
     try:
         if "{{" in value and "}}" in value:
-            result = NativeTemplate(
-                value, undefined=jinja2.StrictUndefined
-            ).render(context)
+            result = _SANDBOXED_ENV.from_string(value).render(context)
             if isinstance(result, jinja2.runtime.StrictUndefined):
                 raise InjectorMissingKeyException(f"{value} is undefined")
     except jinja2.exceptions.UndefinedError:
         raise InjectorMissingKeyException(f"{value} is undefined")
+    except jinja2.exceptions.SecurityError as e:
+        raise InjectorInvalidTemplateKey(str(e))
 
 
 def _validate_format(
--- a/src/aap_eda/core/utils/strings.py
+++ b/src/aap_eda/core/utils/strings.py
@@ -17,20 +17,30 @@ from typing import Any, Dict, List, Union
 
 import jinja2
 import yaml
-from django.conf import settings
-from jinja2.nativetypes import NativeTemplate
-
-from aap_eda.api.constants import EDA_SERVER_VAULT_LABEL
-from aap_eda.api.vault import encrypt_string
+from jinja2.exceptions import SecurityError
+from jinja2.nativetypes import NativeEnvironment
+from jinja2.sandbox import ImmutableSandboxedEnvironment
 
 LOGGER = logging.getLogger(__name__)
 
 
+class _NativeSandboxedEnvironment(
+    ImmutableSandboxedEnvironment, NativeEnvironment
+):
+    """Sandboxed Jinja2 environment that preserves native Python types."""
+
+    pass
+
+
+_SANDBOXED_ENV = _NativeSandboxedEnvironment(undefined=jinja2.StrictUndefined)
+
+
 def _render_string(value: str, context: dict) -> str:
     if "{{" in value and "}}" in value:
-        return NativeTemplate(value, undefined=jinja2.StrictUndefined).render(
-            context
-        )
+        try:
+            return _SANDBOXED_ENV.from_string(value).render(context)
+        except SecurityError:
+            raise ValueError(f"Template contains unsafe operations: {value}")
 
     return value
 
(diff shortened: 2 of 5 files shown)


src/aap_eda/core/utils/strings.py
```diff
@@ -17,20 +17,19 @@
...
 
+_SANDBOXED_ENV = ImmutableSandboxedEnvironment(undefined=jinja2.StrictUndefined)
+
 
 def _render_string(value: str, context: dict) -> str:
     if "{{" in value and "}}" in value:
-        return NativeTemplate(value, undefined=jinja2.StrictUndefined).render(
-            context
-        )
+        return _SANDBOXED_ENV.from_string(value).render(context)
```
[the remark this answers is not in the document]
B-Whitt: addressed

src/aap_eda/core/utils/credentials.py
```diff
@@ -509,16 +509,21 @@ def _default_context(schema: dict, injectors: dict) -> dict:
...
 
+_SANDBOXED_ENV = ImmutableSandboxedEnvironment(
+    undefined=jinja2.StrictUndefined,
+)
+
+
 def _check_jinja_string(value: str, context: dict) -> str:
     try:
         if "{{" in value and "}}" in value:
-            result = NativeTemplate(
```
mkanoor, line 515: @B-Whitt A native template will preserve the data types so an integer will remain an integer, with this ImmutableSandboxedEnvironment will we loose the Native types would this be better

```
from jinja2.sandbox import ImmutableSandboxedEnvironment
from jinja2.nativetypes import NativeEnvironment, NativeTemplate

class ImmutableSandboxedNativeEnvironment(ImmutableSandboxedEnvironment, NativeEnvironment):
    environment_class = NativeTemplate
```

B-Whitt: @mkanoor addressed

B-Whitt:
> /run-e2e

- 1 unrelated e2e test failure tracked here: https://redhat.atlassian.net/browse/AAP-77489
- All other tests pass

src/aap_eda/core/utils/strings.py
wfealdel: @B-Whitt 
It seems that the function `substitute_extra_vars()` which exposes Django settings in the template context is **dead code** - it exists in the codebase but is never called. I'm not completely familiar this code yet so not 100% sure, but probably we can delete it completely.  This code was also flagged in `https://redhat.atlassian.net/browse/AAP-76106`

B-Whitt: addressed. thanks

jcraiglo1 (dismissed):

src/aap_eda/core/utils/strings.py
```diff
@@ -17,20 +17,21 @@
...
-from django.conf import settings
-from jinja2.nativetypes import NativeTemplate
+from jinja2.sandbox import ImmutableSandboxedEnvironment
 
 from aap_eda.api.constants import EDA_SERVER_VAULT_LABEL
 from aap_eda.api.vault import encrypt_string
 
 LOGGER = logging.getLogger(__name__)
 
+_SANDBOXED_ENV = ImmutableSandboxedEnvironment(
```
wfealdel: @B-Whitt  I'm sharing some findings from the AI analysis 

✅ Strengths

  - Addresses the core vulnerability: ImmutableSandboxedEnvironment successfully blocks the primary SSTI attack vectors (__class__, __mro__, __subclasses__, etc.)
  - Matches AWX's proven approach: Uses the same fundamental security model that has worked in production
  - Simplicity: Minimal code changes, easy to review, low risk of introducing bugs
  - Leverages Jinja2's built-in security: Uses well-tested library functionality rather than custom security logic

  ⚠️  Gaps to Consider

  1. No explicit filter management: Potentially dangerous filters (attr, map, select, safe) remain available. Defense-in-depth would suggest removing filters not needed for legitimate use cases.
  2. No Django model-specific protections: Internal Django model attributes (_meta, _state, objects) could still be accessed, potentially exposing model structure.
  3. No template compilation caching: Each render compiles the template fresh, which could impact performance for frequently-used templates.
  4. Generic error messages: Default Jinja2 security errors may not provide enough context for debugging or understanding what went wrong.

Summary: This fix will work and blocks the critical SSTI vulnerability. The gaps listed above are defense-in-depth enhancements rather than blocking issues.

---------------------------------------------------

We could move this code to a separated file where we can implement those suggested filters and template caching

AlexSCorey (dismissed):
Looks good to me.  It addresses the most pressing vulnerabilities.

src/aap_eda/core/utils/strings.py
B-Whitt: Gap 1 (dangerous filters) — Not an issue. The sandbox's is_safe_attribute intercepts unsafe attribute access even through the attr filter. {{ ''|attr('__class__')|attr('__mro__')|attr('__subclasses__')() }} raises UndefinedError. The safe filter only affects HTML autoescaping, which isn't enabled here. map/select are standard iteration filters with no security implications.

- Verification: attr filter bypass confirmed blocked via is_safe_attribute assertion; full SSTI chain via attr confirmed raises UndefinedError.

Gap 2 (Django model attributes) — Not an issue. The sandbox blocks all underscore-prefixed attributes by default. is_safe_attribute returns False for _meta, _state, __class__, etc.

- Verification: underscore attributes _meta, _state confirmed return False from is_safe_attribute.

Gap 3 (template compilation caching) — Valid but negligible. Benchmarked at ~0.19ms overhead per from_string() call. For activation startup (1-5 renders), that's <1ms total. Not worth the added complexity.

- Verification: timeit benchmark — 10k calls with vs without caching, 0.19ms delta per call.

Gap 4 (generic error messages) — Valid but already addressed. Each call site wraps SecurityError into domain-specific exceptions (ValueError, InjectorInvalidTemplateKey, ActivationStartError). The raw Jinja2 messages already name the blocked attribute and object type (e.g. "access to attribute '__class__' of 'str' object is unsafe.").

- Verification: 15 SSTI prevention unit tests passing (tests/unit/test_ssti_prevention.py), 10 port integration tests passing (tests/integration/services/activation/engine/test_ports.py), 497 total unit tests passing, linter clean.

wfealdel (approved):

mkanoor (approved):

PR #1578, opened 2026-06-03
Force re-computation with event stream bulk create

AlanCoding:
Hit a failure in

https://github.com/ansible/django-ansible-base/pull/970

And I can explain what's happening, and why it's basically passing accidentally now.

DAB RBAC is a signals-based library. If you `bulk_create` objects, organization permissions will not be properly rebuilt, this is expected. The fixture involved does exactly this, and the test expects that lists show event streams correctly.

This happens to work now, because later on in the process, a credential is created. The signals that fire due to the credential creation happen to re-build role evaluations related to the previously-created event stream objects. This is because the signals are overly-broad. The DAB PR 970 narrows the scope of what those signals touch, which is a good thing, but incompatible with how this test was functioning.

Confirmed locally, after applying this change, the failing tests pass even with the DAB patch.

## Summary by CodeRabbit

* **Tests**
  * Updated integration test fixtures to refresh RBAC permissions after creating test data and to return the created event stream instances. This improves consistency of permission state during tests and reduces intermittent failures.


tests/integration/conftest.py
```diff
@@ -1156,7 +1156,11 @@ def default_event_streams(
     default_user: models.User,
     default_hmac_credential: models.EdaCredential,
 ) -> List[models.EventStream]:
-    return models.EventStream.objects.bulk_create(
+    from ansible_base.rbac.caching import compute_object_role_permissions
```
wfealdel: Minor: These imports are inline while the rest of the file imports DAB-dependent modules at the top level. Consider moving them to the module-level imports for consistency, unless there's a circular import reason to keep them here.

tests/integration/conftest.py
```diff
@@ -1179,6 +1183,17 @@ def default_event_streams(
         ]
     )
 
+    # bulk_create bypasses signals, trigger RBAC permission cache update
+    # for the organization roles that should grant access to these EventStreams
+    org_ct = ContentType.objects.get_for_model(default_organization)
```
wfealdel: Nit: The comment explains *what* the code does but not *why* only organization roles are targeted. A reader unfamiliar with DAB's `parent_field_name="organization"` inheritance model might wonder why per-object roles aren't recomputed. Consider expanding slightly

tests/integration/conftest.py
```diff
@@ -1179,6 +1183,17 @@ def default_event_streams(
...
+    org_roles = ObjectRole.objects.filter(
```
wfealdel: Minor: Other fixtures in this file also use `bulk_create` on RBAC-registered models (e.g., `default_activation_instances` at line 684). If those tests also pass accidentally due to overly-broad signals, they may break when DAB PR #970 lands. Worth auditing as a follow-up.

wfealdel (commented):
Overall looks correct — the fix properly targets org-level role recomputation after bulk_create bypasses signals. Minor suggestions on import placement and comment clarity, plus a note about other bulk_create fixtures that may need the same treatment.

tests/integration/conftest.py
AlanCoding: Solved another way, changing the method used.

I just changed the call to do the evaluation re-computation globally. Since this is a pytest test, it runs in a transaction and the database generally only has resources used in the test, so it isn't a heavy burden for the test runtime. And it avoids reaching into any of the specifics like object-role or the org vs. object roles. So the only thing we need to say is "recomputes role evaluations" and that's a single call.

tests/integration/conftest.py
AlanCoding: with the new pattern, applied it to that fixture as well

tests/integration/conftest.py
AlanCoding: done

wfealdel (approved):
LGTM

jcraiglo1 (approved):

mkanoor (approved):

PR #1584, opened 2026-06-11
fix: restore prior activation create and enable behaviour

ttuffin:
When we switched from redis to dispatcherd we added worker healthchecks to the activation create() and enable() endpoints. The result is that when all activation workers are offline, these actions will fail with a HTTP 503, which is a change in behaviour compared to previous versions running with redis. This change removes the healthchecks from these two actions in order to restore the previous behaviour.

Assisted-by: Claude

https://redhat.atlassian.net/browse/AAP-78450

## Summary by CodeRabbit

* **Refactor**
  * Removed pre-flight worker availability checks from activation creation and enable workflows, so activations proceed directly to project sync and start steps when enabled.

### DIFF
--- a/src/aap_eda/api/views/activation.py
+++ b/src/aap_eda/api/views/activation.py
@@ -99,13 +99,6 @@ class ActivationViewSet(
         serializer = self.get_serializer(data=request.data)
         serializer.is_valid(raise_exception=True)
 
-        # Check if activation workers are available for enabled activations
-        is_enabled = serializer.validated_data.get(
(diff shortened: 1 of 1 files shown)


AlanCoding (approved):

AlanCoding:
Discussed. Understood that user experience may be that things to into pending if background service is not running. This is the same as controller.

This will better position us to move towards standardized platform solutions, because this did have a bit of an odd-one-out solution.

mkanoor (approved):

PR #1593, opened 2026-06-17
Fixes a test that will fail after the rbac cursor work is merged.

AlexSCorey:
This test will begin to fail once [this pr](https://github.com/ansible/django-ansible-base/pull/1025) on the dab repo is merged. So this pr is meant to head that off.  The DAB pr probably should merge first, and then this 1 should merge.
AAP-68017

## Summary by CodeRabbit

* **Tests**
  * Updated test assertion to validate paginated API response by verifying returned results.

### DIFF
--- a/tests/integration/dab_rbac/test_role_permissions.py
+++ b/tests/integration/dab_rbac/test_role_permissions.py
@@ -41,7 +41,7 @@ def test_view_assignments_non_admin(
     url = reverse("roleuserassignment-list")
     r = user_client.get(url)
     assert r.status_code == 200
-    assert r.data["count"] == 1
+    assert len(r.data["results"]) == 1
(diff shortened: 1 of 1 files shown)


ptoscano:
What is the explanation for the behaviour change?

AlexSCorey:
@ptoscano > What is the explanation for the behaviour change?

The change in https://github.com/ansible/django-ansible-base/pull/1025 is that for RBAC queries we will use cursor to help lighten the load of db queries.  Part of that change means that we have removed "count" from the http response as a result this test is failing on the pr because it relies on count being in the response.  So this change will go in right before/after 1025 merges to keep CI clean.

jcraiglo1 (approved):

B-Whitt (approved):

PR #1595, opened 2026-06-18
fix: prevent race condition in project partial_update (AAP-79566)

mkanoor:
## Summary
- Fix race condition in `partial_update` where the API handler could overwrite `import_state` back to `PENDING` after a worker already completed the sync
- Set `import_state=PENDING` under a `select_for_update` row lock **before** dispatching the task, so the worker is never blocked and can freely update `import_state`
- Record `import_task_id` via a targeted `.update()` that only touches that field — the worker never writes `import_task_id`, so no lock needed
- Reduce DB writes from 3 to 2 by merging the user-field save and the `import_state` save into a single `project.save(update_fields=...)`
- Extract `_needs_project_sync()` and `_mark_pending_if_idle()` helpers for clarity

## Root Cause
In `partial_update`, when SCM fields (url, branch, refspec) change:
1. `tasks.sync_project()` dispatched the task to the queue
2. A worker picked it up and completed: PENDING → RUNNING → COMPLETED
3. `project.save()` (no `update_fields`) overwrote **all** fields from the stale in-memory object, resetting `import_state` back to PENDING
4. Project stuck in PENDING permanently — no worker will re-process it

The `sync` action endpoint didn't have this bug because it already used `select_for_update`.

## Test plan
- [ ] `test_partial_update_does_not_overwrite_worker_state` — simulates worker completing before job_id is recorded; verifies COMPLETED state preserved
- [ ] `test_partial_update_does_not_overwrite_unrelated_fields` — verifies only `import_state` and `import_task_id` written, not stale fields like `import_error`
- [ ] `test_partial_update_skips_sync_when_already_pending` — verifies no duplicate dispatch when already PENDING
- [ ] Existing `test_partial_update_project_*` tests pass unchanged

## Release Notes

* **Bug Fixes**
  * Fixed race conditions when updating SCM-related project fields
  * Improved project synchronization reliability during concurrent updates
  * Enhanced state consistency to prevent missed syncs and data loss

### DIFF
--- a/src/aap_eda/api/views/project.py
+++ b/src/aap_eda/api/views/project.py
@@ -213,6 +213,33 @@ class ProjectViewSet(
             ),
         },
     )
+    @staticmethod
+    def _needs_project_sync(update_fields):
+        return bool({"scm_branch", "scm_refspec", "url"} & set(update_fields))
+
+    @staticmethod
+    def _mark_pending_if_idle(project, update_fields):
+        """Read fresh import_state under row lock; set PENDING if idle.
+
+        Appends 'import_state' to update_fields when the state is
+        changed so the caller's single save persists it.
+        Returns True when a sync task should be dispatched.
+        """
+        current_state = (
+            models.Project.objects.select_for_update()
+            .values_list("import_state", flat=True)
+            .get(pk=project.id)
+        )
+        if current_state in [
+            models.Project.ImportState.PENDING,
+            models.Project.ImportState.RUNNING,
+        ]:
+            return False
+
+        project.import_state = models.Project.ImportState.PENDING
+        update_fields.append("import_state")
+        return True
+
     def partial_update(self, request, pk):
         project = self.get_object()
         serializer = serializers.ProjectUpdateRequestSerializer(
@@ -227,7 +254,17 @@ class ProjectViewSet(
             setattr(project, key, value)
             update_fields.append(key)
 
+        needs_sync = self._needs_project_sync(update_fields)
+
+        if needs_sync and not check_default_worker_health():
+            raise api_exc.WorkerUnavailable(worker_type="default")
+
+        sync_needed = False
         with transaction.atomic():
+            if needs_sync:
+                sync_needed = self._mark_pending_if_idle(
+                    project, update_fields
+                )
             project.save(update_fields=update_fields)
             check_related_permissions(
                 request.user,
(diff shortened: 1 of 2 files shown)


hsong-rh:
@mkanoor is this scenario valid?
```
  Timeline:
  1. User PATCHes project with new URL
  2. transaction.atomic() runs:
     - select_for_update → state is COMPLETED
     - Sets import_state = PENDING
     - project.save() commits PENDING + new URL  ← COMMITTED
  3. Transaction ends, lock released
  4. tasks.sync_project(project.id)  ← RAISES EXCEPTION
     (dispatcherd down, pg_notify fails, queue error)
  5. DRF returns 500

  Result:
  - Project stuck in PENDING, no task running
  - User retries PATCH → sees PENDING → _mark_pending_if_idle returns False → no sync
  - Stuck until monitor_project_tasks recovers (DISPATCHERD_PROJECT_TASK_TIMEOUT * 2 seconds)
```

mkanoor:
> @mkanoor is this scenario valid?
> 
> ```
>   Timeline:
>   1. User PATCHes project with new URL
>   2. transaction.atomic() runs:
>      - select_for_update → state is COMPLETED
>      - Sets import_state = PENDING
>      - project.save() commits PENDING + new URL  ← COMMITTED
>   3. Transaction ends, lock released
>   4. tasks.sync_project(project.id)  ← RAISES EXCEPTION
>      (dispatcherd down, pg_notify fails, queue error)
>   5. DRF returns 500
> 
>   Result:
>   - Project stuck in PENDING, no task running
>   - User retries PATCH → sees PENDING → _mark_pending_if_idle returns False → no sync
>   - Stuck until monitor_project_tasks recovers (DISPATCHERD_PROJECT_TASK_TIMEOUT * 2 seconds)
> ```

@hsong-rh Its valid but its not pro-active. We will depend on the monitor to mark it as stuck, but if we can check in the Task Queue and get the status of the job from import_task_id and it says it has failed we should restart the sync when the user applies an update. We implemented something like this for Activation in 2.5/2.6 with RQ now that we have switched to dispatcherd I dont know if it can tell us the status of the old job. As soon as the job starts it sets the import_state to running and if it fails in doing that DB transaction it will stay in Pending state.

hsong-rh (approved):

wfealdel (approved):
LGTM

PR #1596, opened 2026-06-18
fix: stop circuit breaker bypass via failure_count reset (AAP-78375)

hsong-rh:
https://redhat.atlassian.net/browse/AAP-78375

## Summary

- **Root cause:** `_detect_running_status()` unconditionally reset `failure_count` to 0 on every STARTING→RUNNING transition, preventing `ACTIVATION_MAX_RESTARTS_ON_FAILURE` from ever tripping for activations that crash after reaching RUNNING status
- **Fix:** Remove the `_reset_failure_count()` call from `_detect_running_status()` and the now-dead method (7 lines removed)
- **Test:** Added `test_monitor_preserves_failure_count_on_running_transition` to `test_manager.py`

## Details

When an activation instance survived long enough to receive its first heartbeat (transitioning from STARTING to RUNNING), the `failure_count` was reset to 0. If the activation then crashed (e.g., after 15 minutes due to cascading job failures or liveness probe timeouts), the counter started over from scratch. The circuit breaker threshold was never reached, causing infinite restart loops and cluster-wide resource exhaustion.

Customer impact: In Case 04460263, an activation reached **205 failed instances over two days** with `ACTIVATION_MAX_RESTARTS_ON_FAILURE: 5` active but ignored.

`failure_count` is still properly reset by user-initiated actions:
- User enables a disabled activation (`api/views/activation.py`)
- Project sync recovers a failed activation (`tasks/project.py`)

## Verification

Reproduced and verified on local podman pods:

| Scenario | Before fix | After fix |
|---|---|---|
| Set `failure_count=4`, watch restart cycle | Reset to **0** on STARTING→RUNNING | Preserved at **4** through multiple cycles |

## Test plan

- [x] New test: `test_monitor_preserves_failure_count_on_running_transition` verifies `failure_count` is preserved during STARTING→RUNNING transition
- [x] Existing `test_monitor_to_running_status` still passes (only checks `restart_count`, unaffected)
- [x] Project sync tests (`test_projects.py`) still pass — those test the legitimate reset paths
- [ ] CI: full test suite

Fixes: [AAP-78375](https://issues.redhat.com/browse/AAP-78375)

## Release Notes

* **Bug Fixes**
  * Fixed an issue where an activation moving from **STARTING** to **RUNNING** could incorrectly reset its stored **failure count**.

* **Tests**
  * Added an integration test to confirm that `monitor()` preserves **failure_count** during the **STARTING → RUNNING** transition.

### DIFF
--- a/tests/integration/services/activation/test_manager.py
+++ b/tests/integration/services/activation/test_manager.py
@@ -383,6 +383,30 @@ def test_monitor_to_running_status(
 
 
 @pytest.mark.django_db
+def test_monitor_preserves_failure_count_on_running_transition(
+    starting_activation: models.Activation,
+    container_engine_mock: MagicMock,
+    running_container_status_mock: MagicMock,
+):
+    """AAP-78375: failure_count must not reset on STARTING->RUNNING."""
+    starting_activation.failure_count = 3
+    starting_activation.save(update_fields=["failure_count"])
+
+    activation_manager = ActivationManager(
+        db_instance=starting_activation,
+        container_engine=container_engine_mock,
+    )
+    container_engine_mock.get_status.return_value = (
+        running_container_status_mock
+    )
+    activation_manager.monitor()
+
+    starting_activation.refresh_from_db()
+    assert starting_activation.status == enums.ActivationStatus.RUNNING
+    assert starting_activation.failure_count == 3
+
+
+@pytest.mark.django_db
 def test_monitor_container_engine_error_sets_workers_offline(
     running_activation: models.Activation,
     container_engine_mock: MagicMock,
--- a/src/aap_eda/services/activation/activation_manager.py
+++ b/src/aap_eda/services/activation/activation_manager.py
@@ -97,12 +97,6 @@ class ActivationManager(StatusManager):
         self.db_instance.save(update_fields=["failure_count", "modified_at"])
 
     @run_with_lock
-    def _reset_failure_count(self):
-        """Reset the failure count of the activation."""
-        self.db_instance.failure_count = 0
-        self.db_instance.save(update_fields=["failure_count", "modified_at"])
-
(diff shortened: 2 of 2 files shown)


tests/integration/services/activation/test_manager.py
```diff
@@ -381,6 +381,30 @@ def test_monitor_to_running_status(
...
 
 
+@pytest.mark.django_db
+def test_monitor_preserves_failure_count_on_running_transition(
+    starting_activation: models.Activation,
+    container_engine_mock: MagicMock,
+    running_container_status_mock: MagicMock,
+):
+    """AAP-78375: failure_count must not reset on STARTING->RUNNING."""
+    starting_activation.failure_count = 3
```
AlexSCorey, line 392: I think I understand the reason for the change.  We don't want to restart the failure count if the activation starts --> runs --> 3 seconds later it falls over.  What about cases where it runs for hours/days?  Shouldn't we restart the count at that point?

hsong-rh: @AlexSCorey A failure is a failure regardless of its uptime. If an activation fails 5 times over weeks, that's still a pattern worth stopping. The admin can always disable/enable to reset the counter if they know a failure was transient. What do you think?

AlexSCorey: I'm ok with that, and I think we should update any docs/help text to let users know that failure count only reset on disable --> enable.  That said, this request doesn't block this work from merging IMO.  We can do that in a separate ticket.

AlexSCorey (dismissed):
All concerns addressed.  LGTM

AlexSCorey (approved):
LGTM

jcraiglo1 (approved):
Great stuff

PR #1598, opened 2026-06-18
fix: disable service account token automount for activation job pods

TheRealHaoLiu:
## Summary

- Activation job pods do not need Kubernetes API access — they run user-supplied rulebooks and decision environment images. This sets automount_service_account_token to False on the pod spec so the default service account token is not projected into these pods.

## Test plan

- [ ] Verify existing integration tests pass
- [ ] Deploy and confirm activation job pods no longer have /var/run/secrets/kubernetes.io/serviceaccount/token mounted

## Summary by CodeRabbit

* **Bug Fixes**
  * Kubernetes pods created for jobs now explicitly disable automatic service account token mounting in pod templates, improving security and reducing unnecessary token exposure.

* **Tests**
  * Updated integration tests to verify the service account token auto-mounting is disabled in pod templates across job configuration scenarios.

### DIFF
--- a/src/aap_eda/services/activation/engine/kubernetes.py
+++ b/src/aap_eda/services/activation/engine/kubernetes.py
@@ -346,6 +346,7 @@ class Engine(ContainerEngine):
         spec_kwargs: dict = {
             "restart_policy": "Never",
             "containers": [container],
+            "automount_service_account_token": False,
         }
         if request.credential:
             self._create_secret(request, log_handler)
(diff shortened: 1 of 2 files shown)


src/aap_eda/services/activation/engine/kubernetes.py
```diff
@@ -346,6 +346,7 @@ def _create_pod_template_spec(
         spec_kwargs: dict = {
             "restart_policy": "Never",
             "containers": [container],
+            "automount_service_account_token": False,
```
mkanoor, line 349: @TheRealHaoLiu Will this break if the pod needs to monitor for Kubernetes Events via the API we have a source plugin for Kubernetes https://github.com/sabre1041/sabre1041.eda/blob/a9c2508ce03ac292cdfe524aaeb431552013de4f/plugins/event_source/k8s.py#L160

mkanoor: There was some discussion about this in the past that the token has to be passed via Credentials but I am not sure if that was implemented. @kaiokmo

mkanoor (approved):

PR #1600, opened 2026-06-18
fix: sonar issues - Reused strings set to constants

jcraiglo1:
https://redhat.atlassian.net/browse/AAP-77394

One of multiple PRs addressing sonar quality gate issues. Sets several reused strings to constants.

Attached is a document describing the actions taken here.

[sonarcloud-remediation-report.md](https://github.com/user-attachments/files/29103602/sonarcloud-remediation-report.md)

Assisted-by: Claude

## Summary by CodeRabbit

* **Documentation**
  * Standardized API field help-text for fired-rule details and user username to improve consistency in generated API docs.
* **Bug Fixes**
  * Made validation messaging consistent when a required “Decision Environment” is missing during activation create/update and reactivation.
* **Chores**
  * Improved and centralized PEM private-key type handling using shared constants for clearer, more maintainable logic.

PR #1601, opened 2026-06-18
fix: Reduce duplicate code percentage from sonar by using class inhertiance within rulebook serializers

jcraiglo1:

One of multiple PRs addressing sonar quality gate issues. This one pertains to duplicated code in our rulebook serializer. This is addressed by creating a base class and applying to the serializers in the module.

This PR is blocking another PR currently https://github.com/ansible/eda-server/pull/1600

Assisted-by: Claude

## Summary by CodeRabbit

* **Refactor**
  * Consolidated audit rule serializers by introducing a shared base class, eliminating code duplication and improving maintainability of the serialization architecture.

### DIFF
--- a/src/aap_eda/api/serializers/rulebook.py
+++ b/src/aap_eda/api/serializers/rulebook.py
@@ -104,7 +104,7 @@ class AuditRuleSerializer(serializers.ModelSerializer):
         read_only_fields = ["id", "organization_id", "created_at"]
 
 
-class AuditRuleDetailSerializer(serializers.Serializer):
+class AuditRuleSerializerBase(serializers.Serializer):
     id = serializers.IntegerField(
         required=True,
         help_text="ID of the fired rule",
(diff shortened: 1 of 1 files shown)


ptoscano:
In practice the difference between `AuditRuleDetailSerializer` and `AuditRuleListSerializer` is that `AuditRuleDetailSerializer` has also two fields, `ruleset_name` and `created_at`. Rather than moving some parts of them in a mixin, I'd rather rename the whole `AuditRuleListSerializer` as e.g. `AuditRuleSerializerBase`, and make `AuditRuleDetailSerializer` & `AuditRuleListSerializer` inherit from `AuditRuleSerializerBase` (`AuditRuleListSerializer` would not add anything to it).

Small nit: it seems that `mixin.py` is used for mixins shared among different modules of serializers, whereas this is only specific to `rulebook.py`, so I'd keep it in this module.

jcraiglo1:
@ptoscano I agree and have instead created a base class and removed the addition to mixins.py. Please review

ptoscano (approved):
Oooh, this refactors sparks joy.

wfealdel (approved):

PR #1602, opened 2026-06-22
feat: add activation instance scoping to JWT tokens for WebSocket aut…

mkanoor:
WebSocket Authentication

This commit adds security enhancements to ensure JWT tokens are scoped to specific activation instances, preventing token reuse across different instances. It also implements secure token refresh with scope preservation and backward- compatible legacy token support with actionable warnings.

https://redhat.atlassian.net/browse/AAP-83971

Key changes:
- Add activation_instance_id to JWT token payload (supports int and UUID)
- Validate token scope in WebSocket consumer before processing messages
- Implement secure token refresh that preserves activation_instance_id scope
- Support backward compatibility with legacy tokens (no forced logout)
- Add actionable warnings for legacy tokens with activation name
- Implement token payload caching for performance

Security improvements:
- Tokens cannot be used for different activation instances
- Token refresh preserves scope (prevents scope escalation vulnerability)
- Two-layer validation: middleware (signature/expiry) + consumer (scope)
- Automatic revocation via database existence checks
- Comprehensive error logging for audit trails

Token Refresh Security:
- RefreshTokenSerializer extracts activation_instance_id from refresh token
- New access tokens preserve the scope from refresh token (trusted source)
- Prevents privilege escalation by not accepting client-provided scope
- Legacy refresh tokens continue to work (no scope added)

Backward Compatibility:
- Legacy tokens without activation_instance_id are accepted with warnings
- Warnings logged once per connection to avoid log spam
- Warnings include activation name for easy operator action
- No event loss, no forced disconnection

Files modified:
- src/aap_eda/services/auth.py: Token creation and validation
- src/aap_eda/services/activation/engine/common.py: Pass activation_instance_id
- src/aap_eda/wsapi/consumers.py: Token scope validation with async warnings
- src/aap_eda/api/serializers/auth.py: Extract scope from refresh tokens
- src/aap_eda/api/views/auth.py: Preserve scope in token refresh endpoint

Tests added:
- 9 tests for auth service (token creation, validation, scope checking)
- 14 tests for consumer validation (scope matching, legacy token warnings)
- 3 tests for token refresh endpoint (scope preservation, legacy handling)
- 7 tests for RefreshTokenSerializer (scope extraction, error handling)
- Total: 33 passing tests

All code formatted with black, sorted with isort, and linted with flake8.

## Release Notes

* **New Features**
  * JWT access/refresh tokens now optionally include an activation-instance scope claim.
  * Refreshing a token preserves any existing activation-instance scope when issuing the new access token.
  * WebSocket requests validate that the token scope matches the message’s activation context; mismatches now fail authentication (connection closes with auth error). Legacy tokens without scope are allowed with a one-time security warning.

* **Tests**
  * Added/expanded integration coverage for scoped vs legacy refresh behavior (including UUID/non-UUID forms) and WebSocket authorization/scope enforcement (including malformed headers).

### DIFF
--- a/src/aap_eda/api/serializers/auth.py
+++ b/src/aap_eda/api/serializers/auth.py
@@ -30,7 +30,12 @@ class RefreshTokenSerializer(serializers.Serializer):
 
     def validate(self, data):
         try:
-            self.user = validate_jwt_token(data["refresh"], "refresh")
+            self.user, token_payload = validate_jwt_token(
+                data["refresh"], "refresh"
+            )
+            self.activation_instance_id = token_payload.get(
+                "activation_instance_id"
+            )
         except InvalidTokenError as e:
             raise serializers.ValidationError("Invalid token") from e
         return data
--- a/src/aap_eda/services/auth.py
+++ b/src/aap_eda/services/auth.py
@@ -12,6 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
+import logging
 from datetime import datetime, timedelta
 from itertools import groupby
 
@@ -21,6 +22,8 @@ from django.conf import settings
 
 from aap_eda.core.models.user import User
 from aap_eda.services.exceptions import InvalidTokenError
+
+logger = logging.getLogger(__name__)
 
 
 def group_permission_resource(permission_data):
@@ -42,10 +45,18 @@ def display_permissions(role_data: dict) -> dict:
 JWT_ALGORITHM = "HS256"
 
 
-def create_jwt_token() -> tuple[str, str]:
+def create_jwt_token(
+    activation_instance_id: str | int | None = None,
+) -> tuple[str, str]:
     """Create JWT access and refresh token pair.
 
     They can be sent to rulebook clients through command line arguments.
+
+    Args:
+        activation_instance_id: Optional activation instance ID to
+            scope the token. Can be string or int (will be converted
+            to string). If provided, the token can only be used for
+            this activation instance.
     """
     with no_reverse_sync():
         user, new = User.objects.get_or_create(
@@ -56,31 +67,54 @@ def create_jwt_token() -> tuple[str, str]:
         user.set_unusable_password()
         user.save(update_fields=["password"])
 
-    access_token = jwt_access_token(user.id)
-    refresh_token = jwt_refresh_token(user.id)
+    # Convert to string for consistent comparison
+    activation_id_str = (
+        str(activation_instance_id)
+        if activation_instance_id is not None
+        else None
+    )
+    access_token = jwt_access_token(user.id, activation_id_str)
+    refresh_token = jwt_refresh_token(user.id, activation_id_str)
     return (access_token, refresh_token)
 
 
-def jwt_access_token(user_id: int) -> str:
+def jwt_access_token(
+    user_id: int, activation_instance_id: str | None = None
+) -> str:
     expiration = timedelta(
         minutes=float(settings.JWT_ACCESS_TOKEN_LIFETIME_MINUTES)
     )
-    return get_jwt_token(user_id, expiration, token_type="access")
+    return get_jwt_token(
+        user_id,
+        expiration,
+        token_type="access",
+        activation_instance_id=activation_instance_id,
+    )
 
 
-def jwt_refresh_token(user_id: int) -> str:
+def jwt_refresh_token(
+    user_id: int, activation_instance_id: str | None = None
+) -> str:
     expiration = timedelta(
         days=float(settings.JWT_REFRESH_TOKEN_LIFETIME_DAYS)
     )
-    return get_jwt_token(user_id, expiration, token_type="refresh")
+    return get_jwt_token(
+        user_id,
+        expiration,
+        token_type="refresh",
+        activation_instance_id=activation_instance_id,
+    )
 
 
 def get_jwt_token(user_id, expiration, **kwargs):
+    activation_instance_id = kwargs.pop("activation_instance_id", None)
     payload = {
         "user_id": user_id,
         "exp": datetime.now() + expiration,
         **kwargs,
     }
+    if activation_instance_id is not None:
+        payload["activation_instance_id"] = activation_instance_id
 
     return jwt.encode(payload, settings.SECRET_KEY, JWT_ALGORITHM)
 
@@ -94,7 +128,26 @@ def parse_jwt_token(token: str) -> dict:
         raise InvalidTokenError("Expired token") from e
 
 
-def validate_jwt_token(token: str, token_type: str) -> User:
+def validate_jwt_token(
+    token: str,
+    token_type: str,
+    activation_instance_id: str | int | None = None,
+) -> tuple[User, dict]:
+    """Validate JWT token and optionally verify activation instance scope.
+
+    Args:
+        token: The JWT token to validate
+        token_type: Expected token type ('access' or 'refresh')
+        activation_instance_id: If provided, verify that the token
+            is scoped to this activation instance. Can be string or
+            int.
+
+    Returns:
+        A tuple of (User, token_payload)
+
+    Raises:
+        InvalidTokenError: If token is invalid or scope doesn't match
+    """
     token_dict = parse_jwt_token(token)
 
     for key in ["user_id", "exp", "token_type"]:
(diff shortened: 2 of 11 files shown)


mkanoor:
@coderabbitai review

src/aap_eda/wsapi/consumers.py
```diff
@@ -100,6 +102,204 @@ class Event(Enum):
...
+
+        # Security: Reject refresh tokens - only access tokens allowed
+        token_type = self._token_payload.get("token_type")
+        if token_type != "access":
+            raise InvalidTokenError(
+                f"Invalid token type '{token_type}' for WebSocket auth. "
+                "Only access tokens are permitted."
+            )
+
+        return self._token_payload
```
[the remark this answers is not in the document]
mkanoor: Will update

ttuffin:
@mkanoor is there a JIRA to track this work?

mkanoor:
@ttuffin This is an additional check on https://redhat.atlassian.net/browse/AAP-79532

ttuffin (dismissed):

src/aap_eda/wsapi/consumers.py
```diff
@@ -100,6 +102,247 @@ class Event(Enum):
...
+            ObjectDoesNotExist: If referenced entities not found
+        """
+        activation_id = data.get("activation_id")
+
+        # Security: For JOB messages, validate against the ID used for
+        # persistence (ansible_rulebook_id) to prevent scope bypass
+        if msg_type == MessageType.JOB:
+            ansible_rulebook_id = data.get("ansible_rulebook_id")
+            # Validate the ID that will be persisted
+            if ansible_rulebook_id is not None:
```
wfealdel: **[Major — Functionality]** JOB message scope bypass when `ansible_rulebook_id` is absent.

When a JOB message has `ansible_rulebook_id=None` but `activation_id` is present, scope validation is skipped entirely. The `if ansible_rulebook_id is not None` guard is `False` so `_validate_token_scope` never runs. The generic fallback at line 300 (`elif activation_id is not None`) is unreachable because the `if msg_type == MessageType.JOB` branch already matched.

This allows a token scoped to activation A to submit JOB messages for activation B by omitting `ansible_rulebook_id` from the payload.

```suggestion
            ansible_rulebook_id = data.get("ansible_rulebook_id")
            # Validate the ID that will be persisted
            if ansible_rulebook_id is not None:
                await self._validate_token_scope(ansible_rulebook_id)
            elif activation_id is not None:
                await self._validate_token_scope(activation_id)
```

src/aap_eda/api/serializers/auth.py
```diff
@@ -30,7 +30,16 @@ class RefreshTokenSerializer(serializers.Serializer):
 
     def validate(self, data):
         try:
+            # Validate the refresh token
             self.user = validate_jwt_token(data["refresh"], "refresh")
+
+            # Extract activation_instance_id from refresh token if present
+            # This is the trusted value from the token payload
+            token_payload = parse_jwt_token(data["refresh"])
```
wfealdel: **[Minor — Quality]** Double-parsing of the refresh token.

`validate_jwt_token` already calls `parse_jwt_token` internally, then `parse_jwt_token` is called a second time here to extract `activation_instance_id`. This redundantly decodes and re-validates the JWT signature. There is also a theoretical clock-boundary race where the token expires between the two calls, causing the first to succeed and the second to raise `ExpiredSignatureError`.

Consider refactoring `validate_jwt_token` to return both the `User` and the parsed payload, so the caller can extract `activation_instance_id` from the already-decoded result.

src/aap_eda/wsapi/consumers.py
```diff
@@ -109,26 +352,18 @@ async def receive(self, text_data=None, bytes_data=None):
...
-                await self.handle_events(AnsibleEventMessage.parse_obj(data))
-            elif msg_type == MessageType.ACTION:
-                await self.handle_actions(ActionMessage.parse_obj(data))
-            elif msg_type == MessageType.SHUTDOWN:
-                logger.info("Websocket connection is closed.")
-            elif msg_type == MessageType.SESSION_STATS:
-                await self.handle_heartbeat(HeartbeatMessage.parse_obj(data))
-            else:
-                logger.warning(f"Unsupported message received: {data}")
+            await self._validate_message_token_scope(msg_type, data)
```
wfealdel: **[Major — Security]** DB lookup on untrusted input before scope validation.

`_set_log_tracking_id` reads `activation_id` from the untrusted incoming message and does a DB query (`RulebookProcess.objects.get`) **before** this scope validation runs. Two concerns:

1. An attacker can force DB queries with arbitrary activation IDs before the token scope check rejects them
2. If the ID does not exist, `ObjectDoesNotExist` is raised outside the `try` block, so it does not follow the intended auth-failure path

Suggestion: move `await self._set_log_tracking_id(data)` after `_validate_message_token_scope` and before `_dispatch_message`, inside the `try` block.

src/aap_eda/services/auth.py
```diff
@@ -104,6 +154,21 @@ def validate_jwt_token(token: str, token_type: str) -> User:
...
+    # Validate activation_instance_id scope if required
+    if activation_instance_id is not None:
+        token_activation_id = token_dict.get("activation_instance_id")
+        if token_activation_id is None:
+            raise InvalidTokenError(
+                "Token is not scoped to an activation instance"
+            )
+        # Compare as strings to handle both int and UUID types
+        if str(token_activation_id) != str(activation_instance_id):
+            raise InvalidTokenError(
```
wfealdel, line 174: **[Minor — Security]** Error message leaks activation instance IDs.

This `InvalidTokenError` message includes both the token's `activation_instance_id` and the requested one. In the WebSocket consumer, this is logged at ERROR level via the `InvalidTokenError` handler.

Consider using a generic error message in the exception (e.g., "Token scope does not match the requested activation instance") and keeping the detailed IDs in server-side debug logs only.

src/aap_eda/wsapi/consumers.py
```diff
@@ -100,6 +102,247 @@ class Event(Enum):
...
+            dict: The decoded JWT token payload
+
+        Raises:
+            InvalidTokenError: If token is missing or invalid
+        """
+        if self._token_payload is not None:
+            return self._token_payload
+
+        # Get the authorization header from the scope
+        headers = dict(self.scope.get("headers", []))
```
wfealdel: **[Minor — Robustness]** `dict()` conversion drops duplicate headers.

`dict(self.scope.get("headers", []))` silently drops duplicate headers with the same key. Low risk since `Authorization` is typically single-valued in ASGI scopes

src/aap_eda/wsapi/consumers.py
```diff
@@ -100,6 +102,247 @@ class Event(Enum):
...
+                raise InvalidTokenError(
+                    f"Inconsistent activation IDs in JOB message: "
+                    f"activation_id={activation_id}, "
+                    f"ansible_rulebook_id={ansible_rulebook_id}"
+                )
+        # Security: For ANSIBLE_EVENT messages, resolve activation from
+        # job_id to prevent scope bypass
+        elif msg_type == MessageType.ANSIBLE_EVENT:
+            logger.warning(
+                "Received deprecated ANSIBLE_EVENT message. "
```
wfealdel: **[Minor ]**  It seems that this ANSIBLE_EVENT deprecation warning logs even if the message subsequently fails scope validation

wfealdel (commented):
Review of JWT activation instance scoping — 2 major findings (JOB scope bypass, DB lookup before auth) and 4 minor items. See inline comments for details.

wfealdel (dismissed):

src/aap_eda/wsapi/consumers.py
```diff
@@ -100,35 +102,221 @@ class Event(Enum):
...
+            self._log_message_error(err, msg_type, data)
         except InvalidEnvKeyError as err:
-            logger.error(f"Failed to parse {data} due to Env error: {err}")
+            self._log_message_error(err, msg_type, data)
+        except InvalidTokenError as err:
+            logger.error(  # NOSONAR
+                f"Token auth failed for message (type={msg_type.value}, "
+                f"activation_id={data.get('activation_id')}): {err}"
+            )
+            await self.close(code=4003)
```
kaiokmo: Unrelated to CodeRabbit's comment:

@mkanoor do we want to document the usage of 4003? I actually want to suggest moving this to a named constant, like `WS_CLOSE_AUTH_FAILURE = 4003`, and adding a brief comment or a docstring explaining the custom close code meaning. WDYT?

kaiokmo (changes requested):

kaiokmo (approved):

ttuffin (approved):

PR #1611, opened 2026-07-09
fix: sonar issues - Unused function parameters

jcraiglo1:

One of multiple PRs addressing sonar quality gate issues. Removes unused function parameters.

Attached is a document describing the actions taken here.

[sonarcloud-remediation-report.md](https://github.com/user-attachments/files/29103602/sonarcloud-remediation-report.md)

Assisted-by: Claude

## Summary by CodeRabbit

* **Chores / Refactor**
  * Clarified unused or internal method parameters across shared-resource validation, pod startup waiting, and revision parsing by renaming them without changing runtime behavior or error handling.

### DIFF
--- a/src/aap_eda/services/project/scm.py
+++ b/src/aap_eda/services/project/scm.py
@@ -132,7 +132,7 @@ class ScmRepository:
         self._executor = _executor
         self.git_hash = None
 
-    def rev_parse(self, rev: str) -> str:
+    def rev_parse(self, _rev: str) -> str:
         """
         Return object identifier for the given revision specifier.
 
@@ -359,10 +359,10 @@ class GitAnsibleRunnerExecutor:
                 )
                 if match:
                     return match.group(1)
-            match = re.search(
-                r'fatal: \[localhost\]: FAILED! => \{.+"msg": (.+)\}',
-                outputs.getvalue(),
-            )
+            output = outputs.getvalue()
+            match = None
+            if "fatal: [localhost]: FAILED!" in output and '"msg":' in output:
+                match = re.search(r'"msg": ([^}]+)\}', output)
             if match:
                 err_msg = match.group(1)
                 if "Authentication failed" in err_msg:
--- a/src/aap_eda/services/project/scm.py
+++ b/src/aap_eda/services/project/scm.py
@@ -380,10 +380,10 @@ class GitAnsibleRunnerExecutor:
                 )
                 if match:
                     return match.group(1)
-            output = outputs.getvalue()
-            match = None
-            if "fatal: [localhost]: FAILED!" in output and '"msg":' in output:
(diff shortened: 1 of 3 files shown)


src/aap_eda/services/project/scm.py
```diff
@@ -359,10 +359,10 @@ def __call__(
                 )
                 if match:
                     return match.group(1)
-            match = re.search(
-                r'fatal: \[localhost\]: FAILED! => \{.+"msg": (.+)\}',
-                outputs.getvalue(),
-            )
+            output = outputs.getvalue()
```
jcraiglo1: This had to be done to comply with sonar. The other way to handle this would've been to make a pr just to get this in so that this PR can go in. Fun new world we live in 😃

mkanoor (dismissed):

src/aap_eda/services/project/scm.py
ttuffin: Claude flagged this change with the following concerns:

  1. The new pattern [^}]+ stops at the first } it finds. The original used .+ (greedy, with dotall default off but single-line) across the whole \{.+\} structure. Real Ansible FAILED output can look like:
  fatal: [localhost]: FAILED! => {"changed": false, "msg": "some error"}
  1. Here [^}]+ before \} would match "some error" correctly. But if msg value itself contains a } (e.g., JSON or dict-like content in the message string), the new pattern truncates prematurely.
  2. The guard condition '"msg":' in output (note: no space) will match "msg": anywhere in the output, not just within the fatal error block. The original regex was anchored to the fatal: [localhost]: FAILED! => \{...\} structure.
  The new split check is less precise — it could be triggered by "msg": appearing in unrelated output lines and then try to extract from the wrong location.
  3. The PR description claims this was a "sonar unused parameter" fix. But the regex in __call__ has no unused parameters — this regex change appears unrelated to the S1172 issue and should not be bundled with parameter renames.
  It changes functional behavior and deserves its own PR and review.
  4. No tests cover GitAnsibleRunnerExecutor.__call__ directly — the unit tests mock the executor entirely and never exercise the real __call__ method's error-parsing logic. The regression risk is undetected by the test suite.

Risk: Medium. The new regex is more fragile for edge cases where msg values contain } characters or where "msg": appears in other output sections.

jcraiglo1: This seems to be a bit of a mess so I am going to make a new PR for this change. I'll link it here when it's up and remove this regex update

jcraiglo1: https://github.com/ansible/eda-server/pull/1613

jcraiglo1: Removed this update

mkanoor (approved):

B-Whitt (approved):

PR #1613, opened 2026-07-13
fix: sonar issues: Add test cases to SCM.py and replace regex with the ansible runner API for event reading

jcraiglo1:
Sonarcloud is complaining about a regex in a different PR. This PR gets rid of that regex in favor of utilizing the ansible runner API. Sonar has since scanned this file and determined that the tests only cover 83% of code in this file and is gating. So this PR will also add test cases to the rest of the file.  

Assisted-By: Claude

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SCM/Ansible execution parsing by using structured failure events to generate more accurate, credential-specific error messages.
  * Enhanced SCM version hash extraction on successful runs, including better handling when version details are missing.
  * Improved clone behavior around SSH/GPG credentials and key management, including proper environment setup.

* **Tests**
  * Added/expanded unit tests for proxy environment restoration, SSH key cloning behavior, GPG verification environment, directory creation, decryption/add-GPG success and failure paths, and executor event extraction.

### DIFF
--- a/src/aap_eda/services/project/scm.py
+++ b/src/aap_eda/services/project/scm.py
@@ -380,12 +380,8 @@ class GitAnsibleRunnerExecutor:
                 )
                 if match:
                     return match.group(1)
-            match = re.search(
-                r'fatal: \[localhost\]: FAILED! => \{.+"msg": (.+)\}',
-                outputs.getvalue(),
-            )
-            if match:
-                err_msg = match.group(1)
+            err_msg = self._extract_error_msg(outputs.getvalue())
+            if err_msg:
                 if "Authentication failed" in err_msg:
                     raise ScmAuthenticationError("Authentication failed")
                 if (
@@ -396,6 +392,29 @@ class GitAnsibleRunnerExecutor:
                     raise ScmAuthenticationError(err_msg)
                 raise ScmError(f"{self.ERROR_PREFIX} {err_msg}")
             raise ScmError(f"{self.ERROR_PREFIX} {outputs.getvalue().strip()}")
+
+    @staticmethod
+    def _extract_error_msg(output):
+        """Extract msg value from Ansible FAILED output.
+
+        Looks for the fatal error marker, then finds the "msg"
+        key and extracts its value up to the last closing brace.
+        Handles msg values that contain } characters.
+        """
+        fatal_marker = "fatal: [localhost]: FAILED! => {"
+        fatal_idx = output.find(fatal_marker)
+        if fatal_idx < 0:
+            return None
+        block = output[fatal_idx + len(fatal_marker) :]
+        msg_marker = '"msg": '
+        msg_idx = block.find(msg_marker)
+        if msg_idx < 0:
+            return None
+        msg_value = block[msg_idx + len(msg_marker) :]
+        brace_idx = msg_value.rfind("}")
+        if brace_idx < 0:
+            return None
+        return msg_value[:brace_idx]
 
 
 def is_refspec_valid(refspec: str, is_branch: bool) -> bool:
--- a/src/aap_eda/services/project/scm.py
+++ b/src/aap_eda/services/project/scm.py
@@ -14,10 +14,8 @@
 from __future__ import annotations
 
 import contextlib
-import io
 import logging
 import os
-import re
 import shutil
 import subprocess
 import tempfile
@@ -365,22 +363,20 @@ class GitAnsibleRunnerExecutor:
             os.makedirs(env_dir)
             safe_yaml.dump(os.path.join(env_dir, "extravars"), extra_vars)
 
-            outputs = io.StringIO()
-            with contextlib.redirect_stdout(outputs):
-                runner = ansible_runner.run(
-                    private_data_dir=data_dir,
-                    playbook=PLAYBOOK,
-                    envvars=env_vars,
-                )
+            runner = ansible_runner.run(
+                private_data_dir=data_dir,
+                playbook=PLAYBOOK,
+                envvars=env_vars,
+                quiet=True,
+            )
+
+            events = list(runner.events)
 
             if runner.rc == 0:
-                match = re.search(
-                    r'"msg": "Repository Version ([0-9a-fA-F]+)"',
-                    outputs.getvalue(),
-                )
-                if match:
-                    return match.group(1)
-            err_msg = self._extract_error_msg(outputs.getvalue())
+                git_hash = self._extract_git_hash(events)
+                if git_hash:
+                    return git_hash
+            err_msg = self._extract_error_msg(events)
             if err_msg:
                 if "Authentication failed" in err_msg:
                     raise ScmAuthenticationError("Authentication failed")
@@ -391,30 +387,29 @@ class GitAnsibleRunnerExecutor:
                     err_msg = "Credentials not provided or incorrect"
                     raise ScmAuthenticationError(err_msg)
                 raise ScmError(f"{self.ERROR_PREFIX} {err_msg}")
-            raise ScmError(f"{self.ERROR_PREFIX} {outputs.getvalue().strip()}")
+            raise ScmError(self.ERROR_PREFIX)
 
     @staticmethod
-    def _extract_error_msg(output):
-        """Extract msg value from Ansible FAILED output.
+    def _extract_git_hash(events):
+        for event in events:
+            if event.get("event") != "runner_on_ok":
+                continue
+            res = event.get("event_data", {}).get("res", {})
+            scm_version = res.get("ansible_facts", {}).get("scm_version")
+            if scm_version:
+                return scm_version
+        return None
 
-        Looks for the fatal error marker, then finds the "msg"
-        key and extracts its value up to the last closing brace.
-        Handles msg values that contain } characters.
-        """
-        fatal_marker = "fatal: [localhost]: FAILED! => {"
-        fatal_idx = output.find(fatal_marker)
-        if fatal_idx < 0:
-            return None
-        block = output[fatal_idx + len(fatal_marker) :]
-        msg_marker = '"msg": '
-        msg_idx = block.find(msg_marker)
-        if msg_idx < 0:
-            return None
-        msg_value = block[msg_idx + len(msg_marker) :]
-        brace_idx = msg_value.rfind("}")
-        if brace_idx < 0:
-            return None
-        return msg_value[:brace_idx]
+    @staticmethod
+    def _extract_error_msg(events):
+        for event in events:
+            if event.get("event") != "runner_on_failed":
+                continue
+            res = event.get("event_data", {}).get("res", {})
+            msg = res.get("msg")
+            if msg:
+                return msg
+        return None
 
 
 def is_refspec_valid(refspec: str, is_branch: bool) -> bool:
--- a/tests/unit/test_scm.py
+++ b/tests/unit/test_scm.py
@@ -452,99 +452,78 @@ def test_git_clone_ssh_key(ssh_credential: models.EdaCredential, url: str):
 
 
 #################################################################
+# Tests for GitAnsibleRunnerExecutor._extract_git_hash
+#################################################################
+
+_extract_hash = scm.GitAnsibleRunnerExecutor._extract_git_hash
+_extract_error = scm.GitAnsibleRunnerExecutor._extract_error_msg
+
+
+def _ok_event(res=None):
+    return {"event": "runner_on_ok", "event_data": {"res": res or {}}}
+
+
+def _failed_event(res=None):
+    return {
+        "event": "runner_on_failed",
+        "event_data": {"res": res or {}},
+    }
+
+
+def test_extract_git_hash_from_set_fact():
+    events = [
+        _ok_event({"changed": True}),
+        _ok_event(
+            {
+                "ansible_facts": {
+                    "scm_version": "abc123def456",
+                },
+            }
+        ),
+        _ok_event({"msg": "Repository Version abc123def456"}),
+    ]
+    assert _extract_hash(events) == "abc123def456"
+
+
+def test_extract_git_hash_no_matching_event():
+    events = [
+        _ok_event({"changed": True}),
+        {"event": "runner_on_start", "event_data": {}},
+    ]
+    assert _extract_hash(events) is None
+
+
+def test_extract_git_hash_empty_events():
+    assert _extract_hash([]) is None
+
+
+#################################################################
 # Tests for GitAnsibleRunnerExecutor._extract_error_msg
 #################################################################
 
-_extract = scm.GitAnsibleRunnerExecutor._extract_error_msg
+
+def test_extract_error_msg_from_failed_event():
+    events = [_failed_event({"msg": "Authentication failed"})]
+    assert _extract_error(events) == "Authentication failed"
 
 
-def test_extract_error_msg_standard():
-    output = (
-        "fatal: [localhost]: FAILED! => "
-        '{"changed": false, "msg": "Authentication failed"}'
-    )
-    assert _extract(output) == '"Authentication failed"'
+def test_extract_error_msg_no_failure_event():
+    events = [_ok_event({"changed": True})]
+    assert _extract_error(events) is None
 
 
-def test_extract_error_msg_multiple_keys():
-    output = (
-        "fatal: [localhost]: FAILED! => "
-        '{"changed": false, "rc": 1, '
-        '"msg": "Could not clone repo"}'
-    )
-    assert _extract(output) == '"Could not clone repo"'
+def test_extract_error_msg_missing_msg_key():
+    events = [_failed_event({"changed": False, "rc": 1})]
+    assert _extract_error(events) is None
 
 
-def test_extract_error_msg_only_key():
-    output = (
-        "fatal: [localhost]: FAILED! => " '{"msg": "could not read Username"}'
-    )
-    assert _extract(output) == '"could not read Username"'
+def test_extract_error_msg_empty_msg():
+    events = [_failed_event({"msg": ""})]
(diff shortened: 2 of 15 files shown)


src/aap_eda/services/project/scm.py
```diff
@@ -397,6 +393,29 @@ def __call__(
...
+        fatal_idx = output.find(fatal_marker)
+        if fatal_idx < 0:
+            return None
+        block = output[fatal_idx + len(fatal_marker) :]
+        msg_marker = '"msg": '
+        msg_idx = block.find(msg_marker)
+        if msg_idx < 0:
+            return None
+        msg_value = block[msg_idx + len(msg_marker) :]
+        brace_idx = msg_value.rfind("}")
```
wfealdel: **Minor: `rfind("}")` scans beyond the fatal JSON line**

`rfind("}")` searches the entire remaining output (including subsequent lines), not just the fatal error's JSON block. If Ansible runner output after the fatal line contains `}` (e.g., rescue handlers, subsequent tasks), the extracted message would include garbage text. The old regex was naturally line-bounded since `.` doesn't match newlines.

Suggested fix — bound `block` to the same line:
```python
block_start = fatal_idx + len(fatal_marker)
newline_idx = output.find("\n", block_start)
block = output[block_start:newline_idx] if newline_idx >= 0 else output[block_start:]
```

src/aap_eda/services/project/scm.py
```diff
@@ -380,12 +380,8 @@ def __call__(
                 )
                 if match:
                     return match.group(1)
```
wfealdel: **Question:** This PR replaces the `re.search` for error extraction with string operations, but the success path still uses `re.search` to extract the Repository Version. Was keeping this regex intentional (e.g., SonarCloud only flagged the other one), or should it also be migrated for consistency?

src/aap_eda/services/project/scm.py
```diff
@@ -397,6 +393,29 @@ def __call__(
                 raise ScmError(f"{self.ERROR_PREFIX} {err_msg}")
             raise ScmError(f"{self.ERROR_PREFIX} {outputs.getvalue().strip()}")
 
+    @staticmethod
+    def _extract_error_msg(output):
```
wfealdel: **Observation: `runner.events` provides structured access to failure data**

The error extraction chain (`redirect_stdout` → parse `fatal: [localhost]: FAILED! => {JSON}` text → substring match) depends on Ansible's callback output format, which is not a stable API. `ansible_runner` already provides structured event data via `runner.events` — filtering for `event == 'runner_on_failed'` gives `event_data['res']['msg']` directly from the module result dict, no text parsing needed. 

Not a blocker for this PR (pre-existing architecture), but worth considering as a follow-up to reduce fragility.

wfealdel (commented):
Reviewed PR #1613. One minor finding (rfind scoping), one consistency question, and one architectural observation. See inline comments for details.

ptoscano (changes requested):
Rather than "fix the regexp", I think the code needs to use better the `ansible_runner` module; in particular, the `Runner` object returned by `ansible_runner.run()` seem to provide already all the needed pieces, without any need to manual parsing:
https://docs.ansible.com/projects/runner/en/latest/python_interface/

My suggestion here is to create a different PR with the better usage of the `ansible_runner` API.

src/aap_eda/services/project/scm.py
jcraiglo1: Sonar only flagged the one. However, in implementing Pino's request that has now been removed as well

src/aap_eda/services/project/scm.py
jcraiglo1: Changed!

src/aap_eda/services/project/scm.py
```diff
@@ -132,7 +130,7 @@ def __init__(
         self._executor = _executor
         self.git_hash = None
 
-    def rev_parse(self, rev: str) -> str:
+    def rev_parse(self, _rev: str) -> str:
```
ptoscano, line 133: This was fixed in `main`, so you need to rebase this PR.

tests/unit/test_scm.py
```diff
@@ -156,6 +156,44 @@ def spy_get(obj):
...
+        ):
+            assert os.environ.get(key) == original_value
+    finally:
+        for key in (
+            "http_proxy",
+            "https_proxy",
+            "HTTP_PROXY",
+            "HTTPS_PROXY",
+        ):
+            os.environ.pop(key, None)
```
ptoscano: Theoretically speaking: if the environment/system that runs these tests has any of the http proxy envvars set (because the system needs a network proxy), at the end of the execution of this test the old values will be lost.

Regardless: use the [monkeypath](https://docs.pytest.org/en/stable/reference/reference.html#std-fixture-monkeypatch) fixture instead.

tests/unit/test_scm.py
```diff
@@ -411,3 +449,303 @@ def test_git_clone_ssh_key(ssh_credential: models.EdaCredential, url: str):
...
+            assert "GNUPGHOME" in call_kwargs[1]["env_vars"]
+
+
+@pytest.mark.django_db
+def test_git_clone_creates_directory():
+    """Clone creates the destination directory if it doesn't exist."""
+    executor = mock.MagicMock()
+
+    with tempfile.TemporaryDirectory() as parent:
+        dest_path = os.path.join(parent, "nonexistent")
```
ptoscano, line 568:
```suggestion
        dest_path = parent / "nonexistent"
```

tests/unit/test_scm.py
```diff
@@ -411,3 +449,303 @@ def test_git_clone_ssh_key(ssh_credential: models.EdaCredential, url: str):
...
+        scm.ScmRepository.clone(
+            "https://git.example.com/repo.git",
+            dest_path,
+            _executor=executor,
+        )
+        assert os.path.isdir(dest_path)
```
ptoscano, line 574: With the suggestion above applied, this can be:

```suggestion
        assert dest_path.is_dir()
```

tests/unit/test_scm.py
```diff
@@ -411,3 +449,303 @@ def test_git_clone_ssh_key(ssh_credential: models.EdaCredential, url: str):
...
+                "git@git.example.com:repo.git",
+                dest_path,
+                credential=credential,
+                _executor=executor,
+            )
+        mock_decrypt.assert_called_once()
+
+
+@pytest.mark.django_db
+def test_git_clone_gpg_add_key(
```
ptoscano: This test is practically redundant, as `test_git_clone_gpg_credential` does everything this one does, and more. Simply add the `assert_called_once()` call to `test_git_clone_gpg_credential`, and drop this.

tests/unit/test_scm.py
```diff
@@ -411,3 +449,303 @@ def test_git_clone_ssh_key(ssh_credential: models.EdaCredential, url: str):
...
+    """Clone calls decrypt_key_file when ssh_key_unlock is provided."""
+    credential = models.EdaCredential.objects.create(
+        name="test-ssh-unlock",
+        inputs={
+            "ssh_key_data": "-----BEGIN OPENSSH PRIVATE KEY-----\n",
+            "ssh_key_unlock": "passphrase",
+        },
+        organization=default_organization,
+    )
+    credential.refresh_from_db()
```
ptoscano: This is the `ssh_credential` fixture, simply use it.

tests/unit/test_scm.py
```diff
@@ -411,3 +449,303 @@ def test_git_clone_ssh_key(ssh_credential: models.EdaCredential, url: str):
...
+def test_git_clone_gpg_credential(
+    default_organization: models.Organization,
+):
+    """GPG credential sets up verify_commit and GNUPGHOME."""
+    gpg_credential = models.EdaCredential.objects.create(
+        name="test-gpg-credential",
+        inputs={"gpg_public_key": "-----BEGIN PGP PUBLIC KEY BLOCK-----\n"},
+        organization=default_organization,
+    )
+    gpg_credential.refresh_from_db()
```
ptoscano: Move this into a new `gpg_credential` fixture, consistent with `ssh_credential`.

ptoscano (changes requested):
Thanks for the updates, in general it is a good improvement. I left some more notes, I should had reviewed it fully previously, sorry.

In addition to the notes, this PR needs:
- to be rebased
- a PR description that explains better what is being done, and why

tests/unit/test_scm.py
jcraiglo1: What are we trying to accomplish here that isn't done by os.path.join? The / operator only works for path objects and parent is actually a str in this case

ptoscano: Oh right, I thought `TemporaryDirectory()` could return a `pathlib.Path` -- nevermind this, then!

tests/unit/test_scm.py
jcraiglo1: Discussed offline, keeping as a str object and as such os.path.isdir(dest_patth) is the right option.

tests/unit/test_scm.py
jcraiglo1: Agree, addressed

tests/unit/test_scm.py
jcraiglo1: Discussed. Updated the fixture and merged this test into test_git_clone_ssh_key

tests/unit/test_scm.py
jcraiglo1: Addressed

tests/unit/test_scm.py
jcraiglo1: Addresses

ptoscano (approved):
Thanks for the updates.

One potential improvement that could be done later (no need here) is to use a hook in `ansible_runner.run()` to process the events as they come, rather than iterating at the end. The current way is still better than checking/matching the whole output of the playbook.

Please do amend the title & description of this PR to reflect the actual changes done, since they are partially obsolete now.

src/aap_eda/services/project/scm.py
```diff
@@ -395,7 +387,29 @@ def __call__(
                     err_msg = "Credentials not provided or incorrect"
                     raise ScmAuthenticationError(err_msg)
                 raise ScmError(f"{self.ERROR_PREFIX} {err_msg}")
-            raise ScmError(f"{self.ERROR_PREFIX} {outputs.getvalue().strip()}")
+            raise ScmError(self.ERROR_PREFIX)
+
+    @staticmethod
+    def _extract_git_hash(events):
```
wfealdel, line 393: @jcraiglo1 
The logic here is similar to `_extract_error_msg()` . Consider refactor the function with parameters so it can be reused. Maybe something like this:

```
@staticmethod
  def _extract_from_events(
      events: list[dict[str, Any]],
      event_type: str,
      extract: Callable[[dict[str, Any]], Any | None],
  ) -> Any | None:
      """Extract a value from the first matching event."""
      for event in events:
          if event.get("event") != event_type:
              continue
          res = event.get("event_data", {}).get("res", {})
          value = extract(res)
          if value:
              return value
      return None
```

  Then replace the call sites:

  ```
# Was: self._extract_git_hash(events)
  self._extract_from_events(
      events,
      "runner_on_ok",
      lambda res: res.get("ansible_facts", {}).get("scm_version"),
  )
```

 ```
 # Was: self._extract_error_msg(events)
  self._extract_from_events(
      events,
      "runner_on_failed",
      lambda res: res.get("msg"),
  )
```

jcraiglo1: I definitely agree that we can save some duplication but I think we'd be trading in a significant amount of readability and it's making me resistant to this change. Thoughts?

wfealdel: I think this is worth to consider as it would be a nice improve if a third extraction is ever needed (e.g., extracting runner.stats or another event field), but you have a valid concern and this is not a blocking issue.

src/aap_eda/services/project/scm.py
```diff
@@ -395,7 +387,29 @@ def __call__(
                     err_msg = "Credentials not provided or incorrect"
                     raise ScmAuthenticationError(err_msg)
                 raise ScmError(f"{self.ERROR_PREFIX} {err_msg}")
-            raise ScmError(f"{self.ERROR_PREFIX} {outputs.getvalue().strip()}")
+            raise ScmError(self.ERROR_PREFIX)
```
wfealdel, line 390: **[Minor]** The previous code included `outputs.getvalue().strip()` in this fallback error, giving operators the full runner output for debugging. The new version raises a bare `ScmError(self.ERROR_PREFIX)` with no diagnostic context. Consider appending `runner.stdout.read()` or a summary of the events so failures without a `runner_on_failed` event are still diagnosable.

src/aap_eda/services/project/scm.py
```diff
@@ -365,27 +363,21 @@ def __call__(
...
-                    return match.group(1)
-            match = re.search(
-                r'fatal: \[localhost\]: FAILED! => \{.+"msg": (.+)\}',
-                outputs.getvalue(),
-            )
-            if match:
-                err_msg = match.group(1)
+                git_hash = self._extract_git_hash(events)
+                if git_hash:
+                    return git_hash
```
wfealdel, line 378: **[Minor]** When `runner.rc == 0` but `_extract_git_hash` returns `None` (playbook succeeded but didn't set the `scm_version` fact), the code falls through to the error path and raises the same `ScmError(self.ERROR_PREFIX)` as a runner failure with no error event. A distinct message like `f"{self.ERROR_PREFIX} Runner succeeded but no version hash was found"` would make this case easier to debug.

tests/unit/test_scm.py
```diff
@@ -393,21 +439,267 @@ def test_is_refspec_valid(ref: str, is_branch: bool, expected: bool):
...
+    assert "Project Import Error:" in str(exc_info.value)
+
+
+def test_executor_call_auth_failure():
+    events = [_failed_event({"msg": "Authentication failed"})]
+    with pytest.raises(scm.ScmAuthenticationError):
+        _run_executor(1, events)
+
+
+def test_executor_call_username_prompt():
```
wfealdel, line 686: **[Minor]** This test covers the `"could not read Username"` branch, but there's no corresponding test for `"could not read Password"` (scm.py line 385). Both sides of the `or` condition should be exercised.

tests/unit/test_scm.py
```diff
@@ -393,21 +439,267 @@ def test_is_refspec_valid(ref: str, is_branch: bool, expected: bool):
...
+
+
+def _failed_event(res=None):
+    return {
+        "event": "runner_on_failed",
+        "event_data": {"res": res or {}},
+    }
+
+
+def test_extract_git_hash_from_set_fact():
```
wfealdel, line 485: **[Nit]** The third event `_ok_event({"msg": "Repository Version abc123def456"})` simulates the old message-based output but is never matched by the new `_extract_git_hash` (which looks for `ansible_facts.scm_version`). It's dead data that may mislead future readers into thinking the `msg` field matters. Consider removing it or adding a comment explaining it's intentionally inert.

src/aap_eda/services/project/scm.py
```diff
@@ -395,7 +387,29 @@ def __call__(
...
                 raise ScmError(f"{self.ERROR_PREFIX} {err_msg}")
-            raise ScmError(f"{self.ERROR_PREFIX} {outputs.getvalue().strip()}")
+            raise ScmError(self.ERROR_PREFIX)
+
+    @staticmethod
+    def _extract_git_hash(events):
+        for event in events:
+            if event.get("event") != "runner_on_ok":
+                continue
+            res = event.get("event_data", {}).get("res", {})
```
wfealdel, line 397: **[Nit]** If an event has explicit `"event_data": None`, `.get("event_data", {})` returns `None` (the default only applies when the key is absent), and the chained `.get("res", {})` would raise `AttributeError`. Using `event.get("event_data") or {}` would handle both missing keys and `None` values.

tests/unit/test_scm.py
```diff
@@ -393,21 +439,267 @@ def test_is_refspec_valid(ref: str, is_branch: bool, expected: bool):
...
+            scm.ScmRepository.add_gpg_key("/tmp/gpgkey", "/tmp/gnupg")
+        assert "Failed to import" in str(exc_info.value)
+
+
+#################################################################
+# Tests for GitAnsibleRunnerExecutor.__call__
+#################################################################
+
+
+def _run_executor(runner_rc, events):
```
wfealdel, line 648: **[Nit]** The `_run_executor` helper mocks `ansible_runner.run` but never asserts that `quiet=True` is passed. If someone removes `quiet=True` from the production code, tests would still pass but ansible_runner would print to stdout. Consider asserting the `quiet` kwarg.

wfealdel (commented):
4 Minor findings and 3 Nits — see inline comments for details. No blockers.

src/aap_eda/services/project/scm.py
ptoscano: If you want to refactor, see my other comment: it'd be better as event callback, processing events as they are received.

IMHO the proposed `_extract_from_events()` is a bit overkill, since to make it generic enough it is more complex than what the current inline functions do.

src/aap_eda/services/project/scm.py
ptoscano: While defensive programming is good, would that situation ever happen?

src/aap_eda/services/project/scm.py
ptoscano: AFAICT, the behaviour is the same before these changes: if `runner.rc == 0` and the `re.search()` doesn't return any match, it falls through the error extraction.

So I'd say a potential improvement for a separate PR.

wfealdel (approved):
LGTM! Added some findings that are not blockers for this PR

PR #1619, opened 2026-07-16
fix: prevent ghost pods.

AlexSCorey:
When the monitor detects an unresponsive activation and triggers a
  restart, _watch_pod_deletion falsely reports success if the watch
  stream times out without receiving a DELETED event. This allows
  _unresponsive_policy to restart the activation while the original
  pod is still running, producing ghost pods that compete for shared
  resources like Kafka consumer groups.

  Two fixes:

  1. _watch_pod_deletion now tracks whether a DELETED event was actually
     received. If the watch times out without one, it retries instead of
     reporting success. After all retries, it raises ContainerCleanupError.

  2. _unresponsive_policy now calls _cleanup explicitly via a new
     _handle_unresponsive_activation helper. If cleanup fails, the
     restart is skipped to avoid creating a duplicate pod. The activation
     is still marked FAILED regardless of cleanup outcome.

  Fixes: AAP-83199

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved unresponsive activation handling by centralizing the unresponsive workflow, ensuring cleanup failures reliably prevent restart scheduling and activations are consistently marked **FAILED** with the right per-check message.
  * Enhanced Kubernetes job/pod cleanup to only report success after confirming the pod deletion event, with stronger watch/retry behavior (including more transient error handling).
* **Tests**
  * Expanded integration coverage for Kubernetes deletion watches, including deterministic cursor handling, timeout/empty-stream scenarios, and successful deletion logging.
  * Added integration tests validating unresponsive activation behavior across restart-policy scenarios, including cleanup-failure re-raise behavior and status updates.


src/aap_eda/services/activation/activation_manager.py
```diff
@@ -377,45 +377,52 @@ def _unresponsive_policy(self, check_type: str):
...
+                self.db_instance_type, self.db_instance.id, delay_seconds=1
             )
 
-        else:
-            LOGGER.info(
+    def _handle_unresponsive_activation(self, check_type: str):
+        try:
+            self._cleanup()
+        except engine_exceptions.ContainerCleanupError:
+            raise
```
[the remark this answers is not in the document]
AlexSCorey: not an issue.

src/aap_eda/services/activation/activation_manager.py
```diff
@@ -377,45 +377,59 @@ def _unresponsive_policy(self, check_type: str):
...
+    def _handle_unresponsive_activation(self, check_type: str):
+        """Clean up and fail an unresponsive activation.
+
+        Attempts cleanup of the container, then marks the activation
+        as FAILED regardless of cleanup outcome. If cleanup raises
+        ContainerCleanupError, the exception propagates to the caller
+        after the activation is marked FAILED.
+        """
+        try:
+            self._cleanup()
```
wfealdel: **[Critical]** `_cleanup()` (line 294) catches `ContainerCleanupError` internally and returns normally — it never lets the exception escape. So this `try` block will never see a `ContainerCleanupError`, the `except` in `_unresponsive_policy` (line 382) will never trigger, and `system_restart_activation` will always be called when the restart policy allows it — even when the pod is still running.

Consider either calling `self.container_engine.cleanup()` directly here (bypassing `_cleanup`'s exception swallowing), or adding a `raise_on_error` parameter to `_cleanup()` so this caller can opt in to propagation while preserving best-effort behavior for existing callers.

src/aap_eda/services/activation/activation_manager.py
```diff
@@ -377,45 +377,59 @@ def _unresponsive_policy(self, check_type: str):
...
+                log_message += " Restart policy is not applicable."
+                user_msg += " Restart policy is not applicable."
+            else:
+                log_message += " Restart policy is applied."
+                user_msg += " Activation is going to be restarted."
+
+            LOGGER.info(log_message)
+
             container_logger.write(user_msg, flush=True)
             self._fail_instance(msg=user_msg)
```
wfealdel: **[Major]** `_fail_instance()` calls `_cleanup()` again internally (line 651). Since `_handle_unresponsive_activation` already called `_cleanup()` at line 403, this results in two cleanup round-trips to the Kubernetes API per unresponsive event. If the first cleanup succeeded, the second one hits the API for already-deleted resources.

tests/integration/services/activation/test_manager.py
```diff
@@ -999,3 +999,149 @@ def test_stop_cancels_auto_restart_when_no_instance(
...
+
+    cleanup_calls = 0
+
+    def cleanup_side_effect():
+        nonlocal cleanup_calls
+        cleanup_calls += 1
+        if cleanup_calls == 1:
+            raise engine_exceptions.ContainerCleanupError("cleanup failed")
+
+    with patch.object(
```
wfealdel: **[Major]** All new manager tests mock `_cleanup` at the `ActivationManager` level, replacing it with a side_effect that raises `ContainerCleanupError`. But the real `_cleanup()` catches that exception internally and never raises it (line 294 of activation_manager.py). These tests verify the orchestration logic in isolation but don't exercise the production path — they give false confidence that the ghost pod scenario is prevented.

src/aap_eda/services/activation/activation_manager.py
```diff
@@ -377,45 +377,59 @@ def _unresponsive_policy(self, check_type: str):
...
-                "Activation is going to be restarted."
+                f"{check_type} check for ansible-rulebook timed out."
             )
+
+            if self.db_instance.restart_policy == RestartPolicy.NEVER:
+                log_message += " Restart policy is not applicable."
+                user_msg += " Restart policy is not applicable."
+            else:
+                log_message += " Restart policy is applied."
+                user_msg += " Activation is going to be restarted."
```
wfealdel: **[Minor]** This message says "Activation is going to be restarted" but it's in the `finally` block, which runs regardless of whether cleanup succeeded. If the `_cleanup` propagation issue is fixed so that `_unresponsive_policy` can skip the restart, this message will still promise a restart that won't happen.

src/aap_eda/services/activation/engine/kubernetes.py
```diff
@@ -660,7 +713,7 @@ def _process_pod_start_event(self, event) -> bool:
             )
         return False
 
-    def _wait_for_pod_to_start(self, _log_handler: LogHandler) -> None:
+    def _wait_for_pod_to_start(self, log_handler: LogHandler) -> None:
```
wfealdel: **[Minor]** The parameter was renamed from `_log_handler` to `log_handler`, removing the "unused" underscore convention, but it's still not used in the method body. This signals to readers that the parameter is now used when it isn't.

src/aap_eda/services/activation/engine/kubernetes.py
```diff
@@ -50,7 +50,7 @@
 
 K8S_API_RETRIES = 3
 K8S_API_RETRY_BACKOFF = 1.0
-K8S_API_TRANSIENT_STATUS_CODES = {401, 403, 500, 502, 503, 504}
+K8S_API_TRANSIENT_STATUS_CODES = {401, 403, 410, 500, 502, 503, 504}
```
wfealdel, line 55: **[Nit]** Adding `410` (Gone) is correct for the watch use case (stale resourceVersion), but `K8S_API_TRANSIENT_STATUS_CODES` is also used by `_call_k8s_api` which wraps all K8s API calls — not just watches. A 410 from a non-watch endpoint would now be silently retried. Low risk in practice since 410 is rare outside watch contexts, but worth noting the broader retry surface.

src/aap_eda/services/activation/engine/kubernetes.py
```diff
@@ -581,50 +594,90 @@ def _delete_job_resource(
...
-                        f"Error while waiting for deletion: {exc}",
-                        flush=True,
-                    )
-                    raise ContainerCleanupError(
-                        f"Error during cleanup: {str(exc)}"
-                    ) from exc
-                self._log_and_backoff(exc, attempt, desc)
             finally:
                 watcher.stop()
         raise ContainerCleanupError(
```
wfealdel, line 691: **[Minor]** When all retries are exhausted due to pure timeouts (the watch stream ends without a DELETED event, no `ApiException` raised), `last_exc` is still `None`. This produces the message `"watch pod deletion test-job failed after 3 retries: None"`. Consider a clearer message distinguishing timeout exhaustion from API errors, e.g. `"timed out waiting for pod deletion after {K8S_API_RETRIES} attempts"`.

wfealdel (commented):
@AlexSCorey see inline comments

wfealdel (approved):

kaiokmo (approved):

andresberejnoi (approved):

PR #1622, opened 2026-07-17
fix: sonar issues: Address security concerns regarding github actions

jcraiglo1:

One of multiple Sonarcloud PRs. Addresses multiple security concerns by:
 - Pinning all GitHub Actions to immutable commit SHAs to prevent supply-chain attacks via tag mutation 
 - Move secrets out of run: block interpolation into env: declarations in e2e-dispatch.yml and ui-e2e.yml 
 - Add persist-credentials: false to checkout steps that don't require authenticated git operations, reducing token exposure to subsequent steps 
 
 Assisted-by: Claude

## Summary by CodeRabbit

* **Security & Reliability**
  * Pinned third-party GitHub Actions to specific commit SHAs across linting, validation, CI, Docker image builds, E2E (UI/API), OpenAPI syncing, scheduled tests, and Sonar analysis for stronger supply-chain stability.
  * Disabled credential persistence (`persist-credentials: false`) on relevant checkout steps.

* **Workflow Maintenance**
  * Kept existing workflow behavior the same while improving determinism.
  * Updated E2E request/container credential handling to use step environment variables (including PDE uploads and UI E2E container runs).

### DIFF
--- a/.github/workflows/autoupdate-poetry.yml
+++ b/.github/workflows/autoupdate-poetry.yml
@@ -18,13 +18,13 @@ jobs:
       lock-changed: ${{ steps.lock-changed.outputs.lock-changed }}
     steps:
       - name: Checkout main branch
-        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           persist-credentials: false
           ref: main
 
       - name: Set up Python
-        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1  # v6
+        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
         with:
           python-version: "3.11"
 
--- a/.github/workflows/autoupdate-poetry.yml
+++ b/.github/workflows/autoupdate-poetry.yml
@@ -48,7 +48,7 @@ jobs:
           fi
 
       - name: Create a Pull Request
-        uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1  # v8
+        uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1 # v8.1.1
         if: steps.lock-changed.outputs.lock-changed == 'true'
         with:
           # TODO: uses eda-ci pat token, replace with a bot.
--- a/.github/workflows/autoupdate-poetry.yml
+++ b/.github/workflows/autoupdate-poetry.yml
@@ -18,12 +18,13 @@ jobs:
       lock-changed: ${{ steps.lock-changed.outputs.lock-changed }}
     steps:
       - name: Checkout main branch
-        uses: actions/checkout@v7
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7
         with:
+          persist-credentials: false
           ref: main
 
       - name: Set up Python
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1  # v6
         with:
           python-version: "3.11"
 
@@ -47,7 +48,7 @@ jobs:
           fi
 
       - name: Create a Pull Request
-        uses: peter-evans/create-pull-request@v8
+        uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1  # v8
         if: steps.lock-changed.outputs.lock-changed == 'true'
         with:
           # TODO: uses eda-ci pat token, replace with a bot.
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -18,10 +18,12 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v7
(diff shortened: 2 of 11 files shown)


.github/workflows/ui-e2e.yml
```diff
@@ -153,11 +154,14 @@ jobs:
...
+        env:
+          EDA_USERNAME: ${{ secrets.EDA_USERNAME }}
+          EDA_PASSWORD: ${{ secrets.EDA_PASSWORD }}
         run: |
           docker run --rm \
           -e EDA_SERVER=${{ env.EDA_API_URL }} \
-          -e EDA_USERNAME=${{ secrets.EDA_USERNAME }} \
-          -e EDA_PASSWORD=${{ secrets.EDA_PASSWORD }} \
+          -e EDA_USERNAME=$EDA_USERNAME \
+          -e EDA_PASSWORD=$EDA_PASSWORD \
```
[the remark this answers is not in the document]
jcraiglo1: addressed

ptoscano (changes requested):
Regarding the changes to github actions used, there things to change, according to the recent experience in ansible-rulebook; I'll take one one as example, it applies to all of the changes:

```yaml
      uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1  # v6
```
To make these still handled by dependabot, the format must be: `org/repo@sha1`, then **one** space (exactly), then `#`, then again **one** space, and the exact non-floating tag name; this means that
- the extra spaces between the hash and `#` need to removed to be only one
- `v6`/etc are "floating tags", representing the most recent version of that major series; since the hash represents a specific version, then the tag name must be the precise version of that hash; in case of the above, it's `v6.3.0`

------

> Move secrets out of run: block interpolation into env: declarations in e2e-dispatch.yml and ui-e2e.yml

I'm not sure what is the reason behind this; would you please detail more? (Bonus points to documentation pointers about this.)

------

Thanks!

jcraiglo1:
@ptoscano Hey Pino! Thank you for the dependabot instructions, I've got that resolved locally and will push in just a minute. 

Regarding the secrets in the run block, from my understanding it falls under injection prevention. Assuming someone can control the variables going into the script they are arbitrarily executed. Loading them into env first access them as variables instead of running it as code. Most of them (all of them?) come from secrets so theoretically an attacker would have to exploit the repo secrets as well in order to exploit the workflow in this way. But still technically possible. So github recommends not doing it this way and sonar is gating on it. 

Here are some docs I found:

The sonarcloud rule - https://sonarcloud.io/organizations/ansible/rules?open=githubactions%3AS7636&rule_key=githubactions%3AS7636

Github's recommendation that Sonar is basing this off of https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable

ptoscano:
> Regarding the secrets in the run block, from my understanding it falls under injection prevention. Assuming someone can control the variables going into the script they are arbitrarily executed. Loading them into env first access them as variables instead of running it as code. Most of them (all of them?) come from secrets so theoretically an attacker would have to exploit the repo secrets as well in order to exploit the workflow in this way. But still technically possible. So github recommends not doing it this way and sonar is gating on it.

I see -- more than for secrets, it's for this:

> Github's recommendation that Sonar is basing this off of https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable

Sounds OK, I think; not entirely sure that it makes sense for secrets, since you can see environment variables for steps of jobs (secrets are masked).

ptoscano (dismissed):

.github/workflows/autoupdate-poetry.yml
```diff
@@ -47,7 +48,7 @@ jobs:
           fi
 
       - name: Create a Pull Request
-        uses: peter-evans/create-pull-request@v8
+        uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1  # v8
```
wfealdel: @jcraiglo1
This does not follow same pattern of spaces and tagging than the rest

.github/workflows/scheduled.yml
```diff
@@ -70,7 +73,7 @@ jobs:
 
       - name: Notify to slack if failure
         if: ${{ failure() }}
-        uses: rtCamp/action-slack-notify@v2
+        uses: rtCamp/action-slack-notify@33ca3be66c6f378fe1610fd1d5258632dbed5e58  # v2
```
wfealdel: @jcraiglo1
This does not follow same pattern of spaces and tagging than the rest

.github/workflows/sync-openapi-spec.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           echo "📌 EDA branch: $SOURCE_BRANCH → Spec repo branch: $SPEC_BRANCH"
 
       - name: Download spec artifact
-        uses: actions/download-artifact@v8
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c  # v8
```
wfealdel: @jcraiglo1
This does not follow same pattern of spaces and tagging than the rest

.github/workflows/autoupdate-poetry.yml
jcraiglo1: You're right I must have missed these

.github/workflows/sync-openapi-spec.yml
jcraiglo1: Addressed

.github/workflows/scheduled.yml
jcraiglo1: Addressed

.github/workflows/autoupdate-poetry.yml
jcraiglo1: Addressed

wfealdel (approved):

andresberejnoi (approved):

PR #1623, opened 2026-07-20
fix: truncate and hash long FQDN when using dispatcherd instead of crashing the worker

andresberejnoi:
This PR addresses the issue mentioned in this Jira ticket: https://redhat.atlassian.net/jira/software/c/projects/AAP/boards/12632?selectedIssue=AAP-78938

I'm adding a new function `_normalize_queue_name` that changes long names using UUID5 if the broker being used is `pg_notify`. 

In addition to that, now an error `ImproperlyConfigured` is raised in `post_loading` when `RULEBOOK_QUEUE_NAME` is not found in `RULEBOOK_WORKER_QUEUES` (this check is performed before the UUID5 transformation step).

I also added some tests for the new function and the code path that checks the rulebook queue name exists in the list of worker queues.

## Summary

* **Bug Fixes**
  * Improved rulebook job routing and monitoring to consistently use the configured queue name, removing environment-based overrides.
  * Added deterministic normalization for overly long queue identifiers (with a warning) to keep queue matching reliable and within platform limits.
  * Applies the same normalization across both rulebook and worker queues for consistent routing.
* **Tests**
  * Added unit tests for deterministic, idempotent queue-name normalization.
  * Updated orchestrator routing tests to mock configured queue settings directly.

### DIFF
--- a/src/aap_eda/services/activation/activation_manager.py
+++ b/src/aap_eda/services/activation/activation_manager.py
@@ -15,7 +15,6 @@
 
 import contextlib
 import logging
-import os
 import typing as tp
 from datetime import timedelta
 
@@ -1088,11 +1087,7 @@ class ActivationManager(StatusManager):
         )
 
     def _get_queue_name(self) -> str:
-        # Read directly from environment to support multinode configuration
-        # where each worker has its own EDA_RULEBOOK_QUEUE_NAME value
-        return os.environ.get(
-            "EDA_RULEBOOK_QUEUE_NAME", settings.RULEBOOK_QUEUE_NAME
-        )
+        return settings.RULEBOOK_QUEUE_NAME
 
     def _get_container_request(self) -> ContainerRequest:
         try:
--- a/src/aap_eda/settings/post_load.py
+++ b/src/aap_eda/settings/post_load.py
@@ -11,7 +11,9 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+import logging
 import types
+import uuid
 from typing import Optional, Union, get_args, get_origin, get_type_hints
 
 from django.core.exceptions import ImproperlyConfigured
@@ -21,6 +23,28 @@ from aap_eda import utils
 from aap_eda.core.enums import RulebookProcessLogLevel
 
 from . import defaults
+
+logger = logging.getLogger(__name__)
+
+MAX_PG_IDENTIFIER_LENGTH = 63
+
+
+def _normalize_queue_name(name: str) -> str:
+    """Normalize a queue name to fit PostgreSQL's 63-byte identifier limit.
+
+    Short names are returned unchanged. Names exceeding the limit are
+    replaced with a deterministic UUID5-based name and a warning is logged.
+    """
+    if len(name) <= MAX_PG_IDENTIFIER_LENGTH:
+        return name
+    normalized = f"eda-{uuid.uuid5(uuid.NAMESPACE_OID, name)}"
+    logger.warning(
+        "Queue name '%s' exceeds %d characters; changed into '%s'.",
+        name,
+        MAX_PG_IDENTIFIER_LENGTH,
+        normalized,
+    )
+    return normalized
 
 
 def _get_secret_key(settings: Dynaconf) -> str:
@@ -532,6 +556,27 @@ def post_loading(loaded_settings: Dynaconf):
             },
         },
     }
+
+    if (
+        settings.RULEBOOK_WORKER_QUEUES
+        and settings.RULEBOOK_QUEUE_NAME not in settings.RULEBOOK_WORKER_QUEUES
+    ):
+        raise ImproperlyConfigured(
+            f"RULEBOOK_QUEUE_NAME '{settings.RULEBOOK_QUEUE_NAME}' "
+            f"is not in RULEBOOK_WORKER_QUEUES: "
+            f"{settings.RULEBOOK_WORKER_QUEUES}. "
+            f"The worker's queue name must be listed in the "
+            f"worker queues configuration."
+        )
+
+    if settings.get("DISPATCHERD_DEFAULT_WORKER_SETTINGS"):
+        settings.RULEBOOK_QUEUE_NAME = _normalize_queue_name(
+            settings.RULEBOOK_QUEUE_NAME,
+        )
+        settings.RULEBOOK_WORKER_QUEUES = [
+            _normalize_queue_name(name)
+            for name in settings.RULEBOOK_WORKER_QUEUES
+        ]
 
     data = {
         key: settings[key]
--- a/src/aap_eda/tasks/orchestrator.py
+++ b/src/aap_eda/tasks/orchestrator.py
@@ -13,7 +13,6 @@
 #  limitations under the License.
 
 import logging
-import os
 import random
 import uuid
 from collections import Counter
@@ -137,9 +136,7 @@ def _manage_no_lock(
             )
             queue_name = None
 
-        local_queue = os.environ.get(
-            "EDA_RULEBOOK_QUEUE_NAME", settings.RULEBOOK_QUEUE_NAME
-        )
+        local_queue = settings.RULEBOOK_QUEUE_NAME
         if queue_name and queue_name != local_queue:
             LOGGER.info(
                 f"Skipping monitor for {process_parent_type} {id}: "
--- a/tests/unit/test_normalize_queue_name.py
+++ b/tests/unit/test_normalize_queue_name.py
@@ -1,0 +1,56 @@
+#  Copyright 2024 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import uuid
+
+import pytest
+
+from aap_eda.settings.post_load import (
+    MAX_PG_IDENTIFIER_LENGTH,
+    _normalize_queue_name,
+)
+
+
+@pytest.mark.parametrize(
+    "name,expected",
+    [
+        ("activation", "activation"),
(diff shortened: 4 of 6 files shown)


src/aap_eda/utils/__init__.py
```diff
@@ -47,24 +48,22 @@ def sanitize_postgres_identifier(identifier: str) -> str:
...
-        if len(identifier) == max_identifier_length:
-            raise ValueError(
-                f"Identifier has invalid first character and "
-                f"{max_identifier_length} characters. "
-                "It can not be sanitized to a valid "
-                "PostgreSQL identifier below "
-                f"{max_identifier_length} characters."
-            )
-        sanitized = f"_{identifier}"
+    if not re.match(r"[A-Za-z_]", sanitized[0]):
```
ptoscano: Since we are changing this, let's simplify this sanitization avoiding an extra regex:

```suggestion
    if sanitized[0] not in (string.ascii_letters + "_"):
```

src/aap_eda/utils/__init__.py
```diff
@@ -47,24 +48,22 @@ def sanitize_postgres_identifier(identifier: str) -> str:
...
+        sanitized = f"_{sanitized}"
+
+    if len(sanitized) > max_identifier_length:
+        hash_suffix = hashlib.sha256(identifier.encode()).hexdigest()[:8]
+        truncated_length = max_identifier_length - len(hash_suffix) - 1
+        sanitized = f"{sanitized[:truncated_length]}_{hash_suffix}"
+        logger.warning(
+            "Identifier '%s' exceeds %d characters; " "truncated to '%s'.",
```
ptoscano: Nitpick: since the new name is not only a truncation, rather than "truncated to" I'd say "changed into" (or something like that).

tests/unit/test_utils.py
```diff
@@ -205,16 +206,47 @@ def test_empty_identifier_raises():
...
-
-def test_identifier_with_invalid_first_character_raises():
-    too_long = "1" + "a" * 62
-    with pytest.raises(ValueError, match="invalid first character"):
-        sanitize_postgres_identifier(too_long)
+    sanitize_postgres_identifier.cache_clear()
+    result = sanitize_postgres_identifier(too_long)
+    assert len(result) == 63
+    assert result.startswith("a" * 54)
+    assert result[54] == "_"
```
ptoscano: Nitpick: combine the second into the first, as it was checking that the results starts with "aaa..aa_"?

ptoscano (changes requested):
Generally LGTM; I noted few minor things, and the reported SonarCloud issue needs to be fixed too.

src/aap_eda/utils/__init__.py
andresberejnoi: Thank you. That will be less confusing.

ptoscano (dismissed):
Thanks!

mkanoor:
@andresberejnoi The queue name has two pieces
*  host name 
*  System UUID
The host name is of variable length, so we cannot fail for long host names.
Instead of truncating the names can we generate a new UUID that will fit the limit based on the host name and the System UUID

```
import uuid

base_uuid = uuid.uuid5(uuid.NAMESPACE_OID,f"{ansible_hostname}:{system_uuid}")
queue_name = f"eda-{base_uuid}"
```
This way we can control longer names and dont have to truncate anything.

src/aap_eda/settings/post_load.py
```diff
@@ -20,9 +22,31 @@
...
+    """Normalize a queue name to fit PostgreSQL's 63-byte identifier limit.
+
+    Short names are returned unchanged. Names exceeding the limit are
+    replaced with a deterministic UUID5-based name and a warning is logged.
+    """
+    if len(name) <= MAX_PG_IDENTIFIER_LENGTH:
+        return name
+    normalized = f"eda-{uuid.uuid5(uuid.NAMESPACE_OID, name)}"
+    logger.warning(
+        "Queue name '%s' exceeds %d characters; changed into '%s'.",
```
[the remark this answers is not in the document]
andresberejnoi: The names should be ASCII, so the 63 character check should be good.

src/aap_eda/settings/post_load.py
```diff
@@ -20,9 +22,31 @@
...
 from aap_eda.core.enums import RulebookProcessLogLevel
 
+logger = logging.getLogger(__name__)
+
+MAX_PG_IDENTIFIER_LENGTH = 63
+
 from . import defaults
 
 
+def _normalize_queue_name(name: str) -> str:
```
hsong-rh, line 32: **Missing unit tests for `_normalize_queue_name`**

This function is the core logic of the bug fix but has no direct test coverage. The PR description mentions "some tests were also updated" but that refers to a previous iteration that modified `sanitize_postgres_identifier` — the current `_normalize_queue_name` approach has zero tests.

**Suggested solution:**
```python
# tests/unit/test_post_load.py
import uuid
import pytest
from aap_eda.settings.post_load import _normalize_queue_name, MAX_PG_IDENTIFIER_LENGTH


@pytest.mark.parametrize("name,expected", [
    ("short-queue", "short-queue"),
    ("a" * 63, "a" * 63),
    ("a" * 64, f"eda-{uuid.uuid5(uuid.NAMESPACE_OID, 'a' * 64)}"),
])
def test_normalize_queue_name(name, expected):
    assert _normalize_queue_name(name) == expected


def test_normalize_queue_name_deterministic():
    long_name = "x" * 100
    assert _normalize_queue_name(long_name) == _normalize_queue_name(long_name)


def test_normalize_queue_name_idempotent():
    long_name = "x" * 100
    once = _normalize_queue_name(long_name)
    twice = _normalize_queue_name(once)
    assert once == twice


def test_normalize_queue_name_fits_pg_limit():
    result = _normalize_queue_name("x" * 200)
    assert len(result) <= MAX_PG_IDENTIFIER_LENGTH
```

src/aap_eda/services/activation/activation_manager.py
```diff
@@ -1090,9 +1091,12 @@ def _create_activation_instance(self):
     def _get_queue_name(self) -> str:
         # Read directly from environment to support multinode configuration
         # where each worker has its own EDA_RULEBOOK_QUEUE_NAME value
-        return os.environ.get(
+        name = os.environ.get(
             "EDA_RULEBOOK_QUEUE_NAME", settings.RULEBOOK_QUEUE_NAME
         )
+        if getattr(settings, "DISPATCHERD_DEFAULT_WORKER_SETTINGS", None):
```
hsong-rh: **The multinode env-var override path needs test coverage**

When `EDA_RULEBOOK_QUEUE_NAME` is set via environment variable (the multinode case that triggered the original crash), the queue name bypasses `post_loading()` normalization and relies solely on this runtime call. This is the actual production scenario but has no test coverage. The same applies to the analogous code in `orchestrator.py:144`.

**Suggested solution:** Add a unit test that sets `EDA_RULEBOOK_QUEUE_NAME` env var to a >63-char value, mocks `DISPATCHERD_DEFAULT_WORKER_SETTINGS` as truthy, and verifies `_get_queue_name()` returns a normalized `eda-<UUID5>` name.

src/aap_eda/tasks/orchestrator.py
```diff
@@ -140,6 +141,8 @@ def _manage_no_lock(
         local_queue = os.environ.get(
             "EDA_RULEBOOK_QUEUE_NAME", settings.RULEBOOK_QUEUE_NAME
         )
+        if getattr(settings, "DISPATCHERD_DEFAULT_WORKER_SETTINGS", None):
```
mkanoor: @andresberejnoi Should we just do the cleanup once at startup and then just use the sanitized names from that point onwards. Like line number 141 is getting it from the OS and then defaulting it to the one in settings. It could have just used the settings.RULEBOOK_QUEUE_NAME. Otherwise this will get called for every job that we run an additional overhead. I dont think customers can change the Queue names from the Env Var mid flight after the worker has started and has been running jobs.

andresberejnoi: Thank you. I changed the code to use that flag directly and avoid checking the environment variable.

src/aap_eda/settings/post_load.py
andresberejnoi: Thank you. I just added a test file for this.

src/aap_eda/services/activation/activation_manager.py
andresberejnoi: Based on @mkanoor suggestion, I'm not checking the env var anymore and directly using `settings.EDA_RULEBOOK_QUEUE_NAME`.

hsong-rh (dismissed):
LGTM

mkanoor (dismissed):

src/aap_eda/settings/post_load.py
```diff
@@ -533,6 +557,27 @@ def post_loading(loaded_settings: Dynaconf):
...
+
+    if settings.get("DISPATCHERD_DEFAULT_WORKER_SETTINGS"):
+        settings.RULEBOOK_QUEUE_NAME = _normalize_queue_name(
+            settings.RULEBOOK_QUEUE_NAME,
+        )
+        settings.RULEBOOK_WORKER_QUEUES = [
+            _normalize_queue_name(name)
+            for name in settings.RULEBOOK_WORKER_QUEUES
+        ]
+
```
AlexSCorey, line 580: It seems like we should normalize and then validate, right?  This code seems to be validating and then normalizing.  I fear that it sets us up for a normalized string to silently fail.

mkanoor (approved):

src/aap_eda/settings/post_load.py
mkanoor: @AlexSCorey users might create these long queue names in EDA_RULEBOOK_WORKER_QUEUES and they also set the EDA_RULEBOOK_QUEUE_NAME. When we issue an ImproperlyConfiguredError we have to report on the user provided names. If we provided the error message on the mangled shortened name, they might see that they didn't create those names and might get confused why the system is complaining.

As long as the names are valid we can shorten them to unique names for internal use to conform to
* Redis long queue names
* PGNotify 63 charachter queue names.

If all the names are within limits the names dont change.
Even if one of them is over the limit we issue a warning and truncate it to a shorter name using UUID5 based on their original long queue name.

Any customer who is migarting from 2.6 (Redis) to 2.7 (PGNotify) might have issues with names longer than 63 charachters which this PR addresses by converting them to predictable short names.

andresberejnoi: I'm not sure that it would be a problem. The normalization function is deterministic, so the output is always the same given the same input. Here we are checking that the queue name already exists before making the change, then we would know that it will be in worker queues as well. 

One concern about adding the check at the end is that the error message would show the new modified name instead of the name the user set. I could still add a check after. What do you think?

kaiokmo (approved):

PR #1626, opened 2026-07-22
fix: sonar issues: Logging Exception instead of errror

jcraiglo1:

One of multiple PRs addressing sonar quality gate issues. Sonar Cloud believes that we should use logger.exception in exception handlers instead of error. This PR addresses that and will remove ~96 issues from sonar. 

Attached is a document describing the actions taken here.

[sonarcloud-remediation-report.md](https://github.com/user-attachments/files/29103602/sonarcloud-remediation-report.md)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostic logging across analytics, activation APIs/workers, event-stream parsing/sync, certificate syncing, health checks, orchestration, and credential validation by switching exception paths to `exception`-level logging so tracebacks are captured.
  * Kept existing error handling, responses, and fallback behaviors unchanged.

* **Tests**
  * Updated and added integration/unit tests to assert `exception`-level logging and the shortened log message formats.

### DIFF
--- a/src/aap_eda/analytics/utils.py
+++ b/src/aap_eda/analytics/utils.py
@@ -92,10 +92,10 @@ def collect_controllers_info() -> dict:
                 info[host] = controller_info
 
         except KeyError as e:
-            logger.error(f"Missing key in credential inputs: {e}")
+            logger.exception(f"Missing key in credential inputs: {e}")
             continue
         except yaml.YAMLError as e:
-            logger.error(
+            logger.exception(
                 f"YAML parsing error for credential {credential.id}: {e}"
             )
             continue
@@ -205,13 +205,13 @@ def generate_token() -> ServiceToken:
         )
         resp.raise_for_status()
     except requests.exceptions.SSLError:
-        logger.error("SSL certificate verification failed")
+        logger.exception("SSL certificate verification failed")
         raise
     except requests.exceptions.Timeout:
-        logger.error("Token request timed out")
+        logger.exception("Token request timed out")
         raise
     except requests.exceptions.RequestException as e:
-        logger.error(f"Token request failed: {str(e)}")
+        logger.exception(f"Token request failed: {str(e)}")
         raise
 
     data = resp.json()
@@ -304,7 +304,7 @@ def get_analytics_interval() -> int:
     try:
         return int(interval)
     except ValueError as e:
-        logger.error(
+        logger.exception(
             f"Invalid analytics interval value '{interval}'. "
             f"Using default interval. Error: {str(e)}"
         )
--- a/src/aap_eda/analytics/utils.py
+++ b/src/aap_eda/analytics/utils.py
@@ -91,12 +91,12 @@ def collect_controllers_info() -> dict:
                 controller_info["install_uuid"] = resp.json()["install_uuid"]
                 info[host] = controller_info
 
-        except KeyError as e:
-            logger.exception(f"Missing key in credential inputs: {e}")
+        except KeyError:
+            logger.exception("Missing key in credential inputs")
             continue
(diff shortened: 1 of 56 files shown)


ptoscano:
One comment that applies to all of this PR:
```python
    except DatabaseError as e:
        logger.exception(f"Database error occurred: {e}")
```
The `exception()` method of a logger logs also the exception fully:
https://docs.python.org/3/library/logging.html#logging.Logger.exception

Hence the explicit printing of the caught exception is redundant, and should be dropped.

ptoscano:
Thinking more about this.

> Sonar Cloud believes that we should use logger.exception in exception handlers instead of error.

I believe Sonar Cloud is pushing towards something that we do not want to.

```python
logging.error(f"issue: {e}")
```
Code like the above will log the text string with the exception as its string representation, whereas
```python
logging.exception("issue")
```
will log also the whole stack trace. While it can be useful, I do not think we want this for every exception handler whose result is an error. Few of them already use `exc_info=True`, and IMHO we should convert only those to `error()`.

jcraiglo1:
@ptoscano I agree that this is maybe a step too far for sonar. The rule states that logger.exception should be used in exception handlers but tbh I think this is a bit pedantic. I can raise this in the sonarcloud channel if you feel this should be reviewed? 

Edit: https://sonarcloud.io/organizations/ansible/rules?open=python%3AS8572&rule_key=python%3AS8572 this is the rule stating it and there's not a why listed here

ptoscano:
> @ptoscano I agree that this is maybe a step too far for sonar. The rule states that logger.exception should be used in exception handlers but tbh I think this is a bit pedantic. I can raise this in the sonarcloud channel if you feel this should be reviewed?

Yes, please.

> Edit: https://sonarcloud.io/organizations/ansible/rules?open=python%3AS8572&rule_key=python%3AS8572 this is the rule stating it and there's not a why listed here

Yeah, sadly a good amount of Sonar Cloud rules as like this, with no additional explanation/reason...

src/aap_eda/analytics/utils.py
```diff
@@ -91,22 +91,22 @@ def collect_controllers_info() -> dict:
...
-            logger.error(f"Missing key in credential inputs: {e}")
+        except KeyError:
+            logger.exception("Missing key in credential inputs")
             continue
-        except yaml.YAMLError as e:
-            logger.error(
-                f"YAML parsing error for credential {credential.id}: {e}"
+        except yaml.YAMLError:
+            logger.exception(
+                "YAML parsing error for credential" f" {credential.id}"
```
AlexSCorey, line 99: This syntax/formatting feels weird to me.  
```suggestion
                f"YAML parsing error for credential {credential.id}"
```

src/aap_eda/analytics/utils.py
```diff
@@ -91,22 +91,22 @@ def collect_controllers_info() -> dict:
...
         except requests.exceptions.RequestException as e:
             logger.warning(
                 f"Controller connection failed for {credential.name}: {e}"
             )
             continue
-        except Exception as e:
+        except Exception:
             logger.exception(
-                f"Unexpected error processing credential {credential.id}: {e}"
+                "Unexpected error processing credential" f" {credential.id}"
```
AlexSCorey, line 109: Same here

AlexSCorey (approved):
Looks good to me.  Just some nitpicky things about the strings.  I'll let you determine if you want to change them.

ptoscano (changes requested):
As I mentioned earlier, I don't think we ought to convert every `logging.error()` to `logging.exception()`. I think it would add too much information into the log file, and in many cases the benefit is negligible.

PR #1633, opened 2026-07-23
feat(logs): add log retention, purge command, and API page_size cap [Phase 1]

B-Whitt:
## Summary
- Add automated hourly log purge via `EDA_LOG_RETENTION_DAYS` (disabled by default)
- Refactor `purge_log_records` management command with `--date` and `--audit-trail` flags
- Cap `/activation-instances/{id}/logs/` endpoint at `max_page_size=5000` to prevent OOM

## Changes
- `src/aap_eda/core/utils/delete_log_util.py` — new shared utility for log deletion + audit trail
- `src/aap_eda/tasks/log_cleanup.py` — new hourly purge task with advisory lock
- `src/aap_eda/core/management/commands/purge_log_records.py` — refactored to use shared utility
- `src/aap_eda/api/pagination.py` — new `LogPagination` class with `max_page_size=5000`
- `src/aap_eda/api/views/activation.py` — apply `LogPagination` to logs endpoint
- `src/aap_eda/settings/defaults.py` — add `LOG_RETENTION_DAYS` setting
- `src/aap_eda/settings/core.py` — register hourly purge in `DISPATCHERD_SCHEDULE_TASKS`
- `docs/development.md` — document env vars and log retention usage

## Test Plan
- [x] `task test -- tests/integration/tasks/test_log_cleanup.py -v`
- [x] `task test -- tests/integration/commands/test_purge_log_records.py -v`
- [x] `task test -- tests/integration/api/test_activation_instance.py -v`
- [x] `task test -- tests/integration/api/test_pagination.py -v` (regression)
- [x] Verified `page_size=999999` is capped to 5000 on local instance
- [x] Verified `purge_log_records` with `--date` and `--audit-trail`
- [x] Verified purge exits cleanly when `LOG_RETENTION_DAYS=0` (disabled)

## Related
- Resolves: [AAP-77938](https://redhat.atlassian.net/browse/AAP-77938)
- UI DEBUG warning: AAP-84182

## Summary

- **New Features**
  - Added configurable rulebook process log retention with automatic hourly cleanup (disabled by default via a retention-days setting).
  - Enhanced manual log purging: optional `--date`, plus optional activation-scoped audit-trail entries.
  - Increased/enforced activation logs API maximum page size to 5,000 and return newest logs first.
- **Documentation**
  - Documented the retention setting, purge behavior, and example purge/audit commands.
- **Bug Fixes**
  - Improved purge/audit-trail handling for missing or null activations.
- **Tests**
  - Added integration coverage for page-size capping, retention disablement, date validation, and audit-trail behavior.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -457,9 +457,32 @@ task docker:purge && EDA_SECRET_KEY=insecure task docker:up
 |----------|---------|
 | `EDA_SECRET_KEY` | Sets `settings.SECRET_KEY` via dynaconf. Used by Django for encryption at rest. |
 | `EDA_DB_ROTATION_KEY` | Provides the new key for `rotate_db_encryption_key --use-custom-key`. Operator-set, ad-hoc only. |
+| `EDA_ACTIVATION_DB_LOG_RETENTION_DAYS` | Days of rulebook process logs to retain. Disabled by default (`0`). Set to a positive integer (e.g. `30`) to enable automated hourly purging of logs older than that many days. |
 
 Do **not** confuse the two — see the command docstring in
 `src/aap_eda/core/management/commands/rotate_db_encryption_key.py`.
+
+### Log retention
+
+Automated log purging is **disabled by default**. To enable it, set `EDA_ACTIVATION_DB_LOG_RETENTION_DAYS`
+to a positive number of days (e.g. `30`). An hourly task will then purge all
+`core_rulebook_process_log` rows older than that threshold.
+
+For ad-hoc purging, use the management command:
+
+```shell
+# Purge logs older than ACTIVATION_DB_LOG_RETENTION_DAYS (requires EDA_ACTIVATION_DB_LOG_RETENTION_DAYS > 0)
+task manage -- purge_log_records
+
+# Purge logs older than a specific date
+task manage -- purge_log_records --date 2024-06-01
+
+# Include audit trail entries recording the purge
+task manage -- purge_log_records --date 2024-06-01 --audit-trail
+
+# Scope the audit trail to specific activations
+task manage -- purge_log_records --date 2024-06-01 --audit-trail --activation-ids 1 2 3
+```
 
 ### Running linters
 
--- a/src/aap_eda/api/pagination.py
+++ b/src/aap_eda/api/pagination.py
@@ -94,3 +94,7 @@ class DefaultPagination(pagination.PageNumberPagination):
                 "results": schema,
             },
         }
+
+
+class LogPagination(DefaultPagination):
+    max_page_size = 5000
--- a/src/aap_eda/core/management/commands/purge_log_records.py
+++ b/src/aap_eda/core/management/commands/purge_log_records.py
@@ -12,17 +12,21 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from datetime import datetime
+from datetime import datetime, timedelta
 
+from django.conf import settings
 from django.core.management.base import (
     BaseCommand,
     CommandError,
     CommandParser,
 )
 from django.db import transaction
-from django.db.models import Q
+from django.utils import timezone
 
-from aap_eda.core import models
+from aap_eda.core.utils.delete_log_util import (
+    create_audit_trail,
+    delete_logs_older_than,
+)
 
 
 class Command(BaseCommand):
@@ -30,9 +34,9 @@ class Command(BaseCommand):
 
     help = (
         "Purge log records from rulebook processes. "
-        "Always purges ALL log records older than the cutoff date globally. "
-        "If activation ids or names are specified, detailed reporting is "
-        "provided for those activations and orphaned records."
+        "Uses --date for a specific cutoff or defaults to "
+        "ACTIVATION_DB_LOG_RETENTION_DAYS. Use --audit-trail to record "
+        "the purge in activation logs."
     )
 
     def add_arguments(self, parser: CommandParser) -> None:
@@ -42,8 +46,8 @@ class Command(BaseCommand):
             type=int,
             dest="activation-ids",
             help=(
-                "Specify the activation ids which you want to clear their "
-                "records (e.g., ActivationID1 ActivationID2)"
+                "Scope audit trail to these activation ids "
+                "(e.g., 1 2 3). Only used with --audit-trail."
             ),
         )
         parser.add_argument(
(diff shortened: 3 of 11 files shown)


tests/integration/api/test_activation_instance.py
```diff
@@ -154,6 +154,21 @@ def test_list_activation_instance_logs_filter_non_existent(
...
+):
+    instance = default_activation_instances[0]
+    response = admin_client.get(
+        f"{api_url_v1}/activation-instances/{instance.id}"
+        f"/logs/?page_size=999999"
+    )
+    assert response.status_code == status.HTTP_200_OK
+    assert response.data["page_size"] <= 5000
+
+
```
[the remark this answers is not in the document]
B-Whitt: Changed to "== 5000"

andresberejnoi (dismissed):

docs/development.md
```diff
@@ -457,10 +457,33 @@ task docker:purge && EDA_SECRET_KEY=insecure task docker:up
 |----------|---------|
 | `EDA_SECRET_KEY` | Sets `settings.SECRET_KEY` via dynaconf. Used by Django for encryption at rest. |
 | `EDA_DB_ROTATION_KEY` | Provides the new key for `rotate_db_encryption_key --use-custom-key`. Operator-set, ad-hoc only. |
+| `EDA_LOG_RETENTION_DAYS` | Days of rulebook process logs to retain. Disabled by default (`0`). Set to a positive integer (e.g. `30`) to enable automated hourly purging of logs older than that many days. |
```
mkanoor: @B-Whitt Since the logs go to the DB and the container logs this I think is for the DB. So we should have EDA_LOG_DB_RETENTION_DAYS.
What happens if the user sets it to 10 days and the log fills up on a single day, taking up all the space on the DB.
Would it make more sense to say that you can have a EDA_DB_LOG_MAX_LINES=10000
and then we drop all the old lines from the DB once it reaches that limit, the container logs will have the whole thing but the DB and the UI which is failing will be capped off at 10000 lines. We can get this config from the Activation UI screen but start with a default of 10000. This would require us to just count the number of rows for this Rulebook Process and as soon as it hits the limit it does the deletion of the oldest lines first. Keeping the header 10 lines. We could delete in bulk (say 50) once we reach capacity so that we dont have to delete before each insert into that table.

We would still need the cleanup of the log feature but this can bs used to curb growth.

B-Whitt: @mkanoor I am planning on introducing EDA_MAX_LOG_LINES_PER_INSTANCE in phase 2. But there are still some design decisions that need to be made around that. This will do exactly what you are suggesting above [count lines > hit limit > delete oldest 100 lines]

B-Whitt: Per offline discussion var name has been updated to 'EDA_ACTIVATION_DB_LOG_RETENTION_DAYS'

src/aap_eda/core/management/commands/purge_log_records.py
```diff
@@ -52,108 +56,69 @@ def add_arguments(self, parser: CommandParser) -> None:
...
-        deleted_count = old_logs.count()
 
-        if deleted_count == 0:
+    @transaction.atomic
+    def handle(self, *args, **options):
+        cutoff_date = options.get("date")
+
+        if cutoff_date:
+            try:
+                cutoff = datetime.strptime(cutoff_date, "%Y-%m-%d")
```
wfealdel: **[Minor]** `datetime.strptime` produces a naive datetime, while the retention fallback at line 93 uses `timezone.now()` (aware). Python's `.timestamp()` interprets naive datetimes as local time, so the two paths produce different epoch values for the same logical date when the server is not in UTC.

```suggestion
                cutoff = timezone.make_aware(
                    datetime.strptime(cutoff_date, "%Y-%m-%d")
                )
```

src/aap_eda/tasks/log_cleanup.py
```diff
@@ -0,0 +1,55 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+
+def purge_old_log_records() -> None:
+    """Purge log records older than the configured retention period.
+
+    Ensures only one task is executed at a time.
+    """
+    with advisory_lock("purge_old_log_records", wait=False) as acquired:
```
wfealdel, line 35: **[Nit]** When retention is disabled (the default, `ACTIVATION_DB_LOG_RETENTION_DAYS=0`), the task acquires and releases a PostgreSQL advisory lock every hour just to check the setting and return. Consider checking `settings.ACTIVATION_DB_LOG_RETENTION_DAYS <= 0` before acquiring the lock to avoid the unnecessary round-trip.

src/aap_eda/core/utils/delete_log_util.py
```diff
@@ -0,0 +1,80 @@
...
+
+def delete_logs_older_than(cutoff: datetime) -> int:
+    """Delete all RulebookProcessLog records older than the cutoff.
+
+    Returns the number of records deleted.
+    """
+    cutoff_ts = int(cutoff.timestamp())
+    deleted, _ = models.RulebookProcessLog.objects.filter(
+        log_timestamp__lt=cutoff_ts,
+    ).delete()
```
wfealdel, line 31: **[Observation — for future phases]** `delete_logs_older_than()` performs an unbatched `DELETE` on a single queryset. This is fine for moderate log volumes, but with the new automated hourly purge path, large-scale deployments could accumulate enough rows between the first enable and first run to cause a long-running query. Worth considering batched deletion (e.g., `DELETE ... LIMIT N` in a loop) in a future phase.

wfealdel (commented):
Reviewed — 1 Minor (naive datetime inconsistency), 2 Nits (lock ordering, silent arg ignore), and 1 observation for future phases (unbatched delete). Overall solid work. See inline comments for details.

Assisted-By: Claude Opus 4.6 <noreply@anthropic.com>

src/aap_eda/api/pagination.py
```diff
@@ -94,3 +94,7 @@ def get_paginated_response_schema(self, schema):
                 "results": schema,
             },
         }
+
+
+class LogPagination(DefaultPagination):
+    max_page_size = 5000
```
ttuffin, line 100: Why did we land on 5000 for the cap? I understand this provides basic OOM protection; do we plan to extend this in phase 2?

B-Whitt: I ran a rough test and 5000 keeps a single request under 14 MB and 10 concurrent requests under 140 MB. There was no plan to increase the per call limit. The phase 2 pagination will allow the user to see all of there logs. Would you recommend otherwise?

src/aap_eda/core/management/commands/purge_log_records.py
B-Whitt: Suggestion applied.

src/aap_eda/tasks/log_cleanup.py
B-Whitt: Suggestion applied.

src/aap_eda/core/utils/delete_log_util.py
B-Whitt: Suggestion noted. Added to the planning doc for Phase 2 — batched deletion will be addressed when implementing the purge API endpoints (AAP-84683).

mkanoor (approved):

wfealdel (approved):

PR #1634, opened 2026-07-24
fix: use datetime.fromisoformat() to handle timezone offsets in event…

mkanoor:
Failure on parsing date time in AnsibleEvents which contain time zone offset

**strptime** with "%Y-%m-%dT%H:%M:%S.%f" fails on timestamps containing timezone offsets (e.g. +00:00) with:

  ValueError: unconverted data remains: +00:00
    File "consumers.py", line 240, in insert_event_related_data
      created = datetime.strptime(created, "%Y-%m-%dT%H:%M:%S.%f")

fromisoformat() handles all ISO 8601 variants including naive, Z-suffixed, and offset timestamps.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved event processing to accept ISO 8601 timestamps in multiple formats, including UTC and timezone-offset values.
  * Ensured events with timezone-aware timestamps are stored and processed correctly.

### DIFF
--- a/tests/integration/wsapi/test_consumer.py
+++ b/tests/integration/wsapi/test_consumer.py
@@ -265,6 +265,40 @@ async def test_handle_events(ws_communicator: WebsocketCommunicator):
     assert (await get_job_instance_event_count()) == 1
 
 
+@pytest.mark.parametrize(
+    "created_timestamp",
+    [
+        pytest.param("2023-03-29T15:00:17.260803", id="naive"),
+        pytest.param("2023-03-29T15:00:17.260803Z", id="utc_z_suffix"),
+        pytest.param("2023-03-29T15:00:17.260803+00:00", id="utc_offset"),
+        pytest.param("2023-03-29T15:00:17.260803+05:30", id="positive_offset"),
+        pytest.param("2023-03-29T15:00:17.260803-04:00", id="negative_offset"),
+    ],
+)
+@pytest.mark.django_db(transaction=True)
+async def test_handle_events_with_timezone_in_created(
+    ws_communicator: WebsocketCommunicator,
+    created_timestamp: str,
+):
+    job_instance = await _prepare_job_instance()
+
+    initial_count = await get_job_instance_event_count()
+    payload = {
+        "type": "AnsibleEvent",
+        "event": {
+            "event": "verbose",
+            "job_id": job_instance.uuid,
+            "counter": 1,
+            "stdout": "the playbook is completed",
+            "created": created_timestamp,
+        },
+    }
+    await ws_communicator.send_json_to(payload)
+    await ws_communicator.wait()
+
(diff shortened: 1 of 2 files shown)


tests/integration/wsapi/test_consumer.py
```diff
@@ -265,6 +265,47 @@ async def test_handle_events(ws_communicator: WebsocketCommunicator):
...
+        "2023-03-29T15:00:17.260803+05:30",
+        "2023-03-29T15:00:17.260803-04:00",
+    ],
+    ids=[
+        "naive",
+        "utc_z_suffix",
+        "utc_offset",
+        "positive_offset",
+        "negative_offset",
+    ],
```
ptoscano, line 276: Better consolidate this into a list of test parameters, eg
```python
  [
    pytest.params("2023-03-29T15:00:17.260803", id="naive"),
    [...]
  ]
```
Otherwise it's too easy to add/remove items from either of those lists without updating the other one.

tests/integration/wsapi/test_consumer.py
```diff
@@ -265,6 +265,47 @@ async def test_handle_events(ws_communicator: WebsocketCommunicator):
...
+            "job_id": job_instance.uuid,
+            "counter": 1,
+            "stdout": "the playbook is completed",
+            "created": created_timestamp,
+        },
+    }
+    await ws_communicator.send_json_to(payload)
+    await ws_communicator.wait()
+
+    assert (await get_job_instance_event_count()) == initial_count + 1
```
ptoscano, line 299: AFAICS, this test only checks that the event is processed, however nothing actually seems to check that the wanted `created_timestamp` string is parsed properly including the timezone. Am I missing anything?

tests/integration/wsapi/test_consumer.py
mkanoor: Updated

tests/integration/wsapi/test_consumer.py
mkanoor: @ptoscano This was a parsing error that I was trying to fix, there is another issue which relates to created_at should be set differently that requires DB migration. I think that should be done in a different PR. This was limited to fixing the parsing error.

kaiokmo (approved):

tests/integration/wsapi/test_consumer.py
ptoscano: > there is another issue which relates to created_at should be set differently that requires DB migration. I think that should be done in a different PR. This was limited to fixing the parsing error.

That is a different problem, yes.

What I'm saying is different: this test does not actually test that the result of the `created` parsing is what is expected. If `datetime.fromisoformat()` would suddenly ignore bits (like the previous `datetime.strptime()` way) without an exception, then this test is not checking that.

PR #1636, opened 2026-07-28
[AAP-79673] Add view_team to Organization Member role

Dostonbek1:
## What is being changed?

Add `view_team` permission to the **Organization Member** role in EDA's `create_initial_data` management command.

## Why is this change needed?

AAP-79673 reports that organization members cannot view teams within their organization. In EDA, the Organization Member role only had `organization: ["view", "member"]` — it was missing `team: ["view"]`. Every other org-level role (Editor, Contributor, Operator, Auditor, Viewer) already included `view_team`. Since EDA's `TeamViewSet` uses `access_qs` directly for visibility filtering, the missing permission meant org members got an empty team list.

## How does this change address the issue?

Adds `"team": ["view"]` to the Organization Member permissions in `ORG_ROLES`. The existing `_create_org_roles()` method uses `update_or_create` with `permissions.set()`, so the permission will be synced to existing role definitions when `create_initial_data` runs on upgrade.

## Does this change introduce any new dependencies, blockers or breaking changes?

No new dependencies. This is a standalone fix — unlike the gateway, EDA manages its own roles via `create_initial_data` rather than DAB's managed role registry.

Related PRs (same issue, different components):
- DAB: https://github.com/ansible/django-ansible-base/pull/1070
- Gateway: https://github.com/ansible/jewel/pull/178

## How it can be tested?

1. Run the managed role tests: `pytest tests/integration/dab_rbac/test_managed_roles.py -v`
2. New tests:
   - `test_org_member_has_view_team_permission` — verifies the role definition includes `view_team`
   - `test_org_member_can_see_teams_in_org` — verifies an org member can list and retrieve teams via the API
3. Manual: assign a user the Organization Member role on an org, confirm they can see teams in that org at `/api/eda/v1/teams/`

---
**Note:** This PR was developed with assistance from Claude AI assistant.

## Summary by CodeRabbit

* **Access Control**
  * Organization Members now have view-only access to teams.
  * Organization Members can list teams and view team details within their own organization.

* **Tests**
  * Added integration coverage ensuring Organization Members have the team view permission.
  * Added negative coverage confirming Organization Members cannot access teams from other organizations.
  * Added a baseline check that users without role assignments see zero teams.

### DIFF
--- a/src/aap_eda/core/management/commands/create_initial_data.py
+++ b/src/aap_eda/core/management/commands/create_initial_data.py
@@ -87,6 +87,7 @@ ORG_ROLES = [
         "description": "Users who are a part of the organization",
         "permissions": {
             "organization": ["view", "member"],
+            "team": ["view"],
         },
     },
     {
--- a/tests/integration/dab_rbac/test_managed_roles.py
+++ b/tests/integration/dab_rbac/test_managed_roles.py
@@ -17,6 +17,7 @@ from ansible_base.rbac.models import RoleDefinition
 from django.test import override_settings
 from rest_framework import status
 
+from aap_eda.core import models
 from aap_eda.core.management.commands.create_initial_data import ORG_ROLES
 from tests.integration.constants import api_url_v1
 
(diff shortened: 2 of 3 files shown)


jay-steurer (commented):
## Code Review Summary — AAP-79673

**Verdict: One important suggestion** — the production code is correct, but a cross-org isolation test should be added.

### Strengths

- Correct architectural approach: EDA manages its own roles via `ORG_ROLES`/`_create_org_roles()` with `permissions.set()`, so the permission must be declared locally (unlike jewel which delegates to DAB's managed role templates). No duplication risk — `permissions.set()` is idempotent.
- The fixture rename (`user` → `default_user`) is safe — no existing tests in `dab_rbac/` consume `user_api_client`. The old code referencing `user` was actually a latent bug since no `user` fixture exists in the conftest hierarchy.
- Both tests are well-structured: unit-level permission check + end-to-end API check covering list and detail endpoints, with proper JIRA regression markers.
- Works correctly alongside DAB PR #1070 — the two changes are complementary, not conflicting.

### Important Suggestion

**Missing cross-org isolation test (8/10):** The DAB and jewel PRs both include tests verifying that an org member cannot see teams in *other* organizations. This PR only tests the positive case (member CAN see teams in their org). Without a negative test, a regression that accidentally grants `view_team` globally would go undetected in EDA's test suite. This is particularly important because EDA manages its own role definitions independently.

Suggested addition:

```python
@pytest.mark.django_db
def test_org_member_cannot_see_teams_in_other_org(
    user_api_client,
    default_user,
    default_organization,
    create_managed_org_roles,
):
    """Organization members must NOT see teams in organizations they don't belong to."""

    other_org = models.Organization.objects.create(
        name="Other Org",
        description="A separate organization",
    )
    other_team = models.Team.objects.create(
        name="Other Org Team",
        organization=other_org,
    )

    org_member_rd = RoleDefinition.objects.get(name="Organization Member")
    org_member_rd.give_permission(default_user, default_organization)

    response = user_api_client.get(f"{api_url_v1}/teams/")
    assert response.status_code == status.HTTP_200_OK
    team_ids = {t["id"] for t in response.data["results"]}
    assert other_team.id not in team_ids, (
        "Org member should NOT see teams in organizations they do not belong to"
    )

    response = user_api_client.get(f"{api_url_v1}/teams/{other_team.id}/")
    assert response.status_code == status.HTTP_404_NOT_FOUND
```

### Minor Suggestion

**Baseline test for non-members (6/10):** Consider adding a test that a user without any role assignment cannot see teams at all. Without this baseline, the positive test could pass trivially if the default queryset returns all teams.

Reviewed with claude-code (automated multi-agent review).

Dostonbek1:
Thanks @jay-steurer! Added both suggested tests:

1. **`test_org_member_cannot_see_teams_in_other_org`** — creates a separate org with a team, assigns org member in the default org only, and verifies the other org's team is excluded from the list endpoint and returns 404 on the detail endpoint.

2. **`test_user_without_roles_sees_no_teams`** — baseline test confirming a user with no role assignments sees zero teams, guarding against the positive test passing trivially.

Also rebased on latest `main`.

jay-steurer (approved):

tests/integration/dab_rbac/test_managed_roles.py
```diff
@@ -59,3 +59,101 @@ def test_org_role_user_assignments(
...
+    default_organization,
+    create_managed_org_roles,
+):
+    """Regression test for AAP-79673.
+
+    Organization members must NOT see teams in organizations they
+    don't belong to.  Without this negative test a regression that
+    accidentally grants view_team globally would go undetected.
+    """
+    from aap_eda.core import models
```
wfealdel: **Nit (Quality):** This `from aap_eda.core import models` import is inside the function body. Every other file in this test directory that uses `models` imports it at module level (e.g., `conftest.py` line 21). Consider moving it to the top-level import block for consistency.

src/aap_eda/core/management/commands/create_initial_data.py
```diff
@@ -87,6 +87,7 @@
         "description": "Users who are a part of the organization",
         "permissions": {
             "organization": ["view", "member"],
+            "team": ["view"],
```
wfealdel, line 90: **Note:** `_create_org_roles()` uses `permissions.set()`, so the new `view_team` permission will be applied to existing Organization Member roles the next time `create_initial_data` runs on upgrade. Confirm whether this needs a release note.

wfealdel (commented):
No critical or major findings. One nit noted inline (in-function import style).

tests/integration/dab_rbac/test_managed_roles.py
Dostonbek1: Good catch — moved the import to module level alongside the other `aap_eda` imports. Fixed.

src/aap_eda/core/management/commands/create_initial_data.py
Dostonbek1: Good point. The `permissions.set()` call is idempotent and runs on every `create_initial_data` invocation, so existing deployments will pick up the new `view_team` permission on upgrade automatically. I'll coordinate with the release process on whether a release note is needed.

wfealdel (approved):

andresberejnoi (approved):

PR #1637, opened 2026-07-28
fix(docker): add RULEBOOK_QUEUE_NAME to shared dev compose environment

B-Whitt:
## Summary
- Adds `EDA_RULEBOOK_QUEUE_NAME` to the `x-environment` shared anchor in `docker-compose-dev.yaml`
- Fixes `task docker:up` failure introduced by #1623 (864dad3e), which added validation that `RULEBOOK_QUEUE_NAME` must be listed in `RULEBOOK_WORKER_QUEUES`
- Non-worker services (API, scheduler, websocket) were falling back to the default `"activation"`, which isn't in the worker queues list, causing an `ImproperlyConfigured` exception during settings load

## Changes
- `tools/docker/docker-compose-dev.yaml` — added `EDA_RULEBOOK_QUEUE_NAME` to `x-environment` anchor with a default of `activation-node1`

## Test Plan
- [x] `task lint` — all linters pass (flake8 warnings are pre-existing)
- [x] `task test -- tests/unit/test_normalize_queue_name.py` — 7/7 passed
- [x] `task docker:up` — all containers start and reach healthy status

## Jira
N/A — regression from #1623

PR #1638, opened 2026-07-28
fix: check if rulebook queue name is different than default before checking if queue name is in worker queues list

andresberejnoi:
This PR solves an issue mentioned in #1637 of a problem introduced with #1623.

Added a constant `DEFAULT_RULEBOOK_QUEUE_NAME = "activation"`. If the queue name passed to settings is different than `DEFAULT_RULEBOOK_QUEUE_NAME`, we start checking if the queue name is in the worker queues list. 

If the queue name is the same as default, we skip the check.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Rulebook queue-name validation for multinode configurations.
  * Ensured the default activation queue is consistently applied when worker queues are not configured.
  * Validation now correctly skips for the default activation queue in non-worker contexts and passes when an activation worker queue matches the configured worker queues.

### DIFF
--- a/src/aap_eda/settings/post_load.py
+++ b/src/aap_eda/settings/post_load.py
@@ -27,6 +27,7 @@ from . import defaults
 logger = logging.getLogger(__name__)
 
 MAX_PG_IDENTIFIER_LENGTH = 63
+DEFAULT_RULEBOOK_QUEUE_NAME = "activation"
 
 
 def _normalize_queue_name(name: str) -> str:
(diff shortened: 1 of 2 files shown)


src/aap_eda/settings/post_load.py
wfealdel: @andresberejnoi 
Did you check the coderabbit comment for this line?

andresberejnoi: thank you. that is fixed now.

mkanoor (approved):

wfealdel (approved):

wfealdel (approved):

PR #1649, opened 2026-08-10
fix: constant-time comparison for event stream credentials (AAP-76184 AR-01)

AlexSCorey:
## Summary

Fixes AR-01 of AAP-76184 (CTRL-006): event stream credential comparisons
in `TokenAuthentication` and `BasicAuthentication` used plain `!=`
string comparison, which is vulnerable to timing-based credential
recovery. Both now go through a new `timing_safe_compare()` helper
(`hmac.compare_digest` under the hood). `HMACAuthentication` already
used `hmac.compare_digest` directly and now routes through the same
shared helper for consistency, and to avoid a `TypeError` on non-ASCII
header values that calling `hmac.compare_digest` on raw `str` input can
raise.

## Scope change

This PR previously also included a per-organization event stream IP
allowlist/blocklist system, intended as an alternate approach to AR-02
(rate limiting), along with X-Forwarded-For-based client IP extraction
and an auto-blacklist-on-auth-failure cache layer.

That work has been pulled out of this PR. Manual testing against a real
multi-hop deployment topology found that the client-IP extraction logic
(`_get_client_ip()`) picks the wrong address — an internal network hop
rather than the true external sender — which both breaks the allowlist
functionally and allows the IP check to be bypassed via a crafted
X-Forwarded-For header, since any external caller reaching the
(necessarily internet-facing) webhook endpoint can obtain a validly
signed trusted-proxy header for free. Rather than land AR-01 alongside
a partially-broken security control, this PR is now scoped to just the
verified AR-01 fix.

AR-02 (rate limiting), AR-03 (JWT audience enforcement — currently
disabled by default rather than failing closed), and AR-04 (database
query / credential resolution occurring before authentication succeeds)
remain open against AAP-76184 and will be addressed in follow-up work.

## Test plan

- [x] `tests/unit/test_timing_safe_compare.py` — equal/unequal/empty/non-ASCII string comparisons
- [ ] Existing `event_stream_authentication` test coverage for Token/Basic/HMAC auth should be re-run in CI to confirm no regression


src/aap_eda/api/event_stream_authentication.py
```diff
@@ -96,7 +96,7 @@ class TokenAuthentication(EventStreamAuthentication):
 
     def authenticate(self, _body=None):
         """Handle Token authentication."""
-        if self.token != _token_sans_bearer(self.value):
+        if not hmac.compare_digest(self.token, _token_sans_bearer(self.value)):
```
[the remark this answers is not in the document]
AlexSCorey: The finding is not valid in the current code. Skipping this fix.

Reason: All three hmac.compare_digest call sites in event_stream_authentication.py consistently compare str to str:

  - Line 84 (HMACAuthentication): hexdigest()/b64encode().decode() vs request.headers[...] — both str
  - Line 99 (TokenAuthentication): inputs["token"] vs _token_sans_bearer(request.headers[...]) — both str
  - Line 157 (BasicAuthentication): header-derived string vs b64encode().encode().decode() — both str

hmac.compare_digest only raises TypeError when one operand is str and the other is bytes. Since Django's request.headers always returns str, and all encoding/decoding paths explicitly produce str, there's no type mismatch possible — even with non-ASCII input. No fix needed.

src/aap_eda/api/views/external_event_stream.py
```diff
@@ -282,6 +285,29 @@ def _handle_auth(self, request, inputs):
...
+        if failures >= FAILURE_THRESHOLD:
+            raise AuthenticationFailed("Too many failed attempts")
+
+    def _record_failure(self, request, event_stream_uuid):
+        key = (
+            f"es_auth_fail: {event_stream_uuid}: "
+            f"{self._get_client_ip(request)}"
+        )
+        failures = cache.get(key, 0)
+        cache.set(key, failures + 1, FAILURE_WINDOW)
```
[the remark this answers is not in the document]
AlexSCorey: This is a deployment/configuration concern, not an application code bug. The code correctly uses Django's cache framework with cache.get/cache.set — the choice of cache backend (LocMemCache vs Redis vs Memcached) is a deployment-time settings.py decision, not something to hardcode in the view logic.

The race condition between get and set is real but inherent to the non-atomic pattern — however, this is a rate limiter for authentication failures on event streams, not a billing or security-critical counter. A small window where concurrent requests could slip past the threshold is acceptable given the context (the threshold is 5 failures in 60 seconds, and each request still goes through full authentication). Making this atomic would require coupling the view to a specific backend (e.g., Redis INCR), which would be a larger architectural change beyond the scope of a minimal fix.

src/aap_eda/api/views/external_event_stream.py
```diff
@@ -282,6 +285,30 @@ def _handle_auth(self, request, inputs):
                 )
             raise
 
+    def _get_client_ip(self, request):
```
mkanoor, line 311: @AlexSCorey Shouldn't these be done at the Envoy level instead of the underlying application. If there are multiple Event Stream workers then each of them will have its own cache.

src/aap_eda/api/views/external_event_stream.py
```diff
@@ -282,6 +285,30 @@ def _handle_auth(self, request, inputs):
...
+    def _check_rate_limit(self, request, event_stream_uuid):
+        key = (
+            f"es_auth_fail: {event_stream_uuid}: "
+            f"{self._get_client_ip(request)}"
+        )
+        failures = cache.get(key, 0)
+        if failures >= FAILURE_THRESHOLD:
+            raise AuthenticationFailed("Too many failed attempts")
+
+    def _record_failure(self, request, event_stream_uuid):
```
mkanoor: @AlexSCorey If we are trying to thwart an attacker who is sending arbitrary unique uuids this will create multiple keys and with the uuid plus the client ip.
I think the key should just have the client_ip if we detect a bogus request.

src/aap_eda/api/event_stream_authentication.py
```diff
@@ -154,7 +158,7 @@ def authenticate(self, _body=None):
 
         user_pass = f"{self.username}:{self.password}"
         b64_value = base64.b64encode(user_pass.encode()).decode()
-        if auth_str != b64_value:
+        if not hmac.compare_digest(auth_str.encode(), b64_value.encode()):
```
mkanoor: @AlexSCorey It might be better to pass in 2 strings to a function which does the compare_digest with a comment to indicate that it is for Preventing Timing Attacks (Constant-Time Comparison). And when we have to compare 2 strings we use this function. It can be used in other places too.

src/aap_eda/api/event_stream_authentication.py
```diff
@@ -152,9 +153,9 @@ def authenticate(self, _body=None):
         if self.authorization.startswith("Basic"):
             auth_str = self.authorization.split("Basic ")[1]
 
-        user_pass = f"{self.username}:{self.password}"
+        user_pass = f"{self.username}:{self.password}"  # noqa: E231
```
mkanoor: @AlexSCorey to slove the noqa issue
```
f"{self.username}:{self.password}"
```
```
auth_str = "%s:%s" % (self.username, self.password)
```

src/aap_eda/api/blacklist.py
```diff
@@ -0,0 +1,128 @@
...
+            get_org_settings,
+        )
+
+        org_settings = get_org_settings(org_id)
+        if org_settings is None:
+            return
+
+        normalized = normalize_ip(client_ip)
+
+        if org_settings["allowed_ips"] and not ip_in_allowlist(
```
wfealdel, line 90: `blocked_ips` is loaded into the organization policy but never checked here. If this field is intended to be an administrator-managed deny list, an IP explicitly added to `blocked_ips` can still post when the allowlist is empty. Please reject normalized client IPs found in `org_settings["blocked_ips"]` and consider add an integration test for this case.

src/aap_eda/api/blacklist.py
```diff
@@ -0,0 +1,128 @@
...
+
+        if len(setting.blocked_ips) >= MAX_BLOCKED_IPS:
+            logger.warning(
+                "Blocked IPs cap (%d) reached for org %s",
+                MAX_BLOCKED_IPS,
+                org_id,
+            )
+            return
+
+        setting.blocked_ips = [*setting.blocked_ips, normalized]
```
wfealdel, line 122: This read-modify-write is not safe under concurrent rejected requests. Two workers can read the same `blocked_ips`, append different addresses, and the last save silently discard the other update. Please update this list inside `transaction.atomic()` using `select_for_update()`, for example:

```python
from django.db import transaction

with transaction.atomic():
    setting = EventStreamSetting.objects.select_for_update().get(
        organization_id=org_id
    )

    if normalized not in setting.blocked_ips:
        setting.blocked_ips = [
            *setting.blocked_ips,
            normalized,
        ]
        setting.save(update_fields=["blocked_ips", "modified_at"])
```

src/aap_eda/api/serializers/event_stream_setting.py
```diff
@@ -0,0 +1,158 @@
...
+                    f"'{entry}' is not a valid IP address."
+                )
+            normalized.append(normalize_ip(entry))
+    return normalized
+
+
+class EventStreamSettingCreateSerializer(
+    serializers.ModelSerializer,
+):
+    organization_id = serializers.IntegerField(
```
wfealdel, line 59: `organization_id` remains writable during updates. If a setting moves from organization A to B, the post-save signal only invalidates B's cache entry, leaving A's cached policy stale for up to the TTL. Please make organization ownership immutable after creation or invalidate both the old and new organization cache keys.

src/aap_eda/services/event_stream_settings_cache.py
```diff
@@ -0,0 +1,82 @@
...
+    try:
+        setting = EventStreamSetting.objects.get(organization_id=org_id)
+        data = {
+            "allowed_ips": set(setting.allowed_ips),
+            "blocked_ips": set(setting.blocked_ips),
+        }
+    except EventStreamSetting.DoesNotExist:
+        data = None
+
+    cache.set(key, data, SETTINGS_CACHE_TTL)
```
wfealdel, line 62: This policy cache relies on Django's global cache without enforcing a shared production backend. With the default process-local cache, different workers can continue using stale allowlists after a policy change. Please require/configure a shared backend for production, or avoid caching this security policy across workers.

src/aap_eda/api/blacklist.py
AlexSCorey: I am removing this logic from the PR because it really is not enforceable at the eda-server level.

wfealdel (approved):
LGTM!

src/aap_eda/api/event_stream_authentication.py
```diff
@@ -152,9 +153,9 @@ def authenticate(self, _body=None):
         if self.authorization.startswith("Basic"):
             auth_str = self.authorization.split("Basic ")[1]
 
-        user_pass = f"{self.username}:{self.password}"
+        user_pass = "%s:%s" % (self.username, self.password)
```
ttuffin, line 156: @AlexSCorey why the change from f-string?

ttuffin: Nevermind I see it was already discussed [here](https://github.com/ansible/eda-server/pull/1649#discussion_r3799105749)

ttuffin (approved):

PR #1651, opened 2026-08-13
[AAP-76187] fix(security): suppress credential exposure in EDA debug logging

wfealdel:
Prevent credentials, authentication signatures, and sensitive request data from being written to persistent EDA activation logs when debug log level is enabled. Relates to CVE-2025-2877.

Multiple code paths exposed sensitive data in debug logs: ECDSA public keys and signatures were logged in plaintext, activation logs stored raw container output without sanitization, and the `_redacted_headers` method failed to redact the auth header when it appeared in a specific `additional_data_headers` list.

- Remove ECDSA debug logs that exposed public keys and signatures (`event_stream_authentication.py`)
- Create a reusable log sanitizer module (`log_sanitizer.py`) with `sanitize_data`, `sanitize_string`, and a `SensitiveDataFilter` logging filter
- Sanitize activation logs in `DBLogger.write()` before database persistence (AR-03)
- Sanitize activation logs in `ActivationInstanceLogSerializer` for historical data (AR-03)
- Fix `_redacted_headers` to redact auth header in specific header lists with case-insensitive comparison
- Add `SensitiveDataFilter` to all Django logging handlers for defense-in-depth
- Consolidate `REDACTED_STRING` constant into `log_sanitizer.py`

No new dependencies. No breaking changes. The `REDACTED_STRING` constant is now imported from `aap_eda.utils.log_sanitizer` instead of being defined in `external_event_stream.py`.

- Unit tests: `pytest tests/unit/test_log_sanitizer.py` (25 tests covering sanitize_data, sanitize_string, SensitiveDataFilter)
- Integration tests: `pytest tests/integration/api/test_event_stream_ecdsa.py` (ECDSA log-leak verification)
- Integration tests: `pytest tests/integration/api/test_event_stream_token.py` (header redaction including case-insensitive match)
- Integration tests: `pytest tests/integration/api/test_activation_instance.py` (serializer sanitization)
- Integration tests: `pytest tests/integration/services/activation/engine/test_podman.py` (DBLogger write-time sanitization)
- Manual: enable DEBUG logging, trigger ECDSA event stream auth, verify no sensitive values in container logs

https://redhat.atlassian.net/browse/AAP-76187

Assisted-by: Claude

## Summary by CodeRabbit

* **Security**
  * Sensitive tokens, authorization headers, signatures, and cryptographic keys are redacted from logs and API responses.
  * Redaction supports nested data, case-insensitive header matching, configurable sensitive fields, and exception tracebacks.
  * ECDSA authentication details are no longer exposed in diagnostic logging.

* **Bug Fixes**
  * Activation and event-stream logs consistently conceal sensitive values while preserving useful labels and context.

* **Tests**
  * Added coverage for log, API response, header, authentication, traceback, and persisted-data redaction scenarios.

### DIFF
--- a/src/aap_eda/api/serializers/activation.py
+++ b/src/aap_eda/api/serializers/activation.py
@@ -60,6 +60,7 @@ from aap_eda.core.utils.rulebook import (
     swap_event_stream_sources,
 )
 from aap_eda.core.utils.strings import substitute_variables
+from aap_eda.utils.log_sanitizer import sanitize_string
 
 logger = logging.getLogger(__name__)
 DE_NEEDED_MSG = "Decision Environment is needed"
@@ -1079,6 +1080,24 @@ class ActivationInstanceLogSerializer(serializers.ModelSerializer):
         model = models.RulebookProcessLog
         fields = "__all__"
         read_only_fields = ["id"]
+
+    def to_representation(self, instance: models.RulebookProcessLog) -> dict:
+        """Sanitize the log field to redact any sensitive data.
+
+        Handles historical log entries that were stored before
+        write-time sanitization was added.
+
+        Args:
+            instance: The RulebookProcessLog model instance.
+
+        Returns:
+            The serialized representation with sensitive values
+            in the log field replaced by a redaction marker.
+        """
+        data = super().to_representation(instance)
+        if "log" in data:
+            data["log"] = sanitize_string(data["log"])
+        return data
 
 
 class ActivationReadSerializer(
--- a/src/aap_eda/api/views/external_event_stream.py
+++ b/src/aap_eda/api/views/external_event_stream.py
@@ -48,10 +48,10 @@ from aap_eda.core.exceptions import CredentialPluginError, PGNotifyError
 from aap_eda.core.models import EventStream
 from aap_eda.core.utils.credentials import get_resolved_secrets
 from aap_eda.services.pg_notify import PGNotify
+from aap_eda.utils.log_sanitizer import REDACTED_STRING
 
 logger = logging.getLogger(__name__)
 UNSAFE_HEADER_KEYS = {"X-Trusted-Proxy", "X-Forwarded-For", "X-Real-IP"}
-REDACTED_STRING = "********"
 
 
 class ExternalEventStreamViewSet(viewsets.GenericViewSet):
@@ -157,26 +157,50 @@ class ExternalEventStreamViewSet(viewsets.GenericViewSet):
     def _redacted_headers(
         self, headers: HttpHeaders, header_key: str
     ) -> dict[str, Any]:
+        additional = self.event_stream.additional_data_headers
+        if not additional:
+            return {}
+
+        header_key_lower = header_key.lower()
+
+        if additional == "*":
+            return self._redact_wildcard_headers(headers, header_key_lower)
+
+        return self._redact_listed_headers(
+            headers, header_key_lower, additional
+        )
+
+    def _redact_wildcard_headers(
+        self, headers: HttpHeaders, header_key_lower: str
+    ) -> dict[str, Any]:
+        event_headers = {
+            key: (
+                REDACTED_STRING if key.lower() == header_key_lower else value
+            )
+            for key, value in headers.items()
+        }
+        return {
+            key: value
+            for key, value in event_headers.items()
+            if not key.startswith("X-Envoy") and key not in UNSAFE_HEADER_KEYS
+        }
+
+    def _redact_listed_headers(
+        self,
+        headers: HttpHeaders,
+        header_key_lower: str,
+        additional: str,
+    ) -> dict[str, Any]:
         event_headers = {}
-        if self.event_stream.additional_data_headers:
-            if self.event_stream.additional_data_headers == "*":
-                event_headers = dict(headers)
-                if header_key in event_headers:
-                    event_headers[header_key] = REDACTED_STRING
-                event_headers = {
-                    key: value
-                    for key, value in event_headers.items()
-                    if not key.startswith("X-Envoy")
-                    and key not in UNSAFE_HEADER_KEYS
-                }
+        for key in additional.split(","):
+            key = key.strip()
+            value = headers.get(key)
+            if not value:
+                continue
+            if key.lower() == header_key_lower:
+                event_headers[key] = REDACTED_STRING
             else:
-                for key in self.event_stream.additional_data_headers.split(
-                    ","
-                ):
-                    key = key.strip()
-                    value = headers.get(key)
-                    if value:
-                        event_headers[key] = value
+                event_headers[key] = value
         return event_headers
 
     def _create_payload(
--- a/src/aap_eda/utils/log_sanitizer.py
+++ b/src/aap_eda/utils/log_sanitizer.py
@@ -1,0 +1,105 @@
+#  Copyright 2026 Red Hat, Inc.
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+import logging
+import re
+import traceback
+
+REDACTED_STRING = "********"
+
+SENSITIVE_FIELD_NAMES = (
+    "access_token",
+    "api_key",
+    "authorization",
+    "client_secret",
+    "credential",
+    "password",
+    "private_key",
+    "public_key",
+    "refresh_token",
+    "secret",
+    "signature",
+    "token",
+)
+
+_SENSITIVE_FIELDS_RE = "|".join(re.escape(f) for f in SENSITIVE_FIELD_NAMES)
+
+# PEM markers per RFC 7468: exactly 5 dashes, uppercase label.
+_PEM_BLOCK_PATTERN = re.compile(
+    rf"\"?({_SENSITIVE_FIELDS_RE})\"?(\s*[:=]\s*)\"?"
+    r"(-----BEGIN\s[A-Z\s]+-----.*?-----END\s[A-Z\s]+-----)",
+    re.IGNORECASE | re.DOTALL,
+)
+
+_SENSITIVE_KV_PATTERN = re.compile(
+    rf"\"?({_SENSITIVE_FIELDS_RE})\"?(\s*[:=]\s*)\"?(?:.+)",
+    re.IGNORECASE,
+)
+
+
+def sanitize_string(text: str) -> str:
+    """Redact sensitive key=value pairs and Authorization headers.
+
+    Applies regex substitution to mask values following sensitive
+    field names (e.g. ``token=abc123``) and Authorization header
+    values (e.g. ``Authorization: Bearer abc123``).
+
+    Args:
+        text: The input string to sanitize. Non-string values are
+            returned unchanged.
+
+    Returns:
+        The input string with sensitive values replaced by the
+        REDACTED constant.
+    """
+    if not isinstance(text, str):
+        return text
+
+    text = _PEM_BLOCK_PATTERN.sub(rf"\1\2{REDACTED_STRING}", text)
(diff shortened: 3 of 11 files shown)


src/aap_eda/api/views/external_event_stream.py
```diff
@@ -156,27 +156,44 @@ def _parse_body(self, content_type: str, body: bytes) -> dict:
...
+        self, headers: HttpHeaders, header_key: str
+    ) -> dict[str, Any]:
+        event_headers = dict(headers)
+        if header_key in event_headers:
+            event_headers[header_key] = REDACTED_STRING
+        return {
+            key: value
+            for key, value in event_headers.items()
+            if not key.startswith("X-Envoy") and key not in UNSAFE_HEADER_KEYS
+        }
```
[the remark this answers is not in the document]
wfealdel: fixed

src/aap_eda/utils/log_sanitizer.py
```diff
@@ -0,0 +1,143 @@
...
+
+_SENSITIVE_KEYS = re.compile(
+    rf"({_SENSITIVE_FIELDS_RE})",
+    re.IGNORECASE,
+)
+
+_SENSITIVE_KV_PATTERN = re.compile(
+    rf"({_SENSITIVE_FIELDS_RE})(\s*[:=]\s*)(.+)",
+    re.IGNORECASE,
+)
```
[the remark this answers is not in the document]
wfealdel: low risk. DBLogger.write() processes plain-text log lines, not JSON. Fixing this would significantly complicate the regex for minimal gain

src/aap_eda/utils/log_sanitizer.py
```diff
@@ -0,0 +1,143 @@
...
+        """Sanitize the log record after formatting.
+
+        Args:
+            record: The log record to inspect and sanitize.
+
+        Returns:
+            Always ``True`` so the record is emitted after sanitization.
+        """
+        record.msg = sanitize_string(record.getMessage())
+        record.args = None
```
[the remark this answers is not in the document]
wfealdel: Marginal. The codebase already avoids putting credentials in exception messages — auth classes raise generic messages like "Credential mismatch", never the actual values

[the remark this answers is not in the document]
wfealdel: Exception tracebacks are now sanitized, with test coverage.

src/aap_eda/utils/log_sanitizer.py
```diff
@@ -0,0 +1,139 @@
...
+_SENSITIVE_KV_PATTERN = re.compile(
+    rf"({_SENSITIVE_FIELDS_RE})(\s*[:=]\s*)(.+)",
+    re.IGNORECASE,
+)
+
+
+def sanitize_data(
+    data: Any,
+    extra_sensitive_keys: set[str] | None = None,
+) -> Any:
```
ptoscano: This function seems to be used only in tests. If not really used/needed, simply drop it altogether (together with `_SENSITIVE_KEYS`, used only here).

src/aap_eda/utils/log_sanitizer.py
```diff
@@ -0,0 +1,139 @@
...
+
+_SENSITIVE_FIELDS_RE = "|".join(re.escape(f) for f in SENSITIVE_FIELD_NAMES)
+
+_SENSITIVE_KEYS = re.compile(
+    rf"({_SENSITIVE_FIELDS_RE})",
+    re.IGNORECASE,
+)
+
+_SENSITIVE_KV_PATTERN = re.compile(
+    rf"({_SENSITIVE_FIELDS_RE})(\s*[:=]\s*)(.+)",
```
ptoscano: The 3rd capture seems unused.

src/aap_eda/api/views/external_event_stream.py
```diff
@@ -156,29 +156,66 @@ def _parse_body(self, content_type: str, body: bytes) -> dict:
...
-                    event_headers[header_key] = REDACTED_STRING
-                event_headers = {
-                    key: value
-                    for key, value in event_headers.items()
-                    if not key.startswith("X-Envoy")
-                    and key not in UNSAFE_HEADER_KEYS
-                }
+        for requested_key in additional.split(","):
+            requested_key = requested_key.strip()
+            canonical_key = self._find_header_key(headers, requested_key)
```
ptoscano: According to the django documentation, it seems the `HttpHeaders` class already handles case-insensitive keys on its own:
https://docs.djangoproject.com/en/5.2/ref/request-response/#django.http.HttpRequest.headers

Which means that the previous logic (the one in `_redacted_headers`) was already working correctly, as the `get()` already did case-insensitive lookups. In this case, the old logic can still be used and easily tweaked:
```python
        for key in additional.split(","):
            key = key.strip()
            value = headers.get(key)
            if value:
                event_headers[key] = REDACTED_STRING if key.lower() == header_key.lower() else value
```

src/aap_eda/api/serializers/activation.py
```diff
@@ -1080,6 +1081,23 @@ class Meta:
...
+
+        Args:
+            instance: The RulebookProcessLog model instance.
+
+        Returns:
+            The serialized representation with sensitive values
+            in the log field replaced by a redaction marker.
+        """
+        data = super().to_representation(instance)
+        data["log"] = sanitize_string(data.get("log", ""))
```
ptoscano: If `log` is not an existing key in `data`, then this line will add it as empty line. I'd argue here that if `log` does not exist it ought to not be added.

src/aap_eda/utils/log_sanitizer.py
```diff
@@ -0,0 +1,139 @@
...
+    "token",
+    "api_key",
+    "access_token",
+    "refresh_token",
+    "authorization",
+    "signature",
+    "public_key",
+    "private_key",
+    "client_secret",
+    "credential",
```
ptoscano: Minor nit: please sort them alphabetically, so it's easier to visually check them.

src/aap_eda/api/views/external_event_stream.py
```diff
@@ -156,29 +156,66 @@ def _parse_body(self, content_type: str, body: bytes) -> dict:
...
         self, headers: HttpHeaders, header_key: str
+    ) -> dict[str, Any]:
+        additional = self.event_stream.additional_data_headers
+        if not additional:
+            return {}
+
+        if additional == "*":
+            return self._redact_wildcard_headers(headers, header_key)
+
+        return self._redact_listed_headers(headers, header_key, additional)
```
ptoscano: If `header_key` is lowercased here before being passed to `_redact_wildcard_headers()` or `_redact_listed_headers()`, it can save more `lower()` calls later on.

ptoscano (changes requested):
Thanks for the PR, there are some things to change and possibly remove.

wfealdel:
> Thanks for the PR, there are some things to change and possibly remove.

@ptoscano 
Thanks for the review! Good catches, I already addressed those issues.

ptoscano (dismissed):

src/aap_eda/utils/log_sanitizer.py
```diff
@@ -0,0 +1,97 @@
...
+    "signature",
+    "token",
+)
+
+_SENSITIVE_FIELDS_RE = "|".join(re.escape(f) for f in SENSITIVE_FIELD_NAMES)
+
+_SENSITIVE_KV_PATTERN = re.compile(
+    rf"({_SENSITIVE_FIELDS_RE})(\s*[:=]\s*)(?:.+)",
+    re.IGNORECASE,
+)
```
B-Whitt, line 48: @wfealdel This finding was partially addressed above, but still gives opportunity for a leak ...

`_SENSITIVE_KV_PATTERN` expects a sensitive field name immediately followed by whitespace/separator (`\s*[:=]\s*`), but in JSON-style log output like `{"token": "abcdef123456"}`, a quote character sits between the field name and the colon. The regex can't consume `"`, so it never matches. `SensitiveDataFilter` applies `sanitize_string` across all Django logging handlers — not just `DBLogger.write()` — so any `logger.debug("payload: %s", json.dumps(data))` call elsewhere in the app would pass through unredacted.

Verified: `sanitize_string('{"token": "abcdef123456"}')` returns the input unchanged.

src/aap_eda/utils/log_sanitizer.py
```diff
@@ -0,0 +1,97 @@
...
+    "refresh_token",
+    "secret",
+    "signature",
+    "token",
+)
+
+_SENSITIVE_FIELDS_RE = "|".join(re.escape(f) for f in SENSITIVE_FIELD_NAMES)
+
+_SENSITIVE_KV_PATTERN = re.compile(
+    rf"({_SENSITIVE_FIELDS_RE})(\s*[:=]\s*)(?:.+)",
```
B-Whitt: The KV pattern's `.+` does not match newlines (no `re.DOTALL`), so only the first line of a multi-line secret is redacted — continuation lines leak verbatim. This is directly relevant since the PR's stated goal is suppressing ECDSA key/signature leakage, and PEM keys are inherently multi-line. This gap is also reachable via `exc_info` tracebacks (which contain real `\n`), not just literal key-value logging.

Verified: `sanitize_string("private_key=-----BEGIN KEY-----\nMIILineTwo\nMIILineThree")` redacts only the first line; `MIILineTwo` and `MIILineThree` survive unredacted.

tests/unit/test_log_sanitizer.py
```diff
@@ -0,0 +1,178 @@
...
+            args=("secret123",),
+            exc_info=None,
+        )
+        log_filter.filter(record)
+
+        assert "secret123" not in record.msg
+        assert REDACTED_STRING in record.msg
+        assert record.args is None
+
+    def test_preserves_format_placeholders(self, log_filter):
```
B-Whitt, line 152: The test suite covers plain `key=value` and `key: value` patterns but has no test case for a JSON-quoted key (`{"token": "x"}`) or a value containing an embedded newline (multi-line PEM). Adding two tests would lock in fixes for the regex gaps above and prevent regressions:

- `test_redacts_json_quoted_key` — assert `sanitize_string('{"token": "abc123"}')` no longer contains `abc123`
- `test_redacts_multiline_pem` — assert a real multi-line PEM block (`"private_key=-----BEGIN KEY-----\nMIILineTwo\n-----END KEY-----"`) is fully masked, not just the first line

src/aap_eda/utils/log_sanitizer.py
ptoscano: > The KV pattern's `.+` does not match newlines (no `re.DOTALL`), so only the first line of a multi-line secret is redacted — continuation lines leak verbatim.

While this AI suggestion seems useful at first, in practice it leads to a mess.

Handling newlines as any other character means that `.+` matches them. Considering that `+` is greedy by default [1], then it means that `.+` + `re.DOTALL` matches _every single character until the end of the input string_, eating up whatever is after the end of the certificate/key. This can be checked easily by tweaking the newly added `test_redacts_multiline_pem` test:
```python
    def test_redacts_multiline_pem(self):
        text = (
            "private_key=-----BEGIN KEY-----\n"
            "MIILineTwo\n"
            "-----END KEY-----\n"
            "foo=bar"
        )
        result = sanitize_string(text)
        assert "MIILineTwo" not in result
        assert REDACTED_STRING in result
        assert "foo" in result
```
The last assert fails because `result` is `private_key=********`.

If certificates/keys must be redacted, then they need an own strategy:
- their matching ought to be the only case with newlines accepted (unlike regular fields, for which newlines must be hard-stop)
- properly match from the known markers (e.g. `-----BEGIN KEY-----`) to the corresponding ends, disabling the greedy behaviour inside (so that a matching doesn't eat up multiple certificates/keys at once)

[1] https://docs.python.org/3/library/re.html#regular-expression-syntax

wfealdel: @ptoscano
Good catch!
I added a dedicated `_PEM_BLOCK_PATTERN` that matches from `-----BEGIN ... ----- to -----END ... -----`  (RFC 7468 compliant markers), and the existing `_SENSITIVE_KV_PATTERN` without `re.DOTALL` which stops at newlines. PEM blocks are fully redacted without consuming past the end marker, and regular KV fields don't eat subsequent lines. Added tests for both boundaries (`test_pem_does_not_consume_past_end_marker`, `test_does_not_consume_past_newline`). Could you check again, pls?
cc: @B-Whitt

B-Whitt: Looks good @wfealdel

B-Whitt (approved):
looks good

ptoscano (approved):
thanks for the changes, LGTM now

PR #1660, opened 2026-08-26
[AAP-78702] Add CleanTextMixin to serializers

daphnemaeve:
### Description
Jira ticket: https://redhat.atlassian.net/browse/AAP-78702

Adds `CleanTextMixin` from django-ansible-base to EDA API serializers that handle create and update operations. This implements two-tier text field validation:

- **Tier 1 (name fields)**: strict character allowlist — only letters, numbers, spaces, hyphens, underscores, dots, and `@`
- **Tier 2 (other text fields)**: dangerous pattern blocklist — rejects HTML tags, JavaScript protocols, event handlers, shell substitution, backticks, null bytes, and ANSI escapes

### **Why is this change needed?**

EDA endpoints currently accept unsanitized text input in resource names and free-text fields, creating potential XSS and injection vulnerabilities.

### **How does this change address the issue?**

By adding `CleanTextMixin` as the first mixin in the inheritance chain for each serializer, validation automatically runs as part of `serializer.validate()`. Each overridden `validate()` now chains to `super().validate(...)` so the mixin's checks actually execute. The mixin:

- Auto-discovers text fields (`CharField`, `TextField`) from the model
- Applies the appropriate validation tier based on field name
- Grandfathers existing invalid data (unchanged values on update pass validation)
- Is gated behind the `ENHANCED_INPUT_VALIDATION_ENABLED` install-time setting (gating logic lives in DAB within `CleanTextMixin`; EDA does not enable it by default in this PR)

### **Changes**

Added `CleanTextMixin` to the write-path serializers across:

- Activations: `ActivationSerializer`, `ActivationCreateSerializer`, `ActivationCopySerializer`, `ActivationUpdateSerializer`
- Credentials: `EdaCredentialSerializer`, `EdaCredentialCopySerializer`, `EdaCredentialCreateSerializer`, `EdaCredentialUpdateSerializer`, `CredentialInputSourceSerializer`, `CredentialInputSourceCreateSerializer`, `CredentialInputSourceUpdateSerializer`, `CredentialTypeSerializer`, `CredentialTypeCreateSerializer`, `AwxTokenCreateSerializer`
- Decision environments: `DecisionEnvironmentSerializer`, `DecisionEnvironmentCreateSerializer`
- Event streams: `EventStreamInSerializer`, `EventStreamOutSerializer`
- Projects: `ProjectSerializer`, `ProjectCreateRequestSerializer`, `ProjectUpdateRequestSerializer`
- Organizations, teams, users: `OrganizationSerializer`, `OrganizationCreateSerializer`, `TeamSerializer`, `TeamCreateSerializer`, `TeamUpdateSerializer`, `UserSerializer`, `UserUpdateSerializerBase`

Fields that legitimately contain non-plain-text content are excluded via `excluded_fields` so they aren't rejected as unsafe:

- `extra_var` (activation) — Jinja2 template syntax from credential injectors
- `injectors` (credential type) — Jinja2 template syntax
- `inputs` (eda credential) and `metadata` (credential input source) — arbitrary credential payloads
- `token` (AWX token) — secret token value

### **Validation Behavior**

- Tier 1 (name fields): strict character allowlist (alphanumeric, hyphen, underscore, space, period, `@`)
- Tier 2 (other text): dangerous pattern blocklist (HTML tags, JS protocol, event handlers, shell syntax)
- Grandfathering: unchanged invalid values on update pass validation (backward compatible)
- Install-time setting: gated behind `ENHANCED_INPUT_VALIDATION_ENABLED`, disabled by default until enabled at the deployment/manifest level

### **Does this change introduce any new dependencies, blockers or breaking changes?**

Adds nh3 and regex.

No behavior change for existing deployments until `ENHANCED_INPUT_VALIDATION_ENABLED` is turned on. Once enabled, new/changed values with unsafe names or content will be rejected on create/update.

#### Manual testing
Prerequisites
• aap-dev environment set up and functional
• Access to the EDA UI

1. Environment Setup

1. Configure your aap-dev sources to point at:
   - DAB → devel branch
   - EDA Server → PR branch (refs/pull/1660/head or the author's feature branch)

2. Start the environment and ensure the EDA UI is accessible.

3. Enable the feature gate — set this in your environment:ENHANCED_INPUT_VALIDATION_ENABLED=True
(The validation is a no-op when this is False, so you must enable it.)

2. Test Inputs

Use these values throughout:

INVALID name:        <script>alert(1)</script>
INVALID description: $(rm -rf /)
VALID name:          My Test Resource
VALID description:   A perfectly normal description

Tier 1 rejects HTML/XSS-style patterns in name fields. Tier 2 rejects shell injection patterns in description and other free-text fields.

##### Test cases
For each resource below, test via the EDA UI. Every create and update action should now validate name and description fields.

3a. Organizations
- Create → Enter <script>alert(1)</script> as the name → expect validation error
- Create → Enter My Test Org as name, $(rm -rf /) as description → expect validation error
- Create → Enter My Test Org as name, A normal description → expect success

3b. Teams
- Create → Enter <script>alert(1)</script> as the name → expect validation error
- Create → Enter Valid Team Name, $(rm -rf /) as description → expect validation error
- Create → Enter Valid Team Name, Normal description → expect success

3c. Projects
- Create → Enter <script>alert(1)</script> as the project name → expect validation error
- Create → Enter a valid name, $(rm -rf /) as description → expect validation error
- Create → Enter a valid name and description with a valid SCM URL → expect success

3d. Decision Environments
- Create → Enter <script>alert(1)</script> as the name → expect validation error
- Create → Enter a valid name with a valid image URL → expect success

3e. Credentials
- Create → Enter <script>alert(1)</script> as the credential name → expect validation error
- Create → Enter $(rm -rf /) as the description → expect validation error
- Create → Enter a valid name and description → expect success

3f. Activations
- Create → Enter <script>alert(1)</script> as the activation name → expect validation error
- Create → Enter $(rm -rf /) as the description → expect validation error
- Create → Enter a valid name and description → expect success

3g. Users
- Create → Enter <script>alert(1)</script> as the username → expect validation error
- Create → Enter a valid username (e.g. valid.user123) → expect success

#### Automated tests
- `tests/integration/api/test_clean_text_mixin.py` covers valid text, rejected unsafe content, and update/grandfathering behavior across all updated serializers with `ENHANCED_INPUT_VALIDATION_ENABLED=True`.
- ATF tests: https://gitlab.cee.redhat.com/ansible/testing/eda-test-suite/-/merge_requests/142

## Bug Fixes

- Improved text validation for activations, credentials, projects, teams, users, organizations, event streams, and decision environments.
- Unsafe or invalid text is rejected while structured metadata, credential inputs, tokens, and other supported fields remain unaffected.
- Activation copies now validate descriptions, service names, and source mappings before duplication.
- Updates continue to support unchanged, pre-existing values that do not meet newer validation rules.

## Tests

- Added integration coverage for valid and rejected text, excluded fields, copy validation, and update handling.


src/aap_eda/api/serializers/activation.py
```diff
@@ -708,7 +718,7 @@ def create(self, validated_data):
         return super().create(validated_data)
 
 
-class ActivationCopySerializer(serializers.ModelSerializer):
+class ActivationCopySerializer(CleanTextMixin, serializers.ModelSerializer):
```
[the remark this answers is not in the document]
daphnemaeve: ActivationCopySerializer.Meta.fields = ["name"] — description is not a declared field on this serializer at all. In [activation.py:648-655](vscode-webview://0fj1q8uen75kpqrm90v5fig0k4o5cl4c4rgqhb6onu8b0of5n10u/src/aap_eda/api/views/activation.py#L648-L655), the view does ActivationCopySerializer(instance=activation, data=request.data); DRF only pulls name into validated_data/attrs, so any description a client tries to slip into the POST body is simply ignored. The copy() method then explicitly sets copied_data["description"] = activation.description — it always copies the source activation's own already-persisted description, never anything from the request.

Since description is never new user input in this endpoint, there's nothing to validate through CleanTextMixin — the mixin's _validate_text_fields correctly skips it because it's absent from attrs (mixins.py: if field_name in self.excluded_fields or field_name not in attrs: continue). The one genuinely new, user-controlled field in this flow is name, and that's already validated: it's declared on the serializer, present in attrs, and "name" is in DAB's DEFAULT_NAME_FIELDS, so it goes through Tier-1 validation as intended.

Copying forward a description that predates ENHANCED_INPUT_VALIDATION_ENABLED (or was valid when the source activation was created) doesn't introduce new exposure — that content is already stored and already rendered wherever the source activation's description is shown.

Verdict: skipping this — no code change needed. The suggested "fix" would mean re-validating data that's copied server-side from a trusted, already-persisted record, not attacker-controlled input, which isn't what CleanTextMixin is meant to guard.

vidyanambiar:
## Code Review: [AAP-78702] Add CleanTextMixin to EDA serializers (github.com/ansible/eda-server#1660)

### Verdict: NEEDS_CHANGES

### Scores
| Lens | Score | Findings |
|---|---|---|
| Functionality | 10/10 | 0 scored (1 Nit, 0 pts) |
| Security | 10/10 | 0 new scored (1 unresolved known blocker, not re-scored) |
| Quality | 7.5/10 | 2 Major, 1 Minor |
| **Overall** | **9.2/10** | |

Verdict is NEEDS_CHANGES because of (a) two Major Quality findings below and (b) an unresolved Major-severity known blocker (CodeRabbit, still present) — see **Known Blockers** section.

---

### Known Blockers (from existing PR discussion — verified, not duplicated as new findings)

1. **`src/aap_eda/api/serializers/activation.py:721-791`** — @coderabbitai (Major): *"When enhanced input validation is enabled, ActivationCopySerializer validates only name. copy() then adds activation.description and passes it to super().create(copied_data), so description bypasses CleanTextMixin and can persist blocklisted text."*

   **Status: CONFIRMED, still unresolved at HEAD (a5d80d7).** Independently verified by both review agents and the coordinator:
   - `ActivationCopySerializer.Meta.fields = ["name"]` — the only field DRF/CleanTextMixin ever sees via `is_valid()`.
   - `copy()` builds `copied_data` including `"description": activation.description` plus ~20 other fields, then calls `return super().create(copied_data)` directly (`serializers.ModelSerializer.create()`), which never invokes `validate()`.
   - View confirms the chain: `serializer.is_valid(raise_exception=True)` (validates only `name`) → `serializer.copy()` (persists everything else unvalidated) — `src/aap_eda/api/views/activation.py:650-655`.
   - **No test coverage**: `tests/integration/api/test_clean_text_mixin.py` has a `TestActivationCleanText` class but no test exercises the `/copy/` action (`grep -i copy` on the file matches only license-header text).
   - **Sibling check (ruled out)**: `EdaCredentialCopySerializer` does **not** have this problem — its `copy()` view action routes through `EdaCredentialCreateSerializer(data=post_data).is_valid()`, so full CleanTextMixin validation runs on the copied description. `EdaCredentialCopySerializer` itself is only ever referenced in `@extend_schema(request=...)` for docs, never instantiated in view logic.
   - **Pattern propagation**: searched the entire diff and all `copy`/`clone`/`duplicate` actions repo-wide — Activation and EdaCredential are the only two copy paths in the codebase; no other instance of this bug class exists.

2. Two false leads checked and ruled out (not real gaps, included here for transparency, not scored):
   - **`UserUpdateIsSuperuserSerializer`** (`user.py:161-173`) overrides `validate()` without calling `super()`, silently skipping `CleanTextMixin` — but `Meta.fields = ["is_superuser"]` is a `BooleanField` only, so there's no text field to protect. Immaterial, and pre-existing (unrelated to this diff).
   - **`Project.proxy`** (`EncryptedTextField`) is not in `ProjectCreateRequestSerializer`/`ProjectUpdateRequestSerializer`'s `excluded_fields`, unlike the analogous `inputs`/`metadata`/`token` fields elsewhere. Considered as a possible gap, but refuted: `proxy` isn't name-pattern-matched into Tier 1, so at worst a legitimate proxy URL with a blocklisted character gets a loud 400 rejection — a usability nit, not an injection/bypass risk. No exploitable attack surface; not scored.

---

### Findings

#### Critical
(none)

#### Major

1. **[Quality]** **`src/aap_eda/api/serializers/activation.py:721-791`** `ActivationCopySerializer.Meta.fields = ["name"]` declares a one-field write contract, but `copy()` persists ~20 additional fields (including `description`) via a hand-built dict passed straight to `super().create()`. This is a leaky abstraction distinct from the security angle above: a maintainer reading `Meta.fields` would reasonably assume that's the entire validated/writable surface of the serializer, which is false.
   - **Evidence**: `class Meta:\n        model = models.Activation\n        fields = ["name"]` … `copied_data = {"name": self.validated_data["name"], "description": activation.description, ...}` … `return super().create(copied_data)`
   - **Confidence**: HIGH
   - **Fix**: Route `copy()` through a serializer whose `Meta.fields` reflects everything actually persisted (so `is_valid()` covers all of it and picks up `CleanTextMixin`), or at minimum validate `description` explicitly before calling `create()`.
   - **Points**: 1

2. **[Quality]** **`tests/integration/api/test_clean_text_mixin.py`** — no test asserts that content in any `excluded_fields` entry (`extra_var`, `injectors`, `inputs`, `metadata`, `token`) is actually **accepted** once `ENHANCED_INPUT_VALIDATION_ENABLED=True`. The 762-line suite thoroughly tests that `name`/`description` get rejected/accepted, but never exercises the exemption path itself. A future typo in an `excluded_fields` value, or an upstream DAB change to auto-discovery, would silently start rejecting legitimate Jinja2/encrypted-secret content with no test catching it — a silent-failure risk one level above a loud one.
   - **Evidence**: no reference to `extra_var`, `injectors`, or `token` anywhere in the test file; no assertion of `HTTP_201_CREATED`/`is_valid()==True` for a payload containing Jinja2 template syntax or shell-metacharacter-bearing secret values.
   - **Confidence**: HIGH
   - **Fix**: Add e.g. `test_extra_var_jinja2_content_accepted` (Activation, posting `extra_var` containing `{{ template }}` and asserting 201), `test_injectors_jinja2_content_accepted` (CredentialType), and `test_token_value_not_validated` (AwxToken, with a token value that would otherwise trip the Tier-2 blocklist).
   - **Points**: 1

#### Minor

3. **[Quality]** `CleanTextMixin` was added to four serializers confirmed never used to process a write (no `data=` instantiation anywhere in `src/` or `tests/`), undercutting the PR's stated "write-path serializers" framing:
   - `ActivationSerializer` (`activation.py:370`) — entirely unused for input anywhere in the codebase.
   - `EventStreamOutSerializer` (`event_stream.py:99`) — only ever instantiated with `instance`/`many=True` for response bodies.
   - `EdaCredentialCopySerializer` (`eda_credential.py:128`) — only referenced via `@extend_schema(request=...)`; the real copy flow uses `EdaCredentialCreateSerializer`.
   - `OrganizationCreateSerializer` (`organization.py:71`) — `get_serializer_class()` always returns `OrganizationSerializer`, even for create; `OrganizationCreateSerializer` is never wired in despite its name.

   Cross-checked every real write path in the 8 affected domains against `get_serializer_class`/`data=` usage — confirmed **no actual write-path serializer was missed** by the PR; this is dead weight plus a process signal, not a coverage gap.
   - **Evidence**: see grep results in Verification Results below.
   - **Confidence**: HIGH
   - **Fix**: Remove `CleanTextMixin` from these four classes (no behavioral effect today), or add a one-line comment noting they're response/schema-only.
   - **Points**: 0.5

#### Nit

4. **[Functionality]** **`organization.py`, `team.py` vs. `user.py`** — inconsistent ordering of `self.validate_shared_resource()` relative to `super().validate(data)` across the touched `validate()` overrides (some call the shared-resource check before the mixin, `UserUpdateSerializerBase` after). Both orders produce a correct 4xx; only which error a client sees first differs when both would fail. No functional impact.
   - **Evidence**: `self.validate_shared_resource()\n        return super().validate(data)` (organization.py) vs. `data = super().validate(data)\n        self.validate_shared_resource()\n        return data` (user.py)
   - **Confidence**: HIGH
   - **Fix**: Not required — purely stylistic.
   - **Points**: 0

---

### MRO / super() Chain Verification (Functionality lens — full trace, all 10 files)

Every one of the ~28 touched serializer classes was traced individually. `CleanTextMixin` is listed first (or otherwise MRO-safe) in every touched base-class tuple, and every overridden `validate()` in a class carrying the mixin correctly calls and returns `super().validate(data)` on **every** return path, including the two-branch case in `EdaCredentialUpdateSerializer.validate()`. No broken chains found.

### excluded_fields Accuracy Check (Functionality lens)

| Serializer(s) | excluded_fields | Verified against model | Result |
|---|---|---|---|
| Activation family | `{"extra_var"}` | `Activation.extra_var = TextField` | Correct |
| CredentialType family | `{"injectors"}` | `CredentialType.injectors = JSONField` | Correct (defensive/redundant — not a TextField subtype anyway) |
| EdaCredential family | `{"inputs"}` | `EdaCredential.inputs = EncryptedTextField(TextField subclass)`, exposed as `serializers.JSONField()` | Correct and **necessary** — without it, dict values would be validated as free text |
| CredentialInputSource family | `{"metadata"}` | same `EncryptedTextField` pattern | Correct and necessary |
| AwxTokenCreateSerializer | `{"token"}` | `AwxToken.token = EncryptedTextField`, real string field | Correct and necessary |

No typos found in any `excluded_fields` entry.

---

### Needs Human Judgment

- **Jira AAP-78702** could not be fetched (auth-gated `redhat.atlassian.net`, no credentials available in this session) — acceptance criteria/DoD not cross-checked against the implementation. Only the PR description was used for change intent.
- **CI/pipeline status** could not be checked — both GitHub auth paths (SSH and the `gh` token wrapper) were unavailable in this session. CI state on PR #1660 is unknown to this review.
- **Other unresolved PR discussions** — only the single CodeRabbit comment pointed out by the requester was checked. Any *other* open review threads on PR #1660 are not accounted for in this review.
- **Backward-compatibility of `ENHANCED_INPUT_VALIDATION_ENABLED`**: turning the flag on rejects new/changed values that were previously accepted. Grandfathering covers *unchanged* values on update, but there's no migration/audit tooling in this PR to help an operator find existing data that would newly fail validation. Legitimate rollout question, not a diff defect.

---

### Verification Results
| Check | Command | Output | Result |
|-------|---------|--------|--------|
| Diff scope | `git diff main...HEAD --stat` | 10 serializer files + 1 new test file, 876+/37- | Matches PR description |
| Dependency changes | `git diff main...HEAD -- pyproject.toml poetry.lock` | (empty) | No dependency changes |
| Copy/clone actions repo-wide | `grep -rn "def copy\|def clone\|def duplicate" src/aap_eda/api/views/*.py` | Only `activation.py:648`, `eda_credential.py:293` | Only 2 copy paths exist; both analyzed |
| Direct create/save bypass scan | `grep -n "super().create(\|super().save(\|.objects.create(\|serializer.save(" <10 files>` | 4 hits; only `activation.py:791` (`ActivationCopySerializer.copy()`) bypasses validation | Confirms scope of the known blocker |
| `ActivationSerializer(` write usage | `grep -rn "ActivationSerializer(" src tests` | Only class def + unrelated `PostActivationSerializer` | Confirmed dead for writes |
| `EventStreamOutSerializer(` write usage | `grep -rn "EventStreamOutSerializer(" src tests` | Only `instance`/`many=True` call sites | Confirmed read-only |
| `EdaCredentialCopySerializer` usage | `grep -rn "EdaCredentialCopySerializer" src tests` | Only `@extend_schema` + class def | Schema-doc only |
| `OrganizationCreateSerializer` usage | `grep -rn "OrganizationCreateSerializer" src tests` | Only `@extend_schema` + class def; `get_serializer_class` always returns `OrganizationSerializer` | Never wired to create flow |
| `extra_var`/`injectors`/`inputs`/`metadata`/`token` real field check | `grep -n "<field>" core/models/*.py` | All 5 confirmed real fields on their respective models | No typos in `excluded_fields` |
| `ENHANCED_INPUT_VALIDATION_ENABLED` wiring | `grep -rn "ENHANCED_INPUT_VALIDATION_ENABLED" src/` | Only in the new test file's `@override_settings` | Setting sourced entirely from django-ansible-base, as expected |
| Test coverage for copy bypass | `grep -in "copy" tests/integration/api/test_clean_text_mixin.py` | Only license-header matches | Confirms no test exercises either copy action |

### Incidental Findings (out of scope — not scored)
- `UserUpdateIsSuperuserSerializer.validate()` skips `super()` entirely — pre-existing, unrelated to this diff, immaterial (no text field in scope).
- `django-ansible-base` is pinned to a floating `branch = "devel"` in `pyproject.toml` (anchored only via `poetry.lock`'s resolved commit) — pre-existing dependency-pinning strategy, unchanged by this diff.
- `EdaCredentialTestSerializer` (dry-run "test credential" action) isn't wrapped in `CleanTextMixin` and wasn't touched by this PR — doesn't persist data, so no persistence-of-unsafe-text risk today, but worth a follow-up if that view's semantics ever change.
- `EventStreamViewSet.partial_update` manually copies `validated_data` onto the instance and calls `.save()` directly rather than `serializer.save()` — unusual but not a bypass, since `is_valid()` already ran.

### Limitations
This review was performed by an AI agent (two-lens: Security, and Functionality+Quality run in parallel) in **local mode without GitHub API/SSH access** (both auth paths were unavailable this session). It does not understand business context, domain intent, organizational constraints, or deployment/rollout environment specifics. LOW-confidence findings and items in "Needs Human Judgment" require human verification. The `django-ansible-base` package (source of `CleanTextMixin`) is not installed in this sandbox, so its exact validation/auto-discovery behavior was inferred from the PR description and the new test file's black-box assertions, not read directly. This review is a first pass, not a final approval.

### Path to 10/10

1. **[Quality/Security] `activation.py:721-791`** (+1 pt, resolves known blocker too)
   > In `ActivationCopySerializer.copy()`, `description` and ~20 other fields are written via `super().create(copied_data)` without ever passing through `validate()`. Change `copy()` to validate the full `copied_data` payload (e.g., build it through a serializer whose `Meta.fields` includes `description`, or explicitly call the description through `CleanTextMixin`'s validation logic) before calling `create()`.

2. **[Quality] `tests/integration/api/test_clean_text_mixin.py`** (+1 pt)
   > Add regression tests proving each `excluded_fields` entry is actually exempted: `extra_var` (Activation) and `injectors` (CredentialType) accept Jinja2 template syntax; `token` (AwxToken) accepts a value containing Tier-2-blocklisted characters (e.g. `$(...)`) — all with `ENHANCED_INPUT_VALIDATION_ENABLED=True`.

3. **[Quality] `activation.py:370`, `event_stream.py:99`, `eda_credential.py:128`, `organization.py:71`** (+0.5 pt)
   > Remove `CleanTextMixin` from `ActivationSerializer`, `EventStreamOutSerializer`, `EdaCredentialCopySerializer`, and `OrganizationCreateSerializer` — none are ever instantiated with `data=`, so the mixin has no effect there. If kept intentionally, add a one-line comment saying so.

### Final Score
**Overall: 9.2/10 — NEEDS_CHANGES**

vidyanambiar:
## Code Review: [AAP-78702] Add CleanTextMixin to serializers

**Verdict**: NEEDS_CHANGES
**Score**: 8.2/10 (Functionality 8, Security 8, Quality 8.5)
**Risk**: CRITICAL (security-sensitive credential/user serializers touched, broad API surface — 10 of 11 changed files are serializers; no breaking changes detected)

### Key Findings

**Critical**
- **`activation.py:749-817`** — `ActivationCopySerializer.copy()`'s fix re-validates the copied `description` but not `k8s_service_name`/`source_mappings`, which are copied from the source activation into `copied_data` and persisted via `super().create()` with zero validation. Same silent-bypass bug class the PR exists to close, just on two sibling fields.

**Major**
- **[Security] `project.py`** — `Project.proxy` (embedded proxy credentials, `EncryptedTextField`) isn't in `excluded_fields` on either Project create/update serializer, unlike every other secret field this PR handles (`inputs`, `metadata`, `token`). Legitimate high-entropy proxy passwords will be blocklist-rejected once `ENHANCED_INPUT_VALIDATION_ENABLED=True`.
- **[Security] `user.py`** — `User.password` is similarly not excluded on `UserUpdateSerializerBase`, inconsistent with how `AwxToken.token` is already handled.
- **[Quality]** 8 serializers (`ActivationSerializer`, `ProjectSerializer`, `CredentialTypeSerializer`, `DecisionEnvironmentSerializer`, `TeamSerializer`, `UserSerializer`, `OrganizationCreateSerializer`, `EdaCredentialCopySerializer`) carry a non-functional `CleanTextMixin` — confirmed via grep none are ever instantiated with `data=`. A prior review pass fixed a different, non-overlapping set of read-only serializers.

Resolved since a prior review pass: `excluded_fields` acceptance tests now exist (`TestExcludedFieldsCleanText`), and `validate_shared_resource()` ordering is now consistent across organization/team/user.

### Path to 10/10
1. Extend `_ActivationCopyTextCheckSerializer.Meta.fields` to also cover `k8s_service_name` and `source_mappings`, and use the re-validated values in `copied_data`. (+2)
2. Add `excluded_fields = frozenset({"proxy"})` to `ProjectCreateRequestSerializer`/`ProjectUpdateRequestSerializer`. (+1)
3. Add `excluded_fields = frozenset({"password"})` to `UserUpdateSerializerBase`. (+1)

Full findings with evidence, verification commands, and confidence levels: available on request.

---
*Automated review — first pass, not a final approval. Human maintainer review still required.*

prat98 (approved):

tests/integration/api/test_clean_text_mixin.py
```diff
@@ -0,0 +1,930 @@
...
+            instance=default_event_stream,
+            data={"name": DANGEROUS_NAME},
+            partial=True,
+        )
+        assert not serializer.is_valid()
+        assert "name" in serializer.errors
+
+
+@pytest.mark.django_db
+class TestCredentialInputSourceCleanText:
```
wfealdel, line 739: Consider add similar test for `CredentialInputSourceCreateSerializer` rejecting invalid description on create

.tekton/run-atf-tests-pull-request.yaml
wfealdel, line 30: This seems an unrelated change, consider remove it from the PR if not needed

daphnemaeve: Yes, it looks like that was out-of-date. Rebasing the PR seems to have resolved this and removed it

tests/integration/api/test_clean_text_mixin.py
daphnemaeve: Thank you! Just pushed up a commit with that test case

wfealdel (dismissed):
LGTM!

wfealdel (approved):

B-Whitt (approved):

ttuffin (approved):

PR #1668, opened 2026-09-02
[AAP-87587] Provide validation patterns for nested fields and top level fields in metadata

daphnemaeve:
## Summary

Jira issue: https://redhat.atlassian.net/browse/AAP-87587

Advertise CleanTextMixin Tier 2 (`validate_free_text`) `pattern` / `pattern_description` on JSON sub-key schemas so API clients (AAP UI, AAP-87604) can show the same frontend hints used for top-level CharFields.

This change only adds metadata, gated by `ENHANCED_INPUT_VALIDATION_ENABLED`.

The only serializer affected is credential type:
- Credential type list/detail: `GET /api/eda/v1/credential-types/` → `inputs.fields`
- Credential type detail: `GET /api/eda/v1/credential-types/{id}/` → `inputs.fields`
- Credential type OPTIONS: `OPTIONS /api/eda/v1/credential-types/` → `actions.POST.{field}`

`OPTIONS /eda-credentials/` and credential `inputs` stay opaque JSON. Those keys depend on `credential_type`, so the schema lives on the credential type, not the instance.

Secret / password-typed / non-string sub-keys are skipped. Field lists are shallow-copied before injection so shared module-level schema dicts are never mutated.

All other top-level input fields that are validation should also have the pattern and patternDescription.

## Testing

**Prerequisites**
• aap-dev environment set up and functional

**Environment Setup**

- Configure your `aap-dev` sources to point at:
   - DAB → devel branch
   - EDA Server → PR branch (refs/pull/1668/head or the author's feature branch)

- Start the environment and ensure the EDA UI is accessible.
- Enable the feature gate — set this in your environment: `ENHANCED_INPUT_VALIDATION_ENABLED=True`

With `ENHANCED_INPUT_VALIDATION_ENABLED=true`:

1. `GET /api/eda/v1/credential-types/` — non-secret string `inputs.fields` entries include `pattern`, `pattern_description`, and `flags: i`. Secret fields (e.g. password, ssh key, token) do not.
2. `GET /api/eda/v1/credential-types/{id}/` for a credential type with string fields — same keys on `inputs.fields` string entries.
3. `OPTIONS /api/eda/v1/credential-types/` — top-level string fields (e.g. `description`) under `[actions.POST](http://actions.post/)` include `pattern` and `patternDescription`.
4. As this also provides the pattern and patternDescription in top-level input fields, they should also be visible in the OPTIONS response for resource names and descriptions.

With the toggle off, none of those schemas include `pattern` / `pattern_description`.

## Summary by CodeRabbit

- **New Features**
  - Added enhanced input validation guidance for credential type fields.
  - Non-secret text fields can display validation patterns and descriptions to clarify acceptable input.
  - Credential type metadata and options details can include clean-text validation guidance when supported and enabled.
  - Credential type creation and editing now use dedicated request handling.

- **Bug Fixes**
  - Secret fields remain excluded from enhanced pattern guidance.
  - Validation details correctly appear or remain hidden based on the enhanced validation setting.


src/aap_eda/api/serializers/credential_type.py
```diff
@@ -37,8 +39,47 @@ class Meta:
...
+        for field in fields
+    ]
+    return {**inputs, "fields": new_fields}
+
+class CredentialTypeCreateSerializer(
+    CleanTextMixin, serializers.ModelSerializer
+):
+    # injectors commonly contain Jinja2 template syntax, so it is excluded
+    # from free-text checks.
+    excluded_fields = frozenset({"injectors"})
```
vidyanambiar, line 84: Should this change live in this PR? 
I also see it in https://github.com/ansible/eda-server/pull/1660.

src/aap_eda/api/serializers/credential_type.py
```diff
@@ -37,8 +39,47 @@ class Meta:
...
+        return data
+
+
+def _with_field_patterns(inputs: dict) -> dict:
+    """Return a copy of the inputs schema with patterns for string fields.
+
+    Only non-secret "string" sub-fields get a pattern, since those are the
+    only ones CleanTextMixin's JSON sub-key validation enforces free-text
+    rules on; secret and boolean fields are left untouched.
+    """
```
vidyanambiar:
```suggestion
python
def _with_field_patterns(inputs: dict) -> dict:
    """Return a copy of the inputs schema with patterns for string fields.

    CleanTextMixin (from DAB) enforces free-text validation rules on
    serializer string fields at write time (when the ENHANCED_INPUT_VALIDATION_ENABLED setting is turned on).  Only non-secret "string"
    sub-fields get a pattern here, since those are the only ones its
    JSON sub-key validation applies to; secret and boolean fields are
    left untouched.
    """
```

vidyanambiar, line 59:
```suggestion
python
def _with_field_patterns(inputs: dict) -> dict:
    """Return a copy of the inputs schema with patterns for string fields.

    CleanTextMixin (from DAB) enforces free-text validation rules on
    serializer string fields at write time (when the ENHANCED_INPUT_VALIDATION_ENABLED setting is turned on).  Only non-secret "string"
    sub-fields get a pattern here, since those are the only ones its
    JSON sub-key validation applies to; secret and boolean fields are
    left untouched.
    """
```
Adding a brief description of CleanTextMixin for context would be helpful.

vidyanambiar (dismissed):

src/aap_eda/api/serializers/credential_type.py
daphnemaeve: This was implemented in 1660 as well, and was added here to help test. Resolving as this has been rebased and should no longer be duplicated!

vidyanambiar (dismissed):

AlexSCorey (commented):
I agree with the code rabbit issue identified above.

src/aap_eda/api/serializers/credential_type.py
```diff
@@ -38,6 +46,17 @@ class Meta:
...
 
+    def to_representation(self, instance):
+        data = super().to_representation(instance)
+        inputs = data.get("inputs")
+        if isinstance(inputs, dict):
+            # Advertise CleanTextMixin Tier 2 patterns on JSON
+            # sub-keys (AAP-87587).  Gated on
+            # ENHANCED_INPUT_VALIDATION_ENABLED; secret fields
+            # are skipped.
+            inject_patterns_into_field_list(inputs.get("fields"))
```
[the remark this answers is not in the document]
AlexSCorey: Yes, lets do this

src/aap_eda/api/serializers/credential_type.py
```diff
@@ -12,9 +12,17 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-from ansible_base.lib.serializers.mixins import CleanTextMixin
+try:
```
ttuffin, line 15: I am not sure about this change. I am concerned that an ImportError could hide a broken DAB installation and silently disable credential validation. Can you help me understand why this change is necessary?

src/aap_eda/api/serializers/credential_type.py
```diff
@@ -12,9 +12,17 @@
...
+    from ansible_base.lib.serializers.mixins import CleanTextMixin
+except ImportError:  # pragma: no cover - DAB without AAP-85987
```
wfealdel, line 17: **Security concern: validation fallback fails open**

If importing `CleanTextMixin` raises `ImportError`, this fallback defines a no-op replacement. `CredentialTypeCreateSerializer` then remains usable, but credential-type writes no longer receive the validation provided by `CleanTextMixin`.

This can silently disable input validation when DAB is missing, incompatible, partially broken, or when an import-time dependency error occurs. The current lockfile should provide the mixin in supported deployments, but the fallback still allows an invalid deployment to accept unvalidated credential-type input without an explicit error.

Please avoid replacing this security-relevant mixin with a no-op. Prefer making the dependency mandatory and failing clearly during startup/import. If backward compatibility is required, keep fallbacks limited to optional metadata helpers, not write-time validation.


---
**Note:** This review was prepared with assistance from OpenAI’s `gpt-5.6-luna` model.

daphnemaeve:
* Fixed mutation issue noted by coderabbit and @AlexSCorey 
* Removed import fallback noted by @ttuffin  and @wfealdel entirely
    * That fallback was an outdated implementation to allow testing and review on this PR before DAB 1119 was merged
    * DAB 1119 is now merged, meaning the try/catch block was no longer needed

wfealdel (approved):
LGTM!

src/aap_eda/api/validation_patterns.py
```diff
@@ -0,0 +1,125 @@
...
+``build_tier2_frontend_pattern`` and ``inject_clean_text_patterns``.
+Until that lands on DAB devel, injection is a no-op so EDA can
+still import and run.
+"""
+
+import copy
+
+from ansible_base.lib.utils.settings import get_setting
+
+try:
```
ttuffin: @daphnemaeve similar to the previous query, are the try/except clauses for these imports necessary? They reference a Jira that has been closed

daphnemaeve: Sorry about that, I thought I had patched up all the try/except instances. Those are not necessary. I'm pushing up another fix to ensure all serializers that use the mixin also provide the pattern info and description in the OPTIONs response, and will also remove that try/except catch.

ttuffin (approved):

wfealdel (approved):

PR #1669, opened 2026-09-03
Add Kubernetes pod affinity support for activation job pods

Francisco-xiq:
### What is being changed?

Adds support for specifying Kubernetes pod affinity rules (`k8s_pod_affinity`) on Activations, following the same pattern already used for `k8s_pod_tolerations` and `k8s_pod_node_selector`. When set, the value is validated for structural shape and passed through to the pod spec of the Kubernetes Job created for the activation.

### Why is this change needed?

Refs ansible/eda-server-operator#226

### How does this change address the issue?

Adds a `k8s_pod_affinity` field to the `Activation` model, exposes it through the relevant serializers, and applies it to the `V1PodSpec` built by the Kubernetes container engine when starting an activation.

Files:
- `src/aap_eda/core/models/activation.py` — add `k8s_pod_affinity` field
- `src/aap_eda/core/migrations/0074_activation_k8s_pod_affinity.py` — migration for the new field
- `src/aap_eda/core/validators.py` — add `check_if_k8s_pod_affinity_valid` shape validator
- `src/aap_eda/api/serializers/activation.py` — expose `k8s_pod_affinity` across the activation serializers
- `src/aap_eda/services/activation/engine/common.py` — pass `k8s_pod_affinity` through to `ContainerRequest`
- `src/aap_eda/services/activation/engine/kubernetes.py` — apply `k8s_pod_affinity` to the pod spec built for activation job pods

### Does this change introduce any new dependencies, blockers or breaking changes?

The field is optional and defaults to empty, matching existing behavior when unset.

### How it was tested?

Automated tests:
- `tests/unit/test_k8s_pod_affinity.py` — shape validation
- `tests/integration/services/activation/engine/test_kubernetes.py` — affinity applied to the pod spec by the Kubernetes engine, with and without affinity set
- `tests/integration/api/test_activation.py` — activation start-validation with `k8s_pod_affinity` set
- `tests/integration/services/activation/test_activation.py` — `k8s_pod_affinity` passed through to `ContainerRequest`

Manual testing:
Verified against a real Kubernetes cluster that Activations created with `k8s_pod_affinity` are scheduled onto nodes matching the specified affinity rules, and that changing the affinity value correctly moves the pod to the matching node. Also verified that when `k8s_pod_affinity` is not set, no affinity is applied to the pod spec, leaving the scheduling decision entirely up to the cluster's scheduler as before.

## Summary by CodeRabbit

- **New Features**
  - Added Kubernetes pod affinity configuration for activations.
  - Affinity settings can be created, updated, copied, viewed, and applied to activation job pods.
  - Supports node affinity, pod affinity, and pod anti-affinity rules.
  - Added validation for supported affinity structures.

- **Bug Fixes**
  - Activations without affinity settings continue to run without scheduling constraints.

- **Tests**
  - Added coverage for affinity validation, persistence, API handling, and Kubernetes job pod configuration.

### DIFF
--- a/src/aap_eda/api/serializers/activation.py
+++ b/src/aap_eda/api/serializers/activation.py
@@ -417,6 +417,7 @@ class ActivationSerializer(serializers.ModelSerializer):
             "k8s_pod_annotations",
             "k8s_pod_node_selector",
             "k8s_pod_tolerations",
+            "k8s_pod_affinity",
         ]
         read_only_fields = [
             "id",
@@ -498,6 +499,7 @@ class ActivationListSerializer(
             "k8s_pod_annotations",
             "k8s_pod_node_selector",
             "k8s_pod_tolerations",
+            "k8s_pod_affinity",
         ]
         read_only_fields = [
             "id",
@@ -570,6 +572,7 @@ class ActivationListSerializer(
             "rule_engine_credential_id": activation.rule_engine_credential_id,
             **_activation_k8s_pod_metadata_payload(activation),
             "k8s_pod_tolerations": activation.k8s_pod_tolerations,
+            "k8s_pod_affinity": activation.k8s_pod_affinity,
         }
 
 
@@ -606,6 +609,7 @@ class ActivationCreateSerializer(
             "k8s_pod_annotations",
             "k8s_pod_node_selector",
             "k8s_pod_tolerations",
+            "k8s_pod_affinity",
         ]
 
     rulebook_id = serializers.IntegerField(
@@ -658,6 +662,11 @@ class ActivationCreateSerializer(
         required=False,
         default=list,
         validators=[validators.validate_k8s_pod_tolerations],
+    )
+    k8s_pod_affinity = serializers.JSONField(
+        required=False,
+        default=dict,
+        validators=[validators.check_if_k8s_pod_affinity_valid],
     )
 
     def validate(self, data):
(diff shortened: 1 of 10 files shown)


Francisco-xiq:
Hi, that's my first contribution on this repository, please let me know if I need to change anything on my approach.

Francisco-xiq:
I noticed the e2e tests (single-node and multinode) failed and wanted to make sure it wasn't something introduced by this PR, so I dug into it a bit.

**Reproducing locally**

```bash
docker compose -p eda -f tools/docker/docker-compose-dev.yaml build
docker compose -p eda -f tools/docker/docker-compose-dev.yaml up -d
docker compose -p eda -f tools/docker/docker-compose-dev.yaml logs eda-api
```

The `eda-api` container exits with code 2, and the last log line is:
`tput: No value for $TERM and no -T specified`

This comes from `scripts/common/logging.sh` (sourced by `scripts/create_superuser.sh`, which is part of the `eda-api` container's startup command). Since the script runs with `set -o errexit`, that `tput` failure aborts the script, which in turn breaks the whole startup chain (`migrate && create_initial_data && create_superuser.sh && runserver`).

**Checking if it's related to this PR**

I ran the exact same commands against a clean checkout of `main` , and got the identical failure, same error, same exit code. So it doesn't look related to the changes in this PR.

I also checked the git history for `TERM` in `docker-compose-dev.yaml`/`Dockerfile` and didn't find it ever being set there. I didn't do a full bisect, so I can't say for certain how long this has been broken, but i didn't find something that points to a recent regression, it looks like it may have been like this for a while

Happy to open a separate issue/PR for this if it's useful

PR #1671, opened 2026-09-04
Skip color output in logging.sh when no usable terminal is available

### LINKED ISSUE #1670: scripts/common/logging.sh breaks container startup when no TTY is available (tput fails)
### Please confirm the following

- [x] I agree to follow this project's [code of conduct](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html).
- [x] I have checked the [current issues](https://github.com/ansible/aap-eda/issues) for duplicates.
- [x] I understand that EDA-Controller is open source software provided for free and that I might not receive a timely response.

### Bug Summary
`scripts/common/logging.sh` calls `tput` unconditionally to set up color codes. When a container is started without an allocated `TTY` (e.g. `docker compose ... up -d running /bin/bash -c "<command>"`, with no `tty: true`), there's no `$TERM` value available, so `tput` fails with tput: No value for `$TERM` and no `-T` specified. Since the script runs with `set -o errexit`, this aborts the script, and since it's sourced by `scripts/create_superuser.sh`, which is chained with `&&` in the `eda-api` service's startup command in `docker-compose-dev.yaml`, the whole `eda-api` container fails to start.

This currently breaks the **e2e** test workflows (single-node and multinode), which bring up the stack via `docker compose -f tools/docker/docker-compose-dev.yaml up -d`.

### Environment

`tools/docker/Dockerfile` base image: CentOS Stream 9 (quay.io/centos/centos:stream9)
Observed in GitHub Actions (ubuntu-latest runners), reproducible locally with Docker/Podman

### Steps to reproduce

```
docker compose -p eda -f tools/docker/docker-compose-dev.yaml build
docker compose -p eda -f tools/docker/docker-compose-dev.yaml up -d
docker compose -p eda -f tools/docker/docker-compose-dev.yaml logs eda-api
```

Minimal reproduction of the underlying issue, independent of this project:

```
podman run quay.io/centos/centos:stream9 /bin/bash -c "echo test && env"
```

**`$TERM` is absent from the output. The base image simply doesn't set it when run without a TTY.**

### Actual results
The eda-api container exits with code 2. The last log line is:

`tput: No value for $TERM and no -T specified`

`scripts/create_superuser.sh` (and therefore `migrate && create_initial_data && create_superuser.sh && runserver`) never completes.

### Expected results
`scripts/common/logging.sh` should not fail when no terminal is available. It should fall back to plain output instead of aborting the whole startup chain.

### Additional information
This isn't caused by a recent change to this repo: `TERM` has never been set in `docker-compose-dev.yaml` or the `Dockerfile`, and the `docker-compose-dev.yaml` service commands have always invoked `/bin/bash -c "<command>"` without `tty: true`.

Francisco-xiq:
### What is being changed?
`scripts/common/logging.sh` only checked whether the `tput` binary existed before using it to set up colored log output. It didn't check whether `tput` would actually work in the current environment. This change strengthens that check to also verify that `stdout` is attached to a real terminal, that `$TERM` is set, and that a real `tput` call succeeds. Falling back to the existing plain-text log functions otherwise.

### Why is this change needed?
Refs #1670

### How does this change address the issue?
When a container runs a command via `/bin/bash -c "..."` without an allocated TTY (as docker-compose-dev.yaml does for eda-api), `$TERM` is unset, so `tput` fails with `tput: No value for $TERM and no -T specified`. Since `scripts/create_superuser.sh` (which sources this file) runs with `set -o errexit`, that failure aborted the whole startup chain, breaking the `eda-api` container and, in turn, the `E2E` test workflows.

### Files:

`scripts/common/logging.sh` — strengthen the guard before attempting colored output

### Does this change introduce any new dependencies, blockers or breaking changes?
No new dependencies. No behavior change for interactive terminal use; environments without a usable terminal now get plain-text logs instead of crashing.

### How it was tested?
Verified in isolation that a failing check inside the if condition doesn't trigger `set -o errexit` in the calling script (tested with a minimal reproduction script).

Reproduced the original failure and confirmed the fix resolves it:

```
bash
docker compose -p eda -f tools/docker/docker-compose-dev.yaml build
docker compose -p eda -f tools/docker/docker-compose-dev.yaml up -d
curl -s http://localhost:8000/_healthz
```

Before this change, `eda-api` exits with code 2 and every dependent service stays stuck waiting. After this change, `eda-api` reports healthy and the health check returns `{"status":"OK"}`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved command-line log output detection to avoid ANSI colors when the terminal does not support them or the required environment settings are unavailable.
  - Plain, uncolored logs are now used automatically in unsupported terminal environments.

### DIFF
--- a/scripts/common/logging.sh
+++ b/scripts/common/logging.sh
@@ -3,7 +3,10 @@
 # timestamp
 TIMESTAMP=$(date "+%Y-%m-%d %H:%M:%S")
 
-if which tput &> /dev/null; then
+if command -v tput &> /dev/null \
+    && [ -t 1 ] \
+    && [ -n "${TERM:-}" ] \
+    && tput setaf 1 &> /dev/null; then
   # colors
   ERR=$(tput setaf 1)
   INFO=$(tput setaf 178)


ttuffin (approved):

ptoscano (approved):
More defensive programming than needed (IMHO), still good though.

Thanks!

B-Whitt:
Jira - https://redhat.atlassian.net/browse/AAP-91748

PR #1672, opened 2026-09-04
[DO NOT MERGE] test(advisory-lock): surface thread exceptions in test_job_uniqueness

B-Whitt:
## Summary
- Diagnostic PR to identify the root cause of `test_job_uniqueness[gather_analytics]` failure
- The test currently swallows exceptions from `ThreadPoolExecutor` futures — `wait()` stores exceptions in `Future` objects but never re-raises them
- Adds `future.result()` diagnostics after `wait()` so CI output shows the actual exception causing `len(call_log) == 0`

## Changes
- `tests/integration/test_advisory_lock.py` — check futures for exceptions/timeouts before the final assert

## Test plan
- [ ] CI runs `test_job_uniqueness[gather_analytics]` with `-m multithreaded` and surfaces the actual error

## Summary by CodeRabbit

- **Tests**
  - Improved integration test diagnostics for concurrent worker failures and timeouts.
  - Failure messages now identify the affected worker and include relevant call activity, making test failures easier to investigate.
  - Retained validation that the expected single call is recorded.

### DIFF
--- a/tests/integration/test_advisory_lock.py
+++ b/tests/integration/test_advisory_lock.py
@@ -117,8 +117,7 @@ def test_job_uniqueness(module_data):
 
     if errors:
         pytest.fail(
-            f"Thread errors (call_log={call_log}):\n"
-            + "\n".join(errors)
+            f"Thread errors (call_log={call_log}):\n" + "\n".join(errors)
         )
 
     assert len(call_log) == 1
--- a/tests/integration/test_advisory_lock.py
+++ b/tests/integration/test_advisory_lock.py
@@ -108,4 +108,17 @@ def test_job_uniqueness(module_data):
         ]
         wait(futures, timeout=3)
 
+    errors = []
+    for i, f in enumerate(futures):
+        if f.done() and f.exception():
+            errors.append(f"Thread {i}: {f.exception()}")
+        elif not f.done():
+            errors.append(f"Thread {i}: timed out (not done after 3s)")
+
+    if errors:
+        pytest.fail(
+            f"Thread errors (call_log={call_log}):\n"
+            + "\n".join(errors)
+        )
+
     assert len(call_log) == 1


B-Whitt:
TL;DR: Test failure is intermittent, only in GH CI. After rerunning the test on pr 1650, it passed.

Troubleshooting context if we need to investigate further at a later time:
## Appendix: `test_job_uniqueness[gather_analytics]` failure investigation

**Date**: 2026-09-04
**Test**: `tests/integration/test_advisory_lock.py::test_job_uniqueness[gather_analytics]`
**Branch**: `feat/AAP-77938-log-retention-phase2` (also reproduces on `main`)
**Failure**: `assert 0 == 1` — `len(call_log)` is 0, meaning `_gather_analytics` was never called by either thread.

### What the test does

Two threads concurrently call `gather_analytics()`, which acquires a PostgreSQL
advisory lock (`pg_try_advisory_lock`, `wait=False`). Only one thread should
acquire the lock and call the mocked `_gather_analytics`. The test asserts
`len(call_log) == 1`.

### What we validated (facts)

1. **`gather_analytics()` and `_gather_analytics()` are unchanged** on this
   branch. `git diff f01fbd6d..HEAD -- src/aap_eda/tasks/analytics.py` shows
   only `schedule_gather_analytics` and `reschedule_gather_analytics` were
   modified (dispatcherd migration cleanup). The functions under test are
   identical to when they last passed.

2. **The test silently swallows thread exceptions.** `wait(futures, timeout=3)`
   at line 109 does not re-raise exceptions from `Future` objects — they are
   stored but never inspected. If both threads raise, `call_log` stays empty
   and the assert sees `len([]) == 0`. Verified by reading `concurrent.futures`
   docs and the test source. Filed diagnostic PR
   [#1672](https://github.com/ansible/eda-server/pull/1672) to surface the
   actual exceptions in CI.

3. **The `advisory_lock` implementation requires a live PostgreSQL cursor.**
   DAB's `advisory_lock` (`.venv/src/django-ansible-base/ansible_base/lib/utils/db.py:129`)
   checks `connection.vendor == "postgresql"`, then calls
   `connections[using].cursor()` + `cursor.execute("SELECT pg_try_advisory_lock(...)")`.
   In child threads, `django.db.connection` is thread-local — each thread gets
   its own connection, not the test's transaction-wrapped connection.

4. **The test was already known to be fragile.** Commit `f01fbd6d` (Alex
   Izquierdo, 2025-06-05) quarantined it behind a `multithreaded` marker with
   the note: "seems to interfere intermittently with other mocks." The marker
   causes it to be skipped unless explicitly selected via `-m multithreaded`.

5. **`importlib.reload` race condition exists by design.** Both threads reload
   the same module object (`aap_eda.tasks.analytics`). Thread 2's reload
   replaces all module-level names — including thread 1's mock of
   `_gather_analytics` — with freshly-defined real functions. This is
   observable from the code structure; no timing luck is needed for the race
   to occur.

6. The test passes when run locally. Only fails in GH CI

**Related Jira issues:**
   - [AAP-47554](https://redhat.atlassian.net/browse/AAP-47554) — `[2.5 Backport] AAP-47111` (Closed)
   - [AAP-47111](https://redhat.atlassian.net/browse/AAP-47111) — `sync and import task is not actually running under advisory lock` (Closed)
   - No existing Jira ticket for this specific test failure.

### Hypothesis (unverified)

The most likely failure sequence:

1. Thread 1 reloads the module, patches `_gather_analytics`, calls
   `gather_analytics()`.
2. Thread 2 reloads the module — this re-executes the module body, replacing
   thread 1's mock with the real `_gather_analytics` function.
3. Thread 1 acquires the advisory lock, calls `_gather_analytics()` — but now
   it's the **real** function (mock was undone by reload).
4. Real `_gather_analytics` calls `utils.get_insights_tracking_state()` →
   returns `True` (mocked). Then calls `collector.gather(logger=logger)` →
   **raises an exception** because no analytics infrastructure exists in the
   test environment.
5. Exception propagates through `gather_analytics()` → caught by `pytest.fail()`
   in the test → stored in `Future`, never checked.
6. Thread 2 patches and calls `gather_analytics()` → advisory lock either
   already released (thread 1 exited the context manager on exception) or
   still held briefly → thread 2 either gets `acquired=False` and returns,
   or also hits the same real-function-runs scenario.
7. `call_log` is empty. `assert len(call_log) == 1` fails.

An alternative hypothesis: the child threads fail to establish a PostgreSQL
connection at all (thread-local `django.db.connection` not properly initialized
in `ThreadPoolExecutor` workers under `@pytest.mark.django_db` without
`transaction=True`), and both threads raise `OperationalError` before reaching
the advisory lock.

### What needs investigation

- **PR #1672 CI results** — the diagnostic PR surfaces `future.exception()` for
  each thread. The CI output will confirm which exception is raised and in which
  thread, distinguishing between the two hypotheses above.
- **Whether `@pytest.mark.django_db(transaction=True)` changes behavior** — if
  child threads need transactional database access, the current marker may be
  insufficient.
- **Whether the `importlib.reload` approach can be replaced** — the reload is
  the root cause of the mock instability. A redesign that avoids reloading
  (e.g., passing the function as a parameter, or using `threading.Barrier` to
  synchronize patch-then-call) would eliminate the race.

Possibly related:
Yes—three PRs by Alex changed or tried to stabilize this test:
- #1334 — directly addressed its flakiness by running multithreaded tests separately. Its description explicitly cites sporadic test_job_uniqueness mock interference. This is the closest prior fix attempt, though it did not remove the test’s internal reload/patch race.
- #1299 — reworked test_job_uniqueness to mock proxy functions.
- #1333 — corrected which project import/sync work executes under the advisory lock and consolidated the uniqueness tests.

In DAB, no PR by someone else specifically addresses this EDA test failure. The relevant DAB PRs are:
- #661 — introduced the advisory-lock utility.
- #713 — improved its error handling.
Those DAB changes predate or concern the helper itself; they do not remove the concurrent reload()/mock mutation in EDA’s test.

PR #1680, opened 2026-09-16
fix: enforces no empty strings for name on remaining models.

AlexSCorey:
Most of our models mandate that the name field must have a value except for 4. This pr brings those 4 into compliance.  https://redhat.atlassian.net/browse/AAP-78706

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented empty names from being saved for AWX tokens, event streams, job instances, and rulebook processes.
  - Automatically assigns generated names when job or rulebook process names are blank.
  - Updated existing records with blank names to use generated names.
  - Added database validation to reject future records with empty names, ensuring these resources retain reliable, non-empty names.


ttuffin (dismissed):

wfealdel (dismissed):
LGTM!

B-Whitt:
@AlexSCorey - This is an observability issue for malformed-but-accepted job messages, not a failure for valid job names.

The new non-empty database constraint on JobInstance.name can be violated by an existing production creation path: this call creates a JobInstance using a name taken directly from the incoming event payload, with no non-empty validation or fallback. An event carrying `name: ""` will now fail at the database layer rather than being handled as before.

User impact:
- The error is only logged; no user-facing validation error or retry is returned ([consumers.py (line 301)](/Users/bwhittin/code/claude-bw-atf/eda-server/src/aap_eda/wsapi/consumers.py:301)).
- The user may see the playbook run, but its job record, status, history, or related events may be missing from the EDA UI/API.

Suggested fix: 
- in src/aap_eda/wsapi/consumers.py around lines 589–592, validate message.name before creation and assign a deterministic fallback such as f"job-{message.job_id}" when it is empty. 
- Add an integration test in tests/integration/wsapi/test_consumer.py covering a blank job name and asserting successful persistence with the fallback.

src/aap_eda/core/migrations/0074_awxtoken_ck_awx_token_name_not_empty_and_more.py
```diff
@@ -0,0 +1,68 @@
...
+    ]
+
+    operations = [
+        migrations.RunPython(
+            backfill_rulebook_process_names, migrations.RunPython.noop
+        ),
+        migrations.RunPython(
+            backfill_job_instance_names, migrations.RunPython.noop
+        ),
+        migrations.AddConstraint(
```
wfealdel, line 68: This migration backfills `RulebookProcess` and `JobInstance`, but it does not backfill `AwxToken` or `EventStream` before adding their constraints.

Will the migration fail if an existing production database contains an empty AwxToken.name or EventStream.name?

AlexSCorey: Good catch!

wfealdel (approved):
LGTM!

ttuffin (approved):

